PR #2, opened 2011-03-31
Develop

jsarch:
Rest router checks if method is callable before invoking. Returns 404 if method is not callable.


jsarch:
Example of issue:

1) Install Fuel as documented

2) Create test REST controller : http://fuelphp.com/docs/general/controllers/rest.html

3) Go to: http://localhost/test/missing_method.json?foo=bar

Rather than cause an exception, I think the REST controller should return a 404 as in Request::execute(). My change above tried to follow the same approach.

philsturgeon:
Done slightly differently, but thank you.

PR #3, opened 2011-04-01
Form::input() bugfix

kbanman:
Fixed bug that appears when a 'value' attribute isn't passed to Form::input()

$attributes['value'] is referenced further on, and caused a 'not defined' notice


jschreuder:
I can't pull this for some reason, I'll change it myself. Thanks for the bugfix.

PR #4, opened 2011-04-02
Fixed bug in email class

huglester:


jschreuder:
Pulled and synced among branches.

Thanks

PR #5, opened 2011-04-03
Fixed #76: Grammatical mistake with pluralize and singularize with sample word 'campus'

### LINKED ISSUE #76: Controllers in added paths
I'm not sure if this is a bug or the intended behavior, but when I add a path to Fuel for inclusion in the file system (through `Fuel::add_path()`), I can't use controllers in the added path location. I tossed a controller in to the added path's `classes/controller` directory and went there in my browser only to be met by a Fuel exception (**Fuel\Core\Fuel_Exception [ Error ]: It appears your _404_ route is incorrect. Multiple Recursion has happened.**).

At this point, my controller is dead simple as I'm just trying to get it to print something out:

``` php
<?php

class Controller_Welcome extends Controller {

    public function before()
    {
        parent::before();
    }

    {
        echo 'Foo!';
    }
}
```

Is this a bug or expected behavior? If it's the latter, I'm just curious why controllers aren't allowed in those added paths. Thanks!

crynobone:
Refer to http://dev.fuelphp.com/issues/76

### DIFF
--- a/classes/inflector.php
+++ b/classes/inflector.php
@@ -43,7 +43,7 @@ class Inflector {
 		'/(m)an$/'					=> '\1en',		// man, woman, spokesman
 		'/(c)hild$/'				=> '\1hildren',	// child
 		'/(buffal|tomat)o$/'		=> '\1\2oes',	// buffalo, tomato
-		'/(bu)s$/'					=> '\1\2ses',	// bus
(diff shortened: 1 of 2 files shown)


jschreuder:
The fix didn't show up in pull requests so I changed it by hand. Thanks for the fix!

PR #7, opened 2011-04-03
Changes to Controller_Rest

tarnfeld:
I needed to change the base node for the XML output and that can't be set anywhere in the ->response() method (doesn't make sense to go there) so I was going to extend the _format_xml method to set my own default but because it's private, I can't.

I've updated the formatting methods to fix this...

### DIFF
--- a/classes/controller/rest.php
+++ b/classes/controller/rest.php
@@ -337,7 +337,7 @@ abstract class Controller_Rest extends \Controller {
 
 	// FORMATING FUNCTIONS ---------------------------------------------------------
 	// Format XML for output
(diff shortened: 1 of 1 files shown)


crynobone:
Controller_Rest works great, but for lazy programmer like myself, I prefer to extends something and works great all the time since it inherit from a solid Class instead of having a whole duplicate class just to have that tiny extra thing.

philsturgeon:
Yes I understand. I'll ask one more time, what were you trying to achieve? What was the issue that made you think "it would be nice to extend this method"? I'm not asking about extending, im asking about the end goal here.

PR #11, opened 2011-04-05
Small fixed on Fuel::find_file() bug and path management

crynobone:
Update two things:
- Return valid path_cache value, currently referring to invalid path
- Reset path_cache everytime a new path is added.


crynobone:
Yeah, I remove that. But can you look again what the diff, the only pull request left is to return valid list of file founds from cached path.

crynobone:
It's already added

PR #12, opened 2011-04-06
Fix a couple things in form class to return valid html

jeffutter:
label should not have a for attribute that is null (no id set)

textarea and select should have no type attribute

### DIFF
--- a/classes/form.php
+++ b/classes/form.php
@@ -433,6 +433,8 @@ class Form {
 			$attributes['id'] = static::get_class_config('auto_id_prefix', '').$attributes['name'];
 		}
 
+		unset($attributes['type']);
+
 		return html_tag('textarea', static::attr_to_string($attributes), $value);
 	}
 
@@ -501,6 +503,8 @@ class Form {
 		}
 		$input .= str_repeat("\t", 0);
 
+		unset($attributes['type']);
+
 		return html_tag('select', static::attr_to_string($attributes), $input);
 	}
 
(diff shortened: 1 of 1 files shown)


jschreuder:
When is this a problem, when using OO generation probably?

You're on the right track but I don't like disallowing individual properties. I'd rather prevent it from happening automaticly, but wouldn't disallow it for someone writing bad HTML (or we'd have to create full Html checking libs, which would add an amount of overhead that's insane). I agree with the line only setting the "for" attribute for labels when there's actually a value for it, but the others I'd rather unset in the build_field() method and not in each individual method. I'll look into this tonight.

Edit: also I might leave out the change to for and change our array_to_attr() method to ignore attributes with value **null**.

jeffutter:
Yeah it was happening in OO generation. Ex:

$note_form = Fieldset::factory('add_note')
            ->add_model('Model_Note', $job, 'set_form_fields_jobpage')
            ->repopulate()
            ->build('jobs/'.$job->id.'/notes/add');

Where the model has things like:

public static function set_form_fields_jobpage(Fieldset $form, $job = null)
    {
        $form->add('note', 'Note', array(
                'type' => 'textarea',
                'value' => ''
            ), array(
                array('required')
            ));

this would result in something like:

<label for="">Note</label><textarea type="textarea" ....

ignoring attributes with a null value sounds like a good solution for the for="" problem

jschreuder:
Fixed in the develop branch of fuel/core.

PR #13, opened 2011-04-06
Pagination - Increase flexibility

kbanman:
Moved the page link assembly into it's own method so they can be retrieved separate from the [hardcoded] previous and next links.
This enables more flexibility in how the pagination is displayed on the page. Still needs more development in that direction, but I think this is pretty essential.

Oh, also added a check to see if current page is null before automatically assigning it. This allows manual specification of the page number, which I had to do when using pagination where the page number isn't always in the same segment due to routes.

### DIFF
--- a/classes/pagination.php
+++ b/classes/pagination.php
@@ -107,7 +107,7 @@ class Pagination {
 
 		static::$total_pages = ceil(static::$total_items / static::$per_page) ?: 1;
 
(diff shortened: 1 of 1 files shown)


kbanman:
Another thing, it would be good to be able to access the total_items property. Wasn't sure if the way to do that was by making the property public or add an getter.

jschreuder:
Merged with dev, don't really have an opinion on the other suggestion as I never use Pagination libs.

PR #14, opened 2011-04-07
form_template

jeffutter:
This should allow using form_template on a Form object just like field_template. I've verrified by setting $form->set_config('form_template', '...') as well as setting it in a config/form.php file.


jschreuder:
You got the branch wrong, you shuold always pull-request develop and make sure it doesn't contain any commits other than those this is about.
Fortuneately I could cherry pick this one into develop through Github, but most often it would take us a lot of time to merge if you get this wrong.

PR #17, opened 2011-04-07
Fix Date class. strptime returns years since 1900 not 1901. Dates were a 

jeffutter:
Fix Date class. strptime returns years since 1900 not 1901. Dates were a year in the future. Fixed


jschreuder:
Pulled, thanks.

PR #18, opened 2011-04-07
Allow setting labels as array including attributes instead of just tring 

jeffutter:
Allow setting labels as array including attributes instead of just tring in form->add. Strings still work as expected


jschreuder:
Pulled, thanks.

PR #19, opened 2011-04-08
Fix form->add type=radio.

jeffutter:
Options wasn't being passed when adding a radio.. thus it wasn't making all the separate fields. Also there was no default tempalte and with no template it didn't render the labels. Still no way to set which field is 'radiod' by default.


classes/fieldset/field.php
```diff
@@ -87,6 +87,7 @@ class Fieldset_Field
 	{
 		$this->name = (string) $name;
 		$this->fieldset = $fieldset;
+		if(isset($attributes['options'])) { $this->set_options($attributes['options']);}
```
jschreuder, line 91: This isn't up to our coding standards, no single line if statements.

classes/form.php
```diff
@@ -637,13 +637,15 @@ class Form {
 			}
 		}
 
+		$options = $field->options();
+
```
jschreuder: this isn't necessary for all, only used by select (directly) and radio/checkbox. As such this should be after the case statement.

classes/form.php
```diff
@@ -725,7 +727,7 @@ class Form {
 					$bf_temp = str_replace('{label}', $label, $bf_temp);
 					$build_fields .= $bf_temp;
 				}
-				$template = str_replace($match[1], $build_fields, $template);
+				$template = str_replace($match[1], $build_fields, $match[1]);
```
jschreuder: Shouldn't this be $template at the end? This way you're removing anything before and after the {fields} and {/fields}  tags.

jeffutter:
Ok, i fixed the above issues. Hope that's good.

PR #20, opened 2011-04-08
Feature #101

jonstuebe:
I'm not sure exactly how to go about this. But I created a feature request on dev.fuelphp.com and went ahead and started the class myself. If this isn't the proper way to go about putting a request for the file, then let me know.

### DIFF
--- a/classes/typography.php
+++ b/classes/typography.php
@@ -1,26 +1,22 @@
 <?php
+/**
+ * Fuel
+ *
+ * Fuel is a fast, lightweight, community driven PHP5 framework.
+ *
+ * @package		Fuel
+ * @version		1.0
+ * @author		Jon Stuebe <http://jonstuebe.com>
+ * @license		MIT License
+ * @copyright	2010 - 2011 Fuel Development Team
+ */
 
 namespace Fuel\Core;
 
 class Typography {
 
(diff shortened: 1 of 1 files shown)


classes/typography.php
```diff
@@ -0,0 +1,83 @@
+<?php
+
+namespace Fuel\Core;
+
+class Typography {
+
+	public static function auto($str)
+	{
+		View::$auto_encode = false;
+
```
jschreuder: I can be quite clear about this: a class that auto-disables output encoding will never go into the core. This instruction should go.

jschreuder:
Looking it over I think it's a good idea in principle, it's a bit feature light currently though.

Also I think you could do without a special html_entities method, you could first htmlentities it and only after that convert newlines to linebreaks and paragraphs. Would be far better practice than partially duplicating build in behavior.

jonstuebe:
I've went ahead and adjusted the class. I know it's a little light on the features. But I think it's something to start with and we can add more later.

dhrrgn:
I don't like this here.  If anything it should go in the Html or the Str class.

dhrrgn:
You obviously don't care about this too much, haven't responded in over a week.  Closing.  Read my comments above if you want this in.

PR #21, opened 2011-04-09
added redis class to bootstrap

tarnfeld:


jschreuder:
I did it myself to keep the order alphabetical, thanks for the correction.

PR #23, opened 2011-04-10
Fix cache path

jeffutter:
The cache path was missing the cache_id. It was causing errors with Lang (at least). I think someone else had another commit with this fix.. but it hasn't been accepted yet due to other problems with other issues addressed in his pull request. This pull fixes just this single issue.


jeffutter:
Looks like this was fixed elsewhere

PR #24, opened 2011-04-10
Populate checkbox radios

jeffutter:
This should take care of repopulating and setting default vals for checkboxes and radios... check commit log for example

### DIFF
--- a/classes/fieldset.php
+++ b/classes/fieldset.php
@@ -323,10 +323,9 @@ class Fieldset
 			{
 				if (($value = $this->input($f->name, null)) !== null)
 				{
-					error_log('name: '.$f->name.' type: '.$f->type.' value: '.$value);
-					if($f->type == 'checkbox')
+					if($f->type == 'checkbox' or $f->type == 'radio')
 					{
-						$f->set_checked();
+						$f->set_checked($value);
 					}
 					$f->set_value($value);
 				}
--- a/classes/fieldset.php
+++ b/classes/fieldset.php
@@ -303,11 +303,19 @@ class Fieldset
 				{
 					if ($value = $input[$f->name])
 					{
+						if($f->type == 'checkbox')
+						{
+							$f->set_checked();
+						}
 						$f->set_value($value);
 					}
 				}
 				elseif (is_object($input) and property_exists($input, $f->name))
 				{
+					if($f->type == 'checkbox')
+					{
+						$f->set_checked();
+					}
 					$f->set_value($input->{$f->name});
 				}
 			}
@@ -315,6 +323,11 @@ class Fieldset
 			{
 				if (($value = $this->input($f->name, null)) !== null)
 				{
+					error_log('name: '.$f->name.' type: '.$f->type.' value: '.$value);
+					if($f->type == 'checkbox')
+					{
+						$f->set_checked();
+					}
 					$f->set_value($value);
 				}
 			}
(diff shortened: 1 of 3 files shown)


jschreuder:
I think it's almost there, but not quite yet. Also noticed a bug by me in [repopulate](https://github.com/jeffutter/fuel-core/blob/95d590fdd5b822da7855a9c0b631ba428bec6782/classes/fieldset.php#L304) where it's index exists checking method will give you errors with incomplete input, which it shouldn't. (this is more of a note to self btw)

But one of the things I noticed, and I'm not sure if I understand, is that you handle array/object input different from repopulation with the POST variable, something I'm not so sure about.
Also I wonder if this already solves the problem of arrays in the POST, when someone uses fieldname[].

jeffutter:
Ok, i think i fixed what you were referring to. I just missed those 2 lines after i changed how it worked with input.

I'm not really sure with posting arrays. I generally do that a lot in my coding, but i havn't tried it out in my fuel projects yet.

jschreuder:
looks good to me now, will test it tonight.

jeffutter:
Any progress on getting this committed?

jschreuder:
I've been swamped lately. I promise that I'll get it into RC3 at least. The reason is that I want to go over the whole thing and make sure if I agree with the way you've solved it, I may prefer to have set_value() set a checked value when its input matches the actual value or even some other way. I still have to go over the whole thing to make sure it plays nice with array input (field names like "name[]" and "name[value]". I'll do the whole thing in one go, just have to find the time somewhere.

jschreuder:
Sorry it took me so long, but I opted to take another approach. This feels better to me, let me know when this causes trouble for you.

jeffutter:
I'm finding there is a little trouble with checkboxes with your implementation. The problem arises with what to set the value field to. I store my checkboxes in my DB as boolean values.

We'll use this field as an example: 

```
$form->add('open', 'Open', array(
    'type' => ($job and $job->id > 0) ? 'checkbox' : 'hidden',
    'value' => true));
```

If I set the value => true i end up with: `<option name="test" value>` (i don't think that's valid html is it?)... When i submit this form fuel see's it as an empty string. In my process_form function i can do is_string and if it is a string (if it's not checked it has no value). If it's a string then i can insert true or false to my DB. This works in most cases except when i need to pre-set it to false. 

If i do: 

```
$form->repopulate(array(
    'open' => true));
```

This ends up being false as the form ends up submitting open = '' which evaluates to false against the initial true. So in this case there is no way to pre-set like this.

The other option is to set value = 'true' this gives me `<option name="test" value="true">` (makes more sense to me). This works in most cases. However, the problem here is 1 from the DB != 'true' and thus my checkboxes always LOOK unchecked even if in the DB they are TRUE.

If i was writing a form in pure html i would probably have value="true" for checkboxes, thus i think we should support this in fuel and perhaps that should be the default value for a checkbox? Also there needs to be a way to compare the values against what ORM pulls from the DB, really for a boolean value it should return true/false not 1/0 and then true == 'true' would work.

jschreuder:
Reopen an issue if you want me to remember it, I'll take a look later today.

jschreuder:
This issue is a bit too vague and discussion doesn't belong here.

DBs return 1/0 for boolean, we can't help that - it's how SQL works. And using non string values and mapping them directly to the Orm model will always get you into trouble. I don't consider this an issue. If you want to talk about this use either the forums or write something up to pull-request.

PR #25, opened 2011-04-12
Date Bug

bencorlett:
Fixed bug in \Date::create_from_string() where the date produced would always be exactly one month behind the actual date. See http://dev.fuelphp.com/issues/111 for more.


jschreuder:
I'm not sure why but PHP keeps surprising me with its inconsistencies.

Pulled it, thanks for the fix!

PR #28, opened 2011-04-13
Fixed issue#116: If fopen() return file handler as FALSE, other code shou

crynobone:
Fixed issue#116: If fopen() return file handler as FALSE, other code should not try to write to file.

Reference: http://php.net/manual/en/function.fopen.php

### DIFF
--- a/classes/file.php
+++ b/classes/file.php
@@ -282,9 +282,11 @@ class File {
 			throw new \File_Exception('Invalid basepath, cannot update a file at this location.');
 		}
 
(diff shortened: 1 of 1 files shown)


WanWizard:
I don't think it's a good idea to simply ignore the update if you can't open for write.

I've committed a fix that throws an error if this situation occurs, and also to have open_file() check if the resource passed really is a resource before we attempt to flock()

crynobone:
Latest changeset only cause the application to stop working, refer to https://github.com/fuel/core/issues/31

WanWizard:
Issue #31 has just been fixed.

PR #32, opened 2011-04-14
Issue #30

### LINKED ISSUE #30: Uri::main() shoul return base URL only
**Entered by Juan Treminio on 4/13/11**

fuel/core/classes/uri.php:

```
/**
 * Gets the current URL, including the BASE_URL
 *
 * @param    string    the url
 */
public static function main()
{
    return static::create(\Request::main()->uri->uri);
}

/**
 * Gets the current URL, including the BASE_URL
 *
 * @param    string    the url
 */
public static function current()
{
    return static::create();
}
```

Uri::main() should return ONLY main URL.

Example, if my website is on http://fuel.com/, Uri::main() should return only "http://fuel.com/", but it's returning the same thing as Uri::current(), which is the base URL plus controller.

Uri::main in welcome/index is currently returning:

http://fuel.com/welcome/index/

For now I've simply extend Uri:

```
static function base()
{
    return \Config::get('base_url');
}
```

**Stuart Hutchinson responded**

According to the docs, URI:current() does the following:

```
The current method allows you to fetch the current URL, including the base URL inside HMVC pattern.
```

whereas URI::main() does the following:

```
The main method allows you to fetch the current URL, including the base URL.
```

I think this should probably be a feature if adding a function such as URI::base() to the class is desired.

eightfivelabs:
Added Uri::base() to get the base URL for uniformity of the URI class rather than manually using \Config::get('base_url'). fixes #30

### DIFF
--- a/classes/uri.php
+++ b/classes/uri.php
@@ -194,6 +194,14 @@ class Uri {
 		return static::create();
 	}
 
+    /**
+     * Gets the base URL
+     */
+    public static function base()
+    {
+        return \Config::get('base_url');
+    }
+
 	/**
 	 * @var	string	The URI string
 	 */


jschreuder:
Dan Added this: https://github.com/fuel/core/commit/a610c9274d3a9e79207b9a5a25801209e9738747

PR #41, opened 2011-04-16
Fixed bug in Uri test

huglester:
Fixed bug in Uri test. Now if we have app/config/config.php index_file set to null, false etc, it won't fail


jschreuder:
Pulled, thanks.

PR #46, opened 2011-04-18
Let Session_Db fall back on the active database group if none is provided

frankdejonge:
Let Session_Db fall back on the active database group if none is provided in app/config/session.php now 'dev' was hardcoded.

### DIFF
--- a/classes/dbutil.php
+++ b/classes/dbutil.php
@@ -48,7 +48,7 @@ class DBUtil {
 	}
 
 	/**
-	 * Creates a table.  Will throw a Database_Exception if it cannot.
+	 * Drops a table.  Will throw a Database_Exception if it cannot.
 	 *
 	 * @throws	Fuel\Database_Exception
 	 * @param	string	$table	the table name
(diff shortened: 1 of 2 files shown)


tmm1:
Hey there,

Soon after the new GitHub Merge Button feature was released today we discovered a bug in the merge procedure in certain cases. This pull request was affected by the bug, causing code to not actually get merged even though it looks from the history as though it was.

It looks like the bad merge of `config/session.php` was already fixed in dfc01ad0a3fb2dafecc7e1a4fed2d6861c53d83f. Committing this additional change will bring the `develop` branch back to its expected state:

``` diff
diff --git a/classes/dbutil.php b/classes/dbutil.php
index 38439dc..5607320 100644
--- a/classes/dbutil.php
+++ b/classes/dbutil.php
@@ -46,7 +46,7 @@ class DBUtil {
    }

    /**
-    * Creates a table.  Will throw a Database_Exception if it cannot.
+    * Drops a table.  Will throw a Database_Exception if it cannot.
     *
     * @throws  Fuel\Database_Exception
     * @param   string  $table  the table name
```

Sorry for the inconvenience. If you have any questions or need help, do not hesitate to contact [support](http://support.github.com/).

dhrrgn:
Thanks.  Love the new feature by the way.

Dan

PR #49, opened 2011-04-18
Modification to Lang class

santiagofs:
The scenario is this:
You need to show the user a message, so you add in your controller / view a lang::line method.
For each line you use, you have to update the correspondant lang file.

So I added a save functionallity to the lang class (following the pattern of the config class), so, if a line is not found in the lang file, it will be saved in the array of the default language.

I've not solved the use of this functionallity using groups or multiple files. The name of the file you want the updates to be saved is stored in the static $context variable.

I think this idea can be done in a better way, such as using groups also or set the save function to be optional, but I think it's a good idea to be evolved. I've implemented it in CI and save me a lot of time.

Please, if I'm not doing this in the proper way, let me know.

Best regards. Santiago.

### DIFF
--- a/base.php
+++ b/base.php
@@ -86,10 +86,6 @@ if ( ! function_exists('array_to_attr'))
 				$property = $value;
 			}
 
-			if (in_array($property, array('value', 'alt', 'title')))
-			{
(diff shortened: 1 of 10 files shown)


jschreuder:
We don't accept pull requests on master, can you make this again on develop?

PR #51, opened 2011-04-20
Fixed incorrect recursion condition in Arr::flatten_assoc

jsarch:
Test case:

```
$multilevel = array(
    'level1' => array(
        'level2' => array(
            'level3'
        )
    )
);
print_r( Arr::flatten_assoc($people) );
```

---

With this patch:

```
Array
(
    [level1:level2:0] => level3
)
```

---

Without this patch

```
Array
(
    [level1:level2] => Array
        (
            [0] => level3
        )
)
```

### DIFF
--- a/classes/arr.php
+++ b/classes/arr.php
@@ -40,7 +40,7 @@ class Arr {
 		foreach ($array as $key => $val)
 		{
 			$curr_key[] = $key;
-			if (is_array($val) and array_values($val) !== $val)
+			if (is_array($val))
 			{
 				static::flatten_assoc($val, $glue, false);
 			}
--- a/classes/arr.php
+++ b/classes/arr.php
@@ -24,9 +24,10 @@ class Arr {
 	 * @param	array	The array to flatten
 	 * @param	string	What to glue the keys together with
 	 * @param	bool	Whether to reset and start over on a new array
+	 * @param	int		Number of levels (array dimensions) remaining to flatten
 	 * @return	array
 	 */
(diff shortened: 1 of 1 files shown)


dhrrgn:
This was actually the intention on the flatten_array method.  It doesn't need patched, it is how it was designed.

jsarch:
The documentation should 1) be changed since the result is not a "1 dimensional associative array", and 2) include an example like above where the input array is larger than 2-dimensions.

jsarch:
The above commit keeps the intended design of flattening just one level, while also allowing for more levels to be flattened.

dhrrgn:
It actually doesn't care about the level.  The original design was so that is won't flatten down a non-keyed array. It does this because this:

``` php
Array
(
    [level1:level2:0] => foo
    [level1:level2:1] => bar
)
```

Is not helpful at all.  It is designed so that all ASSOCIATIVE arrays are flattened.  Its original purpose was so that you could access $array['level1:level2'] and loop through the values.

I think what you are trying to do is collapse the keys completely.  I think it is simply a documentation issue.

jsarch:
Gotcha. My hang-up appears to be more with the expectation that the resulting output will always be 1-dimensional, which is not always the case. Removing the snippet about 1-D and giving an example how non-keyed arrays don't "flatten" would be very, very helpful.

It also sounds like another function "flatten_completely" may be nice to add for v1.1 because while you may not think that 'level1:level2:0' is helpful, it is immensely helpful when working with OAuth parameters.

I'm okay with closing this if the docs are updated. Good luck with finalizing 1.0.

PR #59, opened 2011-04-21
Asset class - file location flexibility

santiagofs:
In the Asset class I added a method to load files in other than the predefined folders (css, js, img) as I found this division very annoying when working with js packages like jquery-ui.

The current Asset class way means you have to reorganize the files in a js package in order to use them (and to it again if you update the package).

### DIFF
--- a/classes/asset.php
+++ b/classes/asset.php
@@ -147,10 +147,11 @@ class Asset {
 			$type = $item['type'];
 			$filename = $item['file'];
 			$attr = $item['attr'];
-
+			$folder = $item['folder'];
+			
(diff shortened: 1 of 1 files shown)


dhrrgn:
Uh.  You can already change the paths for these folders.  You simply copy the asset.php config file from fuel/core/config.  Simple.

santiagofs:
Hi Dan,
The intention is to keep the css/img/js schema but provide some more
flexibility.
I was looking into the asset config file, and indeed I can change the path
for the folders but globally.

I like simplicity also, perhaps I'm missing something.

Best regards.

PR #60, opened 2011-04-21
Uri class new methods

santiagofs:
The first one returns the url without the parameters. This come handy when implementing a function across different modules or controllers which have a differente uri segment lenght. Also may be useful on a later modification to the pagination class (for a more flexible uri_segment).

The second does the same but also removing the action. This is useful when trying to know if different functions belongs to the same controller, eg: you have a menu item 'blog' which has to be 'active' on blog/item, blog/search, etc.

### DIFF
--- a/classes/uri.php
+++ b/classes/uri.php
@@ -264,6 +264,25 @@ class Uri {
 	{
 		return $this->get();
 	}
+	
+	public function without_params()
+	{
+		$segments = \Request::active()->uri->get_segments();
+		
+		$params = \Request::active()->route->method_params;
+		if(count($params)) array_splice($segments,-count($params));
+		return implode('/',$segments);
+	}
+	
+	public function without_action()
+	{
+		$segments = \Request::active()->uri->get_segments();
+		$params = \Request::active()->route->method_params;
+		if(count($params)) array_splice($segments,-count($params));
+
+		if(\Request::active()->route->action != 'index') array_splice($segments,-1);
+		return implode('/',$segments);
+	}
 }
 
 /* End of file uri.php */


dhrrgn:
Not really sure I like these.  I can see where they would be handy, but i don't know.  There are a few problems with them.  First, most methods in Uri should return a full Uri.  You can do this by sending the segment string to `Uri::create()`.

Second, there are things that are missing that are now required before accepting pull requests:
1.  Full DocBlocks on any method.
2.  Must be an accompanying pull request on fuel/docs for documenting of new/update functions.
3.  We prefer Unit Tests to be written.  This will become a strict requirement soon.

Dan

santiagofs:
On Thu, Apr 21, 2011 at 6:53 PM, dhorrigan <
reply@reply.github.com>wrote:

> Not really sure I like these.  I can see where they would be handy, but i
> don't know.  There are a few problems with them.  First, most methods in Uri
> should return a full Uri.  You can do this by sending the segment string to
> 
> ``` Uri::create()```.
> 
> Ok with the first. I was thinking about a parameter to decide whether a full
> Uri should be returned.
> 
> 
> 
> Second, there are things that are missing that are now required before
> accepting pull requests:
> 
> 1.  Full DocBlocks on any method.
> 
> Ok.
> 
> 2.  Must be an accompanying pull request on fuel/docs for documenting of
> new/update functions.
> 
> Ok. Any comments suggestions about this?
> 
> 3.  We prefer Unit Tests to be written.  This will become a strict
> requirement soon.
> 
> I'm sorry, but I need some more directions here. Is there something I can
> read about how to implement Unit Tests?
> 
> ```

Best regards.

dhrrgn:
I don't like these because simply put, they are easily creatable already.  Closing Request.

PR #79, opened 2011-04-26
External assets should be able to have an empty URI scheme/protocol

phillbrown:
If you leave off the protocol (scheme officially) from a URI, then the browser will choose the protocol used to get the page. A classic example is jQuery from the Google CDN:

//ajax.googleapis.com/ajax/libs/jquery/1.5.2/jquery.min.js

So if you're viewing the page through https it'll automatically prepend it; likewise http. This prevents the nasty dialog box you get in IE when requests are made through both secure and non-secure protocols.

### DIFF
--- a/classes/asset.php
+++ b/classes/asset.php
@@ -148,7 +148,7 @@ class Asset {
 			$filename = $item['file'];
 			$attr = $item['attr'];
 
-			if (strpos($filename, '://') === false)
+			if ( ! preg_match('|^(\w+:)?//|', $filename))
 			{
 				if ( ! ($file = static::find_file($filename, static::$_folders[$type])))
 				{


philsturgeon:
Please can you send any future pull requests to the develop branch? We follow the git-flow branching model. I'll pull this one there for now, but it could start leading to problems if done in the future.

Thanks for the fix! :)

phillbrown:
No worries, sorry about that!

Are you just working off the single master branch in the docs repository?

philsturgeon:
Everything should have a develop branch and everything should be using git-flow-style feature branches or being sent to develop. Docs seemed to be missing one but I've just pushed it up.

PR #81, opened 2011-04-27
New Arr method - Arr::filter_keys()

kbanman:
This is quite useful to me. Use it to extract only certain keys from an array.
Passing `true` as the third argument returns the array minus the specified keys.


dhrrgn:
This is awesome.  YOU WROTE TESTS!!! haha.  Pulling.

PR #86, opened 2011-04-30
More complete testing for Html::methods

ghost:
I added tests for every Html class method besides audio and build_list. The only issue I see is that there's no image bundled with the framework by default, so I cannot test that the images are directly linked. (I had to use the url 'index.php/image.png' to test.

Is there a way to reliably test that the image is directly linked? Perhaps bundle the fuel logo along with the framework somewhere so you can test against it?

### DIFF
--- a/tests/html.php
+++ b/tests/html.php
@@ -136,6 +136,174 @@ class Tests_Html extends TestCase {
 		$this->assertEquals($expected, $output);
 	}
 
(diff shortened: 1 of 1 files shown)


nerdsrescueme:
Moving this to develop. I apologize I'm used to developing alone and never had to use pull requests.

PR #94, opened 2011-05-02
Precision fix for Image_Driver

dudeami0:
Fixed an error where floats precision loss adversely affected outputted image. Known to happen when resizing image under ~2% of original size.

### DIFF
--- a/classes/asset.php
+++ b/classes/asset.php
@@ -153,7 +153,7 @@ class Asset {
 			{
 				if ( ! ($file = static::find_file($filename, static::$_folders[$type])))
 				{
-					throw new \Fuel_Exception('Could not find asset: '.$filename);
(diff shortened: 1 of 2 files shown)


nerdsrescueme:
Awesome. I was going off the main page where it says it needs to be installed. Sweet man. I can actually use this in a project I have coming up.

nerdsrescueme:
http://www.php.net/manual/en/bc.installation.php I guess it depends on whether enough default installations have enabled it? Hopefully it goes in, that way I don't need to worry about it for my photo editor project. That would be lovely.

dudeami0:
If it becomes a problem, I'll most likely check for BCMath support or make it a config option

jschreuder:
Good work on making BCMath optional, but you're not entirely up to the coding standards: all brackets must be on their own line.

PR #95, opened 2011-05-03
Added class LDAP and set it to load in the bootstrap file

axelitus:
This class let's you connect and run queries against an LDAP server.

Usage (Test Case):

ldap.php (config file into apps config folder):

```
<?php

return array(
    'domain_controllers' => array(
        0=>'db.debian.org'
    ),
    'domain_suffix' => '',
    'connection' => array(
        'port' => 389,
        'master_user' => '',
        'master_pwd' => '',
        'use_ssl' => false,
        'use_tls' => false,
    ),
);

/* End of file ldap.php */
```

welcome.php (controller):

```
<?php

class Controller_Welcome extends Controller
{

    /**
     * The index action.
     * 
     * @access  public
     * @return  void
     */
    {       
        echo "<pre>";
        echo "Initiatiting...<br />";
        echo "Loading: ";
        if (LDAP::load()) {
            echo "Loaded!<br />";
            echo "Current config: ";
            var_dump(LDAP::get_current_config());
            echo "Connecting: ";
            if (LDAP::connect()) {
                echo "Connected!<br />";
                echo "Binding: ";
                if (LDAP::bind(true)) {
                    echo "Binded! ".LDAP::get_base_dn()."<br />";
                    $filter = LDAP::build_filter(array(
                        '&', 'objectClass=debianServer', 'host=p*'
                    ));
                    echo "Query Filter: " . $filter . "<br />";
                    echo "Querying: ";
                    if(($entries = LDAP::query($filter)) == null){
                        var_dump(LDAP::error());
                    }else{
                        var_dump($entries);
                    }
                    echo "<br />Done!";
                }
                else{
                    echo "Could not bind!<br />";
                    var_dump(LDAP::error());
                }
            }
            else{
                echo "Could not connect!<br />";
                var_dump(LDAP::error());
            }
        }else{
            echo "Could not load!<br />";
            var_dump(LDAP::error());
        }
        echo "</pre>";
    }

    /**
     * The 404 action for the application.
     * 
     * @access  public
     * @return  void
     */
    public function action_404() {
        $messages = array(

            'Aw, crap!', 'Bloody Hell!', 'Uh Oh!', 'Nope, not here.', 'Huh?'
        );
        $data ['title'] = $messages [array_rand($messages)];
        // Set a HTTP 404 output header
        $this->response->status = 404;
        $this->response->body = View::factory('welcome/404', $data);
    }
}

/* End of file welcome.php */
```

This test it's supposed to return all debianServers that start it's host name with 'p'.
The LDAP::build_filter() function accepts an array params. The structure used in this array is like this:

~ For a filter with only one condition:

```
$params = array(
    'key=value'
);
```

~ For a filter with multiple conditions:

```
$params = array(
    'operator',
    'key1=value1',
    'key2=value2'
);
```

where operator is either one of the following:
& <- and
|  <- or

~ This also works with nested conditions:

```
$params = array(
    'operator',
    'key1=value1',
    'key2=value2',
    array(
        'operator',
        'key3=value3',
        'key4=value4',
        array(
            'key5=value5'
        )
    )
);
```

~ For using the NOT operator one would prepend the key=value pair with a !:

```
array(
    '!key=value'
);
```

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -123,6 +123,7 @@ Autoloader::add_classes(array(
 	'Fuel\\Core\\Input'						=> COREPATH.'classes/input.php',
 	'Fuel\\Core\\Lang'						=> COREPATH.'classes/lang.php',
 	'Fuel\\Core\\Log'						=> COREPATH.'classes/log.php',
+	'Fuel\\Core\\LDAP'						=> COREPATH.'classes/ldap.php',
 	'Fuel\\Core\\Migrate'					=> COREPATH.'classes/migrate.php',
 	'Fuel\\Core\\Model'						=> COREPATH.'classes/model.php',
 	'Fuel\\Core\\Output'					=> COREPATH.'classes/output.php',
--- a/classes/ldap.php
+++ b/classes/ldap.php
@@ -1,0 +1,364 @@
+<?php
+/**
+ * Fuel is a fast, lightweight, community driven PHP5 framework.
+ *
+ * @package    Fuel
+ * @version    1.0
+ * @author     Fuel Development Team
+ * @license    MIT License
+ * @copyright  2010 - 2011 Fuel Development Team
+ * @link       http://fuelphp.com
+ */
+namespace Fuel\Core;
+
+class LDAP
+{
+	const LDAP_FOLDER = 'OU';
+	const LDAP_CONTAINER = 'CN';
+	const DEFAULT_PORT = 389;
+	const DEFAULT_SSL_PORT = 636;
+	private static $_config;
+	private static $_connection;
+	private static $_binding;
+	private static $_base_dn;
+
+	private function __construct() {}
+
+	private static function clean($erase_config = false) {
+		static::$_connection = null;
+		static::$_binding = null;
+		static::$_base_dn = null;
+		if ($erase_config) {
+			static::$_config = null;
+		}
+	}
+
+	public static function load($reconnect = false, $config = null) {
+		$response = false;
+		if (static::ldap_supported()) {
+			static::disconnect();
+			if ($config === null || ! is_array($config) || empty($config)) {
+				static::$_config = \Config::load('ldap', true);
+			} else {
+				static::$_config = $config;
+			}
+			$response = true;
+			if ($reconnect) {
+				$response &= static::connect();
+			}
+		} else {
+			throw new \Fuel_Exception('LDAP is not supported.');
+		}
+		return $response;
+	}
+
+	public static function get_current_config() {
+		return static::$_config;
+	}
+
+	public static function connect($reload_config = false, $config = null) {
+		$response = false;
+		if (static::$_connection === null || ! is_resource(static::$_connection)) {
+			if ($reload_config == true || static::$_config === null || ! is_array(static::$_config) || empty(static::$_config)) {
+				static::load(false, $config);
+			}
+			$domain_controller = trim(static::_random_domain_controller());
+			if ($domain_controller !== "") {
+				$host = $domain_controller;
+				if (isset(static::$_config ['connection'] ['use_ssl']) && static::$_config ['connection'] ['use_ssl'] == true) {
+					$host = "ldaps://" . $host;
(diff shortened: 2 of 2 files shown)


axelitus:
Output should be:

```
Initiatiting...
Loading: Loaded!
Current config: array(3) {
  ["domain_controllers"]=>
  array(1) {
    [0]=>
    string(13) "db.debian.org"
  }
  ["domain_suffix"]=>
  string(0) ""
  ["connection"]=>
  array(5) {
    ["port"]=>
    int(389)
    ["master_user"]=>
    string(0) ""
    ["master_pwd"]=>
    string(0) ""
    ["use_ssl"]=>
    bool(false)
    ["use_tls"]=>
    bool(false)
  }
}
Connecting: Connected!
Binding: Bound! dc=debian,dc=org
Query Filter: (&(objectClass=debianServer)(host=p*))
Querying: array(13) {
  ["count"]=>
  int(12)
  [0]=>
  array(2) {
    ["count"]=>
    int(0)
    ["dn"]=>
    string(35) "host=paer,ou=hosts,dc=debian,dc=org"
  }
  [1]=>
  array(2) {
    ["count"]=>
    int(0)
    ["dn"]=>
    string(40) "host=pergolesi,ou=hosts,dc=debian,dc=org"
  }
  [2]=>
  array(2) {
    ["count"]=>
    int(0)
    ["dn"]=>
    string(37) "host=piatti,ou=hosts,dc=debian,dc=org"
  }
  [3]=>
  array(2) {
    ["count"]=>
    int(0)
    ["dn"]=>
    string(35) "host=peri,ou=hosts,dc=debian,dc=org"
  }
  [4]=>
  array(2) {
    ["count"]=>
    int(0)
    ["dn"]=>
    string(39) "host=penalosa,ou=hosts,dc=debian,dc=org"
  }
  [5]=>
  array(2) {
    ["count"]=>
    int(0)
    ["dn"]=>
    string(37) "host=powell,ou=hosts,dc=debian,dc=org"
  }
  [6]=>
  array(2) {
    ["count"]=>
    int(0)
    ["dn"]=>
    string(39) "host=pescetti,ou=hosts,dc=debian,dc=org"
  }
  [7]=>
  array(2) {
    ["count"]=>
    int(0)
    ["dn"]=>
    string(41) "host=praetorius,ou=hosts,dc=debian,dc=org"
  }
  [8]=>
  array(2) {
    ["count"]=>
    int(0)
    ["dn"]=>
    string(39) "host=paganini,ou=hosts,dc=debian,dc=org"
  }
  [9]=>
  array(2) {
    ["count"]=>
    int(0)
    ["dn"]=>
    string(41) "host=pettersson,ou=hosts,dc=debian,dc=org"
  }
  [10]=>
  array(2) {
    ["count"]=>
    int(0)
    ["dn"]=>
    string(38) "host=poulenc,ou=hosts,dc=debian,dc=org"
  }
  [11]=>
  array(2) {
    ["count"]=>
    int(0)
    ["dn"]=>
    string(38) "host=porpora,ou=hosts,dc=debian,dc=org"
  }
}

Done!
```

dhrrgn:
I don't think that an LDAP class belongs in the core.  So few people will use it that there is no benefit.

What does everyone else think?

buritica:
I agree, at least in my cases.. I would not need it..

zacwasielewski:
Agreed. This would be great as a separately-hosted extension, but its audience is too small to be in the core.

tomschlick:
I think this would be best suited as a package. Easy to install for those who need it but out of the way for those who dont.

dilantha:
Personally I've never used LDAP I don't think it belongs in core.

DregondRahl:
I think it would make a great package, not everyone uses LDAP. But as a package it could be useful to those who do, also it makes updating for bug changes or more features easier if its separate from the core.

okhayat:
I had to use LDAP only twice in my life, and doubt most people would need it. Maybe just provide it as a Driver for Auth package.

philsturgeon:
Agreed. This is a great packages or class, which will be a perfect candidate for the new package system. You'll be able to sign up, add it and have it searchable and installable via the web interface and through the oil command.

Keep an eye out for its release. It's gonna be awesome :)

[Closing as we have a few -1 votes now]

Issue #84, opened 2011-04-29
FUEL::find_file + Namespaces


axelitus:
I just find a bug in FUEL::find_file function on line 259:
https://github.com/fuel/core/blob/develop/classes/fuel.php#L259

The class caches the found path so that it has it on it's records. But now we have modules, so the directory structure can 'repeat' itself.
If I have two modules: mod1 and mod2 having this structure (I wont put all folders as they are not need to show the point):

```
- mod1
    |- config
         |- config.php

-mod2
    |- config
         |- config.php
```

and I run:

```
$mod1_config = \Config::load('mod1\\config', true);
$mod2_config = \Config::load('mod2\\config', true);
```

then both vars will have the same contents, because of caching not using the namespace. Let me clarify this (fuel.php L259):

```
$path = $directory.DS.strtolower($file).$ext;
var_dump($path);
```

will output in both cases:

```
string(17) "config\config.php"
```

if then, on line 261 we check for the cached path:

```
if (static::$path_cache !== null and array_key_exists($cache_id.$path, static::$path_cache))&#123;
    return static::$path_cache[$cache_id.$path];
}
```

the array key we are checking ($cache_id.$path) would be in both cases something like this:

```
string(49) "d41d8cd98f00b204e9800998ecf8427econfig\config.php"
```

which is wrong because they are two different paths!!! I think the namespace should be somewhere involved in assigning a key for the cached path.

Reference: http://fuelphp.com/forums/posts/view_reply/1271

axelitus:
I've found a solution for this issue which I think doesn't interfere with how things are supposed to work:

```
public static function find_file($directory, $file, $ext = '.php', $multiple = false, $cache = true)
{
    $cache_id = '';
    $ns = '';

    $paths = static::$_paths;

    // get extra information of the active request
    if (class_exists('Request', false) and $active = \Request::active())
    {
        $cache_id = md5($active->uri->uri);
        $paths = array_merge($active->paths, $paths);
    }

    // the file requested namespaced?
    if($pos = strripos(ltrim($file, '\\'), '\\'))
    {
        $file = ltrim($file, '\\');
        $ns = ucfirst(substr($file, 0, $pos));

        // get the namespace path
        if ($path = \Autoloader::namespace_path('\\'.$ns))
        {
            // and strip the classes directory as we need the module root
            $paths = array(substr($path,0, -8));

            // strip the namespace from the filename
            $file = substr($file, $pos+1);
        }
        $ns .= DS;
    }

[...]

PR #97, opened 2011-05-05
Fixed Autoloader::load() function + Config::save() to work with modules

axelitus:


WanWizard:
As been discussed at issue #84, testing for a backslash to determine the module is a bad idea on Windows systems, where the backslash is the default DIRECTORY_SEPARATOR (DS).

As said there, we need to think about a different way of defining the module, or have it fall through if no module namespace can be found.

axelitus:
[In this post](http://fuelphp.com/forums/posts/view_reply/1402) I propose using a pipe to define a namespace/module

Despite of this still pending issue, the Autoloader did a strtolower() to the namespace, which it is not done anywhere else; so if the find_file() issue is fixed in the future, this still needs to be fixed...

WanWizard:
I have decided to opt for fall-through. If backslash detection doesn't lead to a valid namespace, the file will be looked for in the global namespace.

WanWizard:
Wasn't happy with the fall-through, as it would make folders with the same name as a defined namespace unaccessable (since the namespace path would always be selected).

You now have to use the double colon to access a file in a namespace ('module::folder/file').

axelitus:
I get error:

```
ErrorException [ Warning ]: pathinfo() expects parameter 1 to be string, array given

COREPATH/classes/config.php @ line 124:

123:
124:    $path = pathinfo($path);
125:
```

axelitus:
<del>I put:
    else
    {
        $path = $path[0];
    }
after [Line 113](https://github.com/fuel/core/blob/develop/classes/config.php#L113) in the config.php file</del> Don't do this... it's not a real solution... instead read ahead...

axelitus:
Ok did some more digging and found out that the 'problem' is that when the path is cached it is returned as an array from the [line 285](https://github.com/fuel/core/blob/develop/classes/fuel.php#L285):

```
return static::$path_cache[$cache_id.$path];
```

in the find_file() function. Shouldn't it be:

```
return static::$path_cache[$cache_id.$path][0];
```

or look for the $multiple param?

WanWizard:
find_file() should maintain a different cache for single and multiple file requests. I've just committed a fix.

axelitus:
Fine, I'll check it tomorrow... that should do I think, but I'll do some testing tomorrow...

I'm not sure that will be enough... I'm not decided yet hehe

axelitus:
The last commit breaks again...

I get the "Fatal error: Exception thrown without a stack frame in Unknown on line 0" message...

I've already downloaded a clean copy of the develop branch to make sure it wasn't some code of mine. I'll dig into it to see what's happening...

axelitus:
Found it!
As $cache_id is initialized either with 'M.' or 'S.', on [line 261](https://github.com/dudeami/core/blob/75966039beb6926e43bc4a957323c2789028073e/classes/fuel.php#L261) we cannot check for an empty $cache_id as it won't be empty.

I changed it for

```
if (strlen($cache_id) == 2)
```

and everything is working again... well, at least it shows me the Exception page. I'll keep testing...

jschreuder:
I just changed it to `if (empty($paths))` should work fine, @WanWizard can you check this?

axelitus:
That would work except for the absolute path part which doesn't modify the $paths array, only the $cache_id...
it could be `if(empty($paths) && empty($found))` though...

WanWizard:
decided to revert some of the changes, and only add the multiple/single cache identifier at the end.

axelitus:
Seems to be working alright... and it's just a much cleaner solution!

WanWizard:
Ok, thanks for the feedback.

axelitus:
Thank you for the hard work you're all doing and the fantastic framework you're building...

I think we can consider this issue as closed now after all headaches it gave... :)

PR #106, opened 2011-05-08
Uri::create() method change '&' to '&amp;'

KaneCohen:
In Method Uri::create();
When creating GET variables/values is used symbol '&' which isn't very secure/good for web validation.


KaneCohen:
Hmm, Html::anchor() creates <a> links which in some cases not acceptable. Like generating Uri to files/images. Maybe add extra argument?

PR #108, opened 2011-05-08
DBUtil::create database charset now defaults to 'null'

frankdejonge:


frankdejonge:
Made create_database charset more flexible by allowing simple charsets (like: "utf8") and more specific (like: "uft8_unicode_ci")

PR #110, opened 2011-05-10
Bug Fix for REST Controller

greenwebdev:
On line 103 of core/classes/controller/rest.php, the call to response->set_header was only passing a single parameter. With this change there is no longer an error.


greenwebdev:
Sorry guys! I hadn't seen this had been fixed in the dev branch, just read it on the forums.

jschreuder:
Also always pull-request on the develop branch, we will never accept anything to the master branch and will very often deny out of hand when this is done.

PR #112, opened 2011-05-10
Fixed to correctly reference CamelCased namespaced classes in Autoloader

axelitus:
Fixed to correctly reference CamelCased namespaced classes so they can be loaded.

Everywhere else the namespace is only treated with ucfirst() function, so when we have CamelCased namespaces and the autoloader strtolower() the namespace name, it would never be found as it is not the same thus giving the error: \CamelCasedNamespace\MyClass not found.

### DIFF
--- a/classes/autoloader.php
+++ b/classes/autoloader.php
@@ -233,7 +233,7 @@ class Autoloader {
 		else
 		{
 			// need to stick the trimed \ back on...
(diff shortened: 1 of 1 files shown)


philsturgeon:
Technically that's PascalCase, but the fix is perfectly valid. ;) Thanks.

axelitus:
Hehe you're right! PascalCase... but yes... it broke some modules I have PascalCased :P

PR #131, opened 2011-05-19
Fix namespace for the Exception

MaherSaif:
This is a fix for the namespace of the exception which will be thrown when no files selected while uploading.

### DIFF
--- a/classes/upload.php
+++ b/classes/upload.php
@@ -430,7 +430,7 @@ class Upload {
 		// anything to save?
 		if (empty($files))
 		{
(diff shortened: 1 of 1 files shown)


jschreuder:
This has already been fixed in the develop branch: https://github.com/fuel/core/blob/develop/classes/upload.php#L431

Pull-requests to master are **never** accepted, only to develop.

MaherSaif:
Thanks for your tips.

## 

Maher Muhammad Saif
Freelance Web Developer

PR #138, opened 2011-05-22
Fixed a problem with Rest_Controller,

elementagency:
just changed a . to a , as it seams the require set_header needs an array of key value not a string.


jschreuder:
We don't accept pull requests on master, if you check you'll see this has been fixed in the dev branch already. We'll release RC3 soon with those fixes.

PR #139, opened 2011-05-23
Added alternator() method to Str class

patie:


classes/str.php
```diff
@@ -232,6 +232,28 @@ class Str {
...
+                     /**
+ * Alternator
+ *
+ * @access  public
+ * @param   string (as many parameters as needed)
+ * @return  string
+ */
+            public static function alternator()
+            {
+                static $i;
```
jschreuder: It would be better to give it an initial value (= 0)

classes/str.php
```diff
@@ -232,6 +232,28 @@ class Str {
...
+
+                if (func_num_args() == 0)
+                {
+                    return '';
+                }
+                
+                $args = func_get_args();
```
jschreuder, line 253: There's a problem here: if I use 2 alternators in the same loop they will both increment I and skip elements.

A better way would be to do something like this:

```
static $args = array();
$args_in = func_get_args();
if (count($args_in) == 0)
{
    return '';
}

$key = md5(serialize($args_in));
if ( ! isset($args[$key]))
{
    $args[$key] = $args_in;
}

return $args[$key][($i++ % count($args[$key]))];
```

But there may be some more efficient way to do this (md5 hashing may be a bit overkill)

patie:
"my way" work like a charm for me

ex.:

```
foreach ( $rows as $r )
{
    echo '<tr class="'. \Str::alternator('odd', 'even')  . '">';
    echo '<td>'.$r.'</td>';
    echo '</tr>';
}
```

jschreuder:
Now try something like this:

```
foreach ($rows as $r)
{
    echo '<tr class="'. \Str::alternator('odd', 'even') . '">';
    echo '<td>'.\Str::alternator('Odd row', 'Even row').'</td>';
    echo '</tr>';
}
```

All rows will have the class "odd" and contents "Even row".

jschreuder:
I made a mistake in the code above - I forgot to use different counters. Remove the $i variable and do something like this probably:

```
static $counters = array();
$args = func_get_args();
if (count($args_in) == 0)
{
    return '';
}

$key = md5(serialize($args));
if ( ! isset($counters[$key]))
{
    $counters[$key] = 0;
}

return $args[($counters[$key]++ % count($args))];
```

The truth is btw that the only "right" way to implement an alternator is in its own class and use objects. But I consider that overkill for the core, and better to have a "simple" one included here.

jschreuder:
Lol, sorry for all the comments but you just gave me another idea: have the alternator return a closure:

```
public static function alternator()
{
    $args = func_get_args();
    return function ($next = true) use ($args)
    {
        static $i = -1;
        $next and $i++;

        return $args[$i % count($args)];
    };
}
```

This works as follows:

```
$alt = Str::alternator('one', 'two');

echo $alt(); // returns 'one'
echo $alt(); // returns 'two'
echo $alt(); // returns 'one'
echo $alt(false); // returns 'one'
```

The advantage of this is that you can use many different alternators at the same time and reuse without incrementing:

```
$alt = Str::alternator('odd', 'even');
$alt2 = Str::alternator('Odd row', 'Even row');
foreach ($rows as $r)
{
    echo '<tr>';
    echo '<td class="'.$alt().'">'.$alt2().'</td>'; // outputs 'odd'/'Odd row' or 'even'/'Even row'
    echo '<td class="'.$alt(false).'">second column</td>'; // outputs the same as $alt() above
    echo '</tr>';
}
```

dhrrgn:
I don't like this.  Basically because I hate setting variables in Views, and sending it from the controller would just be dumb.

I prefer a modified version of @jschreuder:

``` php
public static function alternator()
{
    static $counters = array();
    $args = func_get_args();
    if (count($args) == 0)
    {
        return '';
    }

    $key = implode(',', $args);
    ! isset($counters[$key])) and $counters[$key] = 0;

    return $args[($counters[$key]++ % count($args))];
}
```

This is untested, but should work out of the box.  I changed the md5() to an implode with a glue of ','.  I did the comma because if I used empty, and just concatenated them all together, it would cause weird issues where it would use the same iterator:

``` php
foreach ($rows as $r)
{
    echo '<tr class="'. \Str::alternator('11', '22') . '">';
    echo '<td>'.\Str::alternator('112', '2').'</td>';
    echo '</tr>';
}
```

If they were just concatenated both keys would be "1122", now they are "11,22" and "112,2".

jschreuder:
> I don't like this. Basically because I hate setting variables in Views, and sending it from the controller would just be dumb.

I get your point but I think it's a bit over-principled in this case. When you need to reuse your alternator output (for instance a td class) you still need to assign it to a var. This is purely about presentation and does belong fully in the view, as such it's not much different from requesting a string output from \Str::alternator().

Now there might be a point made about usage of this in a templating language, but that should be where ViewModels come in and you assign the alternator closure to a template callback.

patie:
omg ok forget it.. i need only basic alternator for what I was used in CI.. i write a method in my (non-core) class..

jschreuder:
Haha, don't worry - whatever it is we'll write & document it ourselves now once we decide.

jschreuder:
@dhorrigan final word on this? I've said what I think about it, I'll leave the final discission up to you.

PR #146, opened 2011-05-25
Provide the ability to specify a directory of shared packages to allow shared and project specific packages

phil-lavin:
We saw this as a necessity for many organisations who wish to use Fuel as a development platform for many sites. They would ideally like to have a set of tried and tested shared packages which any developer can use in any project. They would also like to have packages local to a particular project. Some local packages would, at the end of the development on a particular project, be put into the shared directory.

This also provides slightly nicer handling of packages not being found. Currently, it just gets a PHP when it fails to include a package.

This is twinned with a commit to phil-lavin/fuel which I have done a pull request for.

### DIFF
--- a/classes/fuel.php
+++ b/classes/fuel.php
@@ -401,7 +401,20 @@ class Fuel {
 	{
 		if ( ! is_array($package))
 		{
-			$package = array($package => PKGPATH.$package.DS);
+			$input = $package;
+
+			if (is_dir(PKGPATH.$package.DS))
+			{
+				$package = array($package => PKGPATH.$package.DS);
+			}
+			elseif (defined('SHAREDPKGPATH') and is_dir(SHAREDPKGPATH.$package.DS))
+			{
+				$package = array($package => SHAREDPKGPATH.$package.DS);
+			}
+			else
+			{
+				throw new \Exception("Package {$input} not found");
+			}
 		}
 		foreach ($package as $name => $path)
 		{


jschreuder:
Hmm, I'm not sure about this. Adding config options (or constants in this case) shouldn't be: I want something extra, so lets add a config option. In this case I think that if we decide this is needed we should allow for an array of paths to be set instead of just one. Otherwise we start supporting this and next month someone comes along who needs 3 package paths and adds another option.

I think it might be a good idea to allow an array of paths, would be consistent with that we allow multiple mododule paths.

phil-lavin:
I agree- that was the original plan however you can't put arrays into constants. Can you suggest a better way that this might be implemented and I'll have a crack at doing that.

jschreuder:
There are ways around that, for example imploding it in the constant using a semi-colon separator. But first we'll have to decide on whether it goes in and which way would be best.

Just going to poke the others to look into this: @dhorrigan @philsturgeon @WanWizard

phil-lavin:
Cool, thanks

dhrrgn:
This highly unnecessary.  You can set your own constant in index or bootstrap and use that to load them.  Not adding this to core.  Adds overhead for something that is already possible...just requires a little bit more code on your end.  This goes against one of our philosophies: Make it simple, but not at the expense of performance or complexity.  This adds both a performance overhead (albeit very minimal) and more complexity.

PR #151, opened 2011-05-25
Fixed label not set on multi field template

leo-nard:


jschreuder:
Almost there, but I think the label tag within the {fields}{/fields} tags should be renamed to something like {field_label} (I'm open to suggestions). Right now it's a bit ambiguous because it's the same tag twice with different meanings.

leo-nard:
Good point. What about {value_label}?

jschreuder:
Hmm, I don't really have a preference. It's within the field tags belonging to what is then called a {field} but it also refers to the value so it's a bit of a wash.

Mitchell64:
I tested it, and it's a big improvement, but I think the required_mark is missing when printing.

leo-nard:
@Mitchell64
That is because the {required} tag is placed within the {fields} tags. It must be placed outside of it. I will fix that, along with the label naming issue, tonight.

leo-nard:
Maybe renaming the label tag that describes the group of fields is a better solution. I was thinking of {group_label} or {legend}. Legend is the actual correct html tag instead of label for this situation if I'm correct. Besides, for what it's worth, it's better bc because the {label} tag within the {fields} tag already works.

jschreuder:
Good point, though I think group_label is easier to understand.

leo-nard:
I think something went wrong with my commit/push. Forget the "Merge remote branch upstream/develop"

jschreuder:
Nothing wrong, that's just your local branch merging with the latest commits.

Mitchell64:
I just tested the new code, and I have just a couple of suggestions:
1. Using the group_label, everything seems ok if the field is required. However if the field is not required, then the {required} text is printed out. So I would suggest a small change to the function field_template(). At the code:
   
   if ($required_mark)
   {
       $template = str_replace('{required}', $required_mark, $template);
   }

Add an else clause to remove the {required} text:

```
else
{
    $template = str_replace('{required}', '', $template);
}
```
1. The 2nd suggestion is to change the default value of 'field_template' from: "\t\t\t{label} {field}\n"  to: "\t\t\t{label} {required }{field}\n" so that if the field is required, the required_mark will be printed.

Other than that, it looks really good.

leo-nard:
1. Good point. I would suggest changing the following

``` php
$required_mark = $required ? $this->get_config('required_mark', null) : null;
```

in

``` php
$required_mark = $required ? $this->get_config('required_mark', "") : "";
```

and removing the if statement for the required tag.
1. Shouldn't the {required} tag be placed outside the {fields} tags?

If you wan't you can fix this and add a pull request? Otherwise I will do it tomorrow.

Mitchell64:
I'm just about to leave for a few days of vacation, and I'm not sure about your question whether the {required" tag should be outside the {fields{ tag. Feel free to make the changes. I'll do more thinking and testing next week. Thanks.

leo-nard:
Have fun! 

Do you have an example where it should sit in between the {fields} tags?

jschreuder:
@leo-nard

Not sure what the advantage is of using `""` over `null`, use single quotes though if there is any. Agreed on removing the if requirement.

I'm not sure where it would be needed or usefull, but there's no reason not to support it. It's already supported in the develop branch though. I only just realized something else: in all cases the field tag should be the last one replaced or you might get into trouble using the Form class for editing templates using similar tags. Thus the order of both the single and multiple fields parsing needs changing.

leo-nard:
@jschreuder
Never realized null evaluates to an empty string.

I don't fully understand your last point about the order of parsing the {field} tag. Do you have an example?

jschreuder:
@leo-nard

There's something to be said for using an empty string instead, personally I don't care because PHP doesn't.

As for my last point: let's say I want to edit a page with some templating technique like mustache or simpletags and I'm editing a page to which an array `$label` is pushed and a variable called `$required`:

```
Our company has the following labels:
<ul>
{label}
    <li>{name}</li>
{/label}
</ul>

<p>And we have the following requirements:</p>
{required}
```

Now if I were to edit a page with contents like that the code would first put this in my textarea when building the field and using `str_replace()` to put it into the form template. After that the `{required}` and `{label}` tags inside my area will be parsed as well inside my textarea as they would be in the form template in which they were placed. If you insert the field last there should never be a problem with this because all form template tags have been replaced already.

leo-nard:
I understand it now i think :). You mean there are situations that similar tags (as are used in the form template) are placed in the builded field which can be replaced by the field_template function. In the most simplest way the solution would be something like this?:

``` php
$template = str_replace(array( '{label}', '{required}'), array($label, $required_mark), $template);
$template = str_replace('{field}', $build_field, $template);
```

Only with this you get the same problem if you extend the field_template function. So both should be in different functions?

jschreuder:
Yeah, that was my original suggestion - we must take care to always put the field in last. And that's  enough as far as I'm concerned, if someone extends the field_template method they'll have to work around this problem themselves. It's not important enough (and easily solved by calling the parent method after doing your own stuff) to build a web of replaceable methods for.

leo-nard:
Ok. Just added fixes for the issues we discussed above. https://github.com/fuel/core/pull/181

Mitchell64:
I finally had some time to do further testing on printing of form fields, and I noticed one small issue. When printing check boxes and radio buttons, the main label has a <label> tag associated with it. For check boxes and radio buttons, <label> tags are only associated with the boxes and buttons. Probably easier with an example, so here's the generated code for a group of check boxes:

```
<label for="checkbox-test">Checkbox Test</label>

<input type="checkbox" value="3" id="checkbox-test_0" name="checkbox_id[0]" >
<label for="checkbox-test_0">Three</label>          

<input type="checkbox" value="4" id="checkbox-test_1" name="checkbox_id[1]" >
<label for="checkbox-test_1">Four</label>
```

The first label tag should not be there. Same issue for radio buttons. Other than that, the printing is working great.

PR #163, opened 2011-05-28
Docs and Image config discrepancy.

simonberube:
The Image class documentation states that GD is the default image driver. However, the config file clearly has ImageMagick set. From my testing, GD gets used anyway. I have to say that this is entirely confusing. I wanted to use a feature exclusive to ImageMagick but it keeps on using GD.


jschreuder:
Another mess, can't accept this.

PR #164, opened 2011-05-28
Minor EOF comment change

simonberube:


jschreuder:
This is a mess, don't pull-request on master - we only accept on develop. And check you pull-requests, something like this we'll always close without looking at it. I did look at yours though, and pull-requesting for a newline at the end of a file is kinda useless.

johnnyfreeman:
@jschreuder I noticed this as well. I'm curious though, where should a simple error such as this be addressed? By email? Twitter? A pull-request seems pretty appropriate to me.

Is there some guidelines for contributing to Fuel. What's considered appropriate, or rather, what would merit you looking at a request before closing it. I'm new to making pull requests (and have not yet made one) but I don't want to waste my time making a pull-request that will just be discarded. Thanks.

jschreuder:
@johnnyfreeman check out the [contribute page](http://fuelphp.com/contribute) and look at this pull-request
1. it's on master and we will never accept anything on master, I'm not even allowed to push to master - only the project lead does that.
2. a pull-request must be clean and only contain commits belonging to that specific feature. This looks like it branches off develop being pushed to master which causes this complete mess.
3. Removing a newline on the end of a file isn't a bugreport or anything usefull, I have better stuff to do than checking individual commits about a single newline you don't like.

Bugreports & pull-requests are addressed and accepted here on a daily basis, but those that are a mess are rejected out of hand. This is where they belong, but we shouldn't have to figure out what you're trying to ask or contribute. Be clear, be clean, follow the coding guidelines and check your pull-requests before hitting send because a mess like this should be obvious to anyone.

[edit] sorry, I mixed up pull-requests in my memory - this one was a good one because it did actually correct something, it was another pull-request that only removed a newline. But points 1 & 2 still stand and are individually already reason enough to reject.

johnnyfreeman:
Ok, thanks for the link, and for the clarification. So just to be clear: If this person had actually sent the pull-request to the development branch, it would have been accepted? In other words, is sending a pull request for something small (like changing /\* End of file config/asset.php _/ to /_ End of file config/rest.php */) a completely valid and good use of the pull request system?

I appreciate you taking the time for this. I was just thrown off by your comment about the pull-request being useless (which you have stated that it was a mix up, so no harm done).

jschreuder:
Something small like that is valid, though if you do I'd prefer it if you include it with something else or do a full check for mistakes like these (makes the commitlog a bit unreadable when we get too many of these small non-functional commits).

About the useless comment: I think I mixed up or looked at the wrong thing because he does change the comment contents here. If it had been something like I described (removing a newline) it would actually have been useless, but in this case it wasn't and I apologise to @simonberube for that comment.

johnnyfreeman:
Thanks for clearing that up. :)

PR #166, opened 2011-05-29
Removed the need to set $url to an instance of the URI class.

tarnfeld:


tarnfeld:
sorry, stupid github fail.

PR #167, opened 2011-05-29
removed the instantiation of a Uri class just to cast it back to a string

tarnfeld:
I had a problem where calling `Response::redirect('/auth/login')` when being on the url '/bla' would endup taking me to 'bla/auth/login' because the Uri class was stripping off the leading slash causing it to turn into a relative url, instead of absolute. There wasnt really a need to turn $url into a URI object just to cast it back to a string so removed that.

### DIFF
--- a/classes/response.php
+++ b/classes/response.php
@@ -90,11 +90,6 @@ class Response {
 
 		$response->status = $redirect_code;
 
-		if (strpos($url, '://') === false)
-		{
-			$url = Uri::create($url);
-		}
-
 		if ($method == 'location')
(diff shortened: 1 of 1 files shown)


jschreuder:
As discussed on IRC, I think this is the wrong approach.

If you want support for paths going from the main domain (for installs not in the root) you might add that to the Uri::create() method and we'll have to discuss it, but I'm open to that. Removing support for short form fuel uri's won't happen, there are far more people for whom this will be usefull than for whom this won't be.

PR #178, opened 2011-06-02
Cleaned up Fuel::init, now all settings use the Config class instead of the config array.

frankdejonge:


frankdejonge:
Will put this on another branch so I can continue developing.

PR #199, opened 2011-06-08
Patch for Image Class

johnnyfreeman:
Sorry to submit this again but it was brought to my attention that the arrays would not be merged with my first two commits, so the last commit is the only one worth your time to look over. Thanks.

### DIFF
--- a/classes/image.php
+++ b/classes/image.php
@@ -48,6 +48,10 @@ class Image {
 	public static function factory($config = array(), $filename = null)
 	{
 		!is_array($config) and $config = array();
+		if (empty($config)) {
+			Config::load('image', 'image');
+			$config = Config::get('image');
+		}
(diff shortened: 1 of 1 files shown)


jschreuder:
I still have to check it out, but on first glance you should use `\Config::get('image', array())` that way there will still be an array to merge if no image config file is found.

johnnyfreeman:
Good catch. It would throw an error if NULL is returned. Do I have to make another pull request or is there a way to make a change to this one? Thanks.

jschreuder:
As far as I know it'll be attached to this one as long as it's the same branch and you don't open another pull-request in the meantime (not sure though, we'll see). If not, just let us know by comment here - we have other ways of pulling.

johnnyfreeman:
Thanks for the help jschreuder!

PR #202, opened 2011-06-09
Moved temp image path outside of substr(). No more truncated absolute paths.

simonberube:
Moved temp image path outside of substr(). Absolute paths were being truncated to 32 characters, resulting in attempted placement of temp images in folders where they should not be or could not be written to.


WanWizard:
We do not accept pull request for any branch other than develop.

PR #204, opened 2011-06-09
Str::bytes_readable

julesjanssen:
Not sure if this belongs in Str.
Let me know if it's okay and i'll update the docs for it.

### DIFF
--- a/classes/str.php
+++ b/classes/str.php
@@ -172,14 +172,10 @@ class Str {
 	  * @param	int		the number of characters
 	  * @return string	the random string
 	  */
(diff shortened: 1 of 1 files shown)


jschreuder:
You're right it doesn't belong in the Str class, that's about working with strings - not formatting output.

There's not (yet?) a place for a method like this, this is about formatting numbers to human readable information. I could imagine a class that would do more of these (might be a better place for a method like `Date::time_ago()` than the Date class), but on the other hand I don't think it's important enough to include in the core. Would be great though if you would release this as a standalone class, especially once we launch our sharing service for 3rd party libs & packages.

Issue #216, opened 2011-06-15
DB: MySQL connection.php does not have a port number option, though docs say it does


markbao:
Database: MySQL connection.php doesn't have a port number option, even though the docs say they do.

Doc page:
http://fuelphp.com/docs/classes/database/introduction.html

I am looking here:
https://github.com/crynobone/fuelphp/blob/master/fuel/core/classes/database/mysql/connection.php

@ commit 205309f478628ada3603

Temporary workaround is to set the hostname to "hostname:port" like "localhost:3306".

I'll try to submit a patch if I have a chance.

WanWizard:
Bug has been fixed for both the MySQL and  MySQLi drivers

PR #219, opened 2011-06-15
Issue #216 : Port option added to mysql connection

ghost:
Added ability to add a port to the db config array. Submitting a patch to fuel/fuel to include it. This was previously possible by simple appending it to the hostname (hostname:3306), but this may be better for clarity sake?


WanWizard:
Has been fixed on both the MySQL and MySQLi drivers.

Issue #220, opened 2011-06-15
Response header isn't set?


huglester:
Today I've been playing with headers, and seems like:

\Response::set_header('Content-type', 'image/png'); - this still gives the output in: text/html

while replacing it with: header("Content-type: image/png"); - works nicely

If I can give more info, just let me know

cheers,
huglester

WanWizard:
That won't work, because the static call will not update the current request.

Try

```
$this->response->set_header('Content-type:', 'image/png');
```

instead, which will update the headers of the current response object.

The method is not defined as static, but I've noticed the docs still mention the static methods. Please add this as an issue to fuel/docs.

huglester:
WanWizard, no luck with
`$this->response->set_header('Content-type:', 'image/png');`
too. 
only using header("Content-type: image/png"); I manage to set up headers.

the status part works well, like; $this->response->status = 304;

WanWizard:
Odd. Where exactly do you use this? If I add that line to the welcome controller, I get a download window...

huglester:
Hey again,
it is really really strange...  I was trying to work with this headers from controller, from module/controller - still no luck

I also tried downloading fresh Fuel, died RC-3 and DEV version - still no luck..

when using fresh install, I just edited welcome controller, index method and changed it to: 
```        {
        $this->response->set_header('Content-type:', 'image/png');
//        header("Content-type: image/png");
//              echo 'test1';
                $this->response->body = 'test2';
        }

```
while using $this->response->set_header('Content-type:', 'image/png'); - I see the text, smth really does not work here, I tried to debug but still no luck :( 
```

jschreuder:
Are you sure the header isn't set? I know that Google Chrome for example will change the content type it uses to interpret if it detects another mime type. Seeing text isn't the same as the wrong content type header.

huglester:
Sorry guys. Seems I had the 'exit()' command at the end of the controller which was probably stopping the response class to set the headers, the correct usage is:
`$this->response->set_header('Content-type', 'image/png');`

thanks for the help and patience, and sorry for a mess

PR #221, opened 2011-06-15
Issue #220 : Fixed send_header

helloer:
send_header did not work because $this->headers does not exist, need change to $this->response->headers


jschreuder:
You're pushing to master, but looks like you're doing that from a branch off develop. Could you redo the pull-request on develop so I can actually see what you did in the diff?

PR #225, opened 2011-06-17
Inflector::friendly_title() bug fix

billmn:
Fixed wrong URL-friendly titles generation ... 
For details : http://www.fuelphp.com/forums/topics/view/2423

### DIFF
--- a/classes/inflector.php
+++ b/classes/inflector.php
@@ -228,27 +228,19 @@ class Inflector {
 		// Allow underscore, otherwise default to dash
 		$sep = $sep != '_' ? '-' : $sep;
 
+		// Remove tags
+		$str = \Security::strip_tags($str);
+
 		// Decode all entities to their simpler forms
 		$str = html_entity_decode($str, ENT_QUOTES, 'UTF-8');
-
-		$trans = array(
-			'\s+' => $sep,        // one or more spaces => seperator
-			$sep.'+' => $sep,   // multiple seperators => 1 seperator
-			$sep.'$' => '',	        // ending seperator => (nothing)
-			'^'.$sep => '',         // starting seperator => (nothing)
-			'\.+$' => '',            // ending dot => (nothing)
-			'\?' => ''                // question mark
-		);
-
-		foreach ($trans as $key => $val)
-		{
-			$str = preg_replace("#".$key."#i", $val, $str);
-		}
 
 		// Only allow 7bit characters
 		$str = static::ascii($str);
 
-		$str = \Security::strip_tags($str);
+		// Strip unwanted characters
+		$str = preg_replace("#[^a-z0-9]#i", $sep, $str);
+		$str = preg_replace("#[/_|+ -]+#", $sep, $str);
+		$str = trim($str, $sep);
 
 		if ($lowercase === true)
 		{
--- a/classes/inflector.php
+++ b/classes/inflector.php
@@ -226,7 +226,7 @@ class Inflector {
 	public static function friendly_title($str, $sep = '-', $lowercase = false)
 	{
 		// Allow underscore, otherwise default to dash
-		$sep = $sep != '_' ? '-' : $sep;
+		$sep = $sep === '_' ? '_' : '-';
 
 		// Remove tags
 		$str = \Security::strip_tags($str);


classes/inflector.php
```diff
@@ -231,24 +231,16 @@ class Inflector {
...
-			'^'.$sep => '',         // starting seperator => (nothing)
-			'\.+$' => '',            // ending dot => (nothing)
-			'\?' => ''                // question mark
-		);
-
-		foreach ($trans as $key => $val)
-		{
-			$str = preg_replace("#".$key."#i", $val, $str);
-		}
-
```
jschreuder, line 247: Any reason why this had to go? It's a lot better than your replacement `$str = preg_replace("/[\/_|+ -]+/", $sep, $str);`?

Other than this it's ready to merge, but I don't get this decision.

billmn, line 247: In my tests I obtain best results with my RegEx ... but I want your opinion and if we can optimize code.

So I would show you the 2 functions ( with $trans (current) and with my RegEx (new) ) and the tests results :
Functions : http://scrp.at/4m
Tests Results : http://scrp.at/4n

EDIT :
I have another fix to my code, I've moved strip_tags at the top ( below $sep ).
You if you get me your opinion on code I make changes, commit and re-pull request

thank you

ghost:
What you guys think about using:
`$sep = $sep === '_' ? '_' : '-';`
instead of: 
`$sep = $sep != '_' ? '-' : $sep;`

jschreuder:
@daniels0xff I don't think that's a good idea, any other separator would be bad practice. Also, next time explain why you want something, don't just paste some code. I don't want to have to deduce why you think a change might be good, we want both the reason and the code.

classes/inflector.php
jschreuder, line 247: The strip_tags change is a good idea because it looks more logical, shouldn't make any functional difference though as Inflector::ascii() will let all tag related characters through.

The comparisson between the 2 methods is invalid because the replacement is done at different places. I think the placement below the removal of unwanted characters is a good call, though both separators should be allowed (you might want to mix both, underscores for spaces and dashes for combining words).  
I don't think your regex matches all the cases the old one does, for example removing prefixed and suffixed separators.

One last nitpick: `preg_replace("#[/_|+ -]+#", $sep, $str);` is better than `preg_replace("/[\/_|+ -]+/", $sep, $str);`, easier to read.

billmn, line 247: @jschreuder

"... (you might want to mix both, underscores for spaces and dashes for combining words)."
I don't think is a good method to make a SEO friendly URL, personally I think is better use dashes OR underscores.

"I don't think your regex matches all the cases the old one does, for example removing prefixed and suffixed separators."
Why not?
Using <pre>"-<b>Fuel</b> is a community driven PHP 5 web framework __"</pre>
Result <pre>fuel-is-a-community-driven-php-5-web-framework</pre>

For last nitpick I've changed Regex, I can commit changes shortly...

ghost:
My bad, next time i will explain myself. This time I thought it was obvious.
Ok, so here it goes:

Currently i see that we are using `$sep = $sep != '_' ? '-' : $sep;` to check if $sep is either '-' or '_'
The comparison made there is not ok, it will work for most of the cases but it's not perfect.
Let's take this example:

```
$sep = true;
$sep = $sep != '_' ? '-' : $sep;
var_dump($sep);
```

We will get `bool(true)` which is not ok as we want only - or _ right?
I recommended to use `$sep = $sep === '_' ? '_' : '-';` instead as this does a identical comparison and will truly allow only '_' or '-' and noting else.

```
$sep = true;
$sep = $sep === '_' ? '_' : '-';
var_dump($sep);
```

This time we get the correct value `string(1) "-"` and also the code is a little bit more readable.

classes/inflector.php
jschreuder, line 247: "I don't think is a good method to make a SEO friendly URL, personally I think is better use dashes OR underscores."
I'm not saying one should do it, I'm just saying there's no reason to disallow it. You won't do this by accident, the underscore/dash must already be there. With that it would be best to have trim() work on both separators, thus `trim($str, '-_');`

And you're completely right about the prefixed/postfixed separators, I forgot about the trim().

jschreuder:
@daniels0xff
The thing about this is that we mostly check this stuff at the end of a long work-day, and we haven't been looking at the code as you may just have been. While it might be obvious if one were to take the time, it's little trouble for anyone posting an issue/question to explain it so it takes us less time to respond.

And I misread your previous post, yeah I think you're right. Only allowing a dash or underscore would be better.

ghost:
@jschreuder: Don't get me wrong, i totally agree with what you said and it's common sense for someone to also explain the changes he submits. And also i am aware of what you guys do, and that you take from your free time to develop this. I appreciate that, we all do. I think due to the fact that English is my second language sometimes i might say something that has a different meaning form what i actually wanted to say. So that part with "I thought it was obvious." was not said with a bad intention. I said that as in "it didn't cross my mind to also explain"

billmn:
So what we can do to close this request?

Can I add @daniels0xff's separator fix ... after this ... there are other fixes to apply?

jschreuder:
If these 2 are fixed I'll hit merge:
- Allow only dash and underscore as separator (@daniels0xff's fix)
- Don't disallow either dash or underscore if the other is the separator (and make sure your regex can deal with that)

billmn:
Ok, I've changed my code with @daniels0xff's fix

For the second request :
If I test this string in the function :
$test = "Peux-tu m'aide_r s'il te plaît? __ - ";

with :
\Inflector::friendly_title($test, $sep = '-', $lowercase = true)

now I obtain :
peux-tu-m-aide-r-s-il-te-plait

if I permit underscore ( in this case ) I can obtain that :
peux-tu-m-aide_r-s-il-te-plait-__

Is this what do you mean?
If separator is a dash I've to permit underscore and viceversa?

billmn:
I've made some other tests with your suggestion, but the result isn't exciting to me :

$test = "Mess'd up --text-- just (to) stress /test/ ?our! `little` \clean\ url fun.ction!?-->";
Inflector::friendly_title($test, $sep = '_', $lowercase = true)

result : 
mess_d_up_--text--_just_to_stress_test_our_little_clean_url_fun_ction_--

Personally I think that "_--text--_" should not be in a URL, this could confuse website's users.

PR #227, opened 2011-06-18
Fixed tests

tarnfeld:
There were changes in the way the URI stuff works with base_url and also the Arr class so some of the tests were failing. These changes make those failing tests pass.


tarnfeld:
Wait, damn i've done it again. Stupid stupid stupid pulling on the wrong branches!

PR #235, opened 2011-06-23
Add ability to "theme" pagination with HTML markup.

ghost:
I've added a $template variable and used it to "theme" the pagination widget. You can access it through the "set_config" method (so now it can be auto done in config file or by passing $config as attribute).

### DIFF
--- a/classes/pagination.php
+++ b/classes/pagination.php
@@ -34,6 +34,24 @@ class Pagination {
 	 * @var	integer	The number of total pages
 	 */
 	public static $total_pages = 0;
+	
+	/**
+	 * @var array The HTML for the display
+	 */
+	public static $template = array(
+		'wrapper_start'  => '<div class="pagination"> ',
+		'wrapper_end'    => ' </div>',
+		'page_start'     => '<span class="page-links"> ',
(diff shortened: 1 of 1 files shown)


WanWizard:
I'm not too sure this is the way it should be configured. 

The current set_config() now only accepts a complete template, not a single value. So you can't have a default in the config file, and then override a single value at runtime.

nerdsrescueme:
I can change it up easily enough. Better to separate each value into class properties? Or perhaps a different approach all together.

WanWizard:
I haven't made up my mind yet.

From a config perspective, it's logical to store it into an array. But then you can't use set_config() to override a single value.

Perhaps use array_merge() in set_config() on $template, so I can pass an array with a subset of parameters to set_config() and leave the others at their default value?

nerdsrescueme:
That's not a bad idea. Simplest to be sure. I'll try and think up another approach, but that seems best to me. Once we decide I'll drop it in and update the pull request.

nerdsrescueme:
Done, changed the string concat to match code standards and added examples to docs.

PR #239, opened 2011-06-23
Numeric helper class

ghost:
Hopefully I'm creating this pull request correctly, for once ;)

Added a numeric helper class, utilizing various methods from all over (sources are noted), and modified the methods to work within Fuel, and to the coding standards provided.

Let me know what you think, how I can improve it, or what I did wrong with my pull request this time.

The following tests were run to validate the functions are behaving as expected:
`echo 'zeroise: ' . Num::zeroise(123, 20) . '<br />';
echo 'human_time_diff: ' . Num::human_time_diff(time(), strtotime('-1 week')) . '<br />';
echo 'bytes: ' . Num::bytes('5MiB') . '<br />';
echo 'format_bytes: ' . Num::format_bytes(18939511627776) . '<br />';
echo 'format: ' . Num::format('1234567890', '000.000-0000') . '<br />';
echo 'mask_string: ' . Num::mask_string('1234567812345678', '**** **** **** 0000') . '<br />';
echo 'format_phone: ' . Num::format_phone('8005550123') . '<br />';
echo 'smart_format_phone: ' . Num::smart_format_phone('8001234567') . '<br />';
echo 'format_ssn: ' . Num::format_ssn('123456789') . '<br />';
echo 'format_exp: ' . Num::format_exp('1111') . '<br />';
echo 'mask_credit_card: ' . Num::mask_credit_card('12345678912345678') . '<br />';
echo 'format_usd: ' . Num::format_usd('6588.5878') . '<br />';`

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -135,8 +135,8 @@ Autoloader::add_classes(array(
 	'Fuel\\Core\\Migrate'    => COREPATH.'classes/migrate.php',
 	'Fuel\\Core\\Model'      => COREPATH.'classes/model.php',
 
-	'Fuel\\Core\\MongoDb'           => COREPATH.'classes/mongodb.php',
-	'Fuel\\Core\\MongoDbException'  => COREPATH.'classes/mongodb.php',
+	'Fuel\\Core\\Mongo_Db'           => COREPATH.'classes/mongo/db.php',
+	'Fuel\\Core\\Mongo_DbException'  => COREPATH.'classes/mongo/db.php',
 
 	'Fuel\\Core\\Output'      => COREPATH.'classes/output.php',
 	'Fuel\\Core\\Pagination'  => COREPATH.'classes/pagination.php',
--- a/classes/database/connection.php
+++ b/classes/database/connection.php
@@ -517,66 +517,7 @@ abstract class Database_Connection {
 	 * @return  string
 	 * @uses    static::table_prefix
 	 */
-	public function quote_identifier($value)
-	{
-		if ($value === '*')
-		{
-			return $value;
-		}
-		elseif (is_object($value))
-		{
-			if ($value instanceof Database_Query)
-			{
-				// Create a sub-query
-				return '('.$value->compile($this).')';
-			}
-			elseif ($value instanceof Database_Expression)
-			{
-				// Use a raw expression
-				return $value->value();
-			}
-			else
-			{
-				// Convert the object to a string
-				return $this->quote_identifier((string) $value);
-			}
-		}
-		elseif (is_array($value))
-		{
-			// Separate the column and alias
-			list ($value, $alias) = $value;
-
-			return $this->quote_identifier($value).' AS '.$this->quote_identifier($alias);
-		}
-
-		if (strpos($value, '"') !== FALSE)
-		{
-			// Quote the column in FUNC("ident") identifiers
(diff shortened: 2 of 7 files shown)


ghost:
Additional methods have been added since the initial request, and their details are described within their docblocks. Each function has been tested in the same manner as above, although I don't feel it necessary to list the manner in which they were tested, as it should be fairly obvious.

kbanman:
You should add the tests as a proper test suite. See the examples in the "tests" directory

ghost:
I'll definitely have to see if I can throw that in there, but would not having it deter from it being potentially added to the core?

I'm working on a package that will also utilize this class, and although the package can provide it, I'd prefer to utilize it from the core, and extending it within the package.

jschreuder:
For new stuff both tests & docs are required. Otherwise if you bail on us, we'd have to deduce how it should work when bugs are found.

jschreuder:
Some other points:
- We can't use code from Wordpress, they have a GPL license which is incompatible with our MIT license (theirs is more restrictive). If it's just inspired by, find your inspiration somewhere else or don't credit as it'll lead to a lack of clarity about licensing.
- You went a bit overboard with credit, some thing like a single line with 2 very basic PHP function don't really require credit - that's just the normal way to do things. Same goes for some of the other methods, some things can be done just one way and aren't complex - those don't require credit. Credit is given for algortims which are complex or unique.
- The human time diff method is already in the Date class, copy it from there to prevent licensing issues. We'll decide where it belongs most.
- Any country (US) specific stuff should go into a driver. Only stuff that's international should go into the core, county/language/location specific stuff should go into drivers and be easily extendeable for other countries.

On the whole I'm still thinking about whether this really belongs in the core, I'm leaning to yes as a very usefull utility. Though I'd like to have @dhorrigan's opinion as well.

ghost:
Remove the country specific methods
Removed the human_time_diff, as although it's not my call, having it in the date object seems more reasonable. I'll let @dhorrigan decide however, and he can add it back if he likes.
Also removed some of the unnecessary copyrights.

I'll continue working on it though, just need to find some extra time ;)

Not sure how to update which commits I'm pull-requesting from though, it's currently within my fork. I'll try to figure it out :P

ghost:
Any chance I can get thoughts from @dhorrigan regarding the potential for this? I don't want to invest too much into this if it's not going anywhere.

tomschlick:
With the changes made I think this would be a great addition to the core

tomschlick:
Am I missing something here? It looks like it was merged but I cant find reference to the class in either fuel:develop or syntaqx:develop...

did @syntaqx revert his repo or something?

ghost:
My repo bugged out, and I had to remove it from Github. But I don't know anything about why the merger wouldn't work.

tomschlick:
ah. well until its merged its still a reference to your repo not a copy of it. so by redoing your repo you removed the reference which is why nothing about the Num.php class is showing up. 

what i would suggest is to resubmit it but do it from a separate branch so that if you need to do anything to the develop branches it doesn't screw anything up (as long as your new branch is off of develop).

ghost:
Alright. I'm just trying to figure out how to branch develop into a separate branch (like develop-1), so I can work on more than one thing at a time. Once I figure that out, I'll resubmit. You wouldn't happen to know anything about it would you?

tomschlick:
checkout develop and then do "git branch branch-name"

ghost:
The new pull request can be found at: https://github.com/fuel/core/pull/276

PR #242, opened 2011-06-24
Updated View to allow files with non .php extensions to be loaded.

grekker:
Added an additional parameter to View->set_filename() that allows the filename extension to be specified. This allows you to use non .php files as views, such as .html, .js, etc.

Signed-off-by: Chuck Liddell liddellc@gmail.com

### DIFF
--- a/classes/view.php
+++ b/classes/view.php
@@ -300,15 +300,17 @@ class View {
 	/**
 	 * Sets the view filename.
 	 *
-	 *     $view->set_filename($file);
+	 *     $view->set_filename($file, $extension);
 	 *
 	 * @param   string  view filename
+	 * @param   string  view filename extension
 	 * @return  View
 	 * @throws  Fuel_Exception
 	 */
-	public function set_filename($file)
+	public function set_filename($file, $extension = null)
 	{
-		if (($path = \Fuel::find_file('views', $file, '.'.$this->extension, false, false)) === false)
+		$extension = $extension ?: '.' . $this->extension;	
+		if (($path = \Fuel::find_file('views', $file, $extension, false, false)) === false)
 		{
 			throw new \Fuel_Exception('The requested view could not be found: '.\Fuel::clean_path($file));
 		}


jschreuder:
Not sure yet if this is the right way, a better way might be to auto-detect if an extension is given and only add the default extension when none was given. But all of this might just be bad practice, why would you want the same view types with different extensions?

grekker:
I use it as a way to deliver mostly static content (like javascript) with the option of including some dynamic content embedded in that javascript.

Of course you could serve javascript that was in a my_javascript.php file and just pass the proper header, but there is a convenience factor to leaving it in the original file extension (syntax highlighting in an IDE, for example).

Thoughts?

grekker:
Didn't mean to close.

jschreuder:
As this might cause ambiguity with the Parser package and not something I would consider good practice I'm closing this. If you need this you can add the ability to do it to a controller to parse and output (which is what I'm doing in my own code btw).

grekker:
Would you mind posting a code sample of how  you might do that?

Forgive my ignorance of the codebase but how can you have an arbitrary file parsed when the View class assumes a .php extension?

Here is my controller code for loading a javascript file, for example:

``` php
$view = View::factory();
$view->set_filename( $view_file, '.js' );
$this->response->body = $view;
$this->response->set_header('content-type', 'text/javascript');
```

jschreuder:
I just use `$this->response->body = File::read_file(APPPATH.'views'.DS.$file, true)` instead of parsing it as a view.

Though I do some checking to see if it's a file-extension I allow (js, css, txt, png, jpg, gif, swf) and have headers set for, all others I disallow as well as double-dots (to prevent going to the upper dir).

grekker:
Ah, I see now. We're doing different things.

My objective is to have the .js files and others actually parsed as a view. As in, I can embed php variables in my javascript so that I can dynamically create javascript.

I think using Fuel to dynamically generate javascript is pretty damn powerful. I'm building an entire web application around that concept.

Here is a simple example:

Controller

```
$data = array( "a variable" => "Dynamic content, perhaps pulled in part from a DB." );
$view = View::factory( null, $data );
$view->set_filename( $view_file, '.js' );
$this->response->body = $view;
$this->response->set_header('content-type', 'text/javascript');
```

Javascript file

```
alert( "<?php echo $a_variable;?>" );
```

That's why I need View to be able to parse non-php filenames. I think it's a worthwhile ability for Fuel to have.

grekker:
Thoughts on my most recent comment? I still believe my code is a viable and useful change.

PR #244, opened 2011-06-25
Stops dates > 10 years from hanging; no rounding up dates by default

simonberube:
Previous to this fix: dates beyond exactly one decade (to the day) caused server to hang, and time difference was rounded. First, while I could see instances where you may want to refer to dates in past as having been decades in the past or rounding upwards, that is very much an exception. E.g., On 9/11/2011, 9/11 happened <?=Date::time_ago(mktime(0,0,0,9,11,2001),mktime(0,0,0,9,11,2011));?>... Should that say 1 decade ago? What about when it is 14 years ago? Without this commit, at 14 years, it will say 1 decade. Then, when it is 15 years ago, it will tell you it happened 2 decades ago, but that would be false. The third parameter introduces the option to round instead of floor the value.

### DIFF
--- a/classes/date.php
+++ b/classes/date.php
@@ -203,10 +203,12 @@ class Date {
 	/**
 	 * Returns the time ago
 	 *
-	 * @param	int		UNIX timestamp from current server
+	 * @param	int		Beginning UNIX timestamp
+	 * @param	int		End UNIX timestamp, null means current time
+	 * @param	bool	Set to true in order to round last value
 	 * @return	string	Time ago
 	 */
-	public static function time_ago($timestamp, $from_timestamp = null)
+	public static function time_ago($timestamp, $from_timestamp = null, $round = false)
 	{
 		if ($timestamp === null)
 		{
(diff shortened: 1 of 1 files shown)


jschreuder:
I agree with the removal of decade, that's just too big a timespan - century/milenium would be good but decade only makes it less usefull.

But I don't agree with adding a param for floor/round, adding params should be logical and in this case I don't think this is. What if someone wants ceil instead? Add a fourth param? Or suddenly change this to string values instead of boolean? What if we decide that we might want to allow the programmer to choose which time-measure is used? Should that go after $round? Makes no sense, but putting it before it would break the existing interface.  
Bottom line: it's too trivial. I think round is fine in this case as a standard. If someone really dislikes it there's always the ability to overwrite core methods.

simonberube:
I can't see there ever being need to round time differences. It goes against logic really. Time ago always returns "ago"; so I think you'd have a hard time making an example use where rounding would be appropriate. I could see some rare cases where you would be OK with this funtion returning something like "about 1 year ago" but if you're at 6 months and it tells you 1 year ago, isn't that broken? There should he higher regard to precision.

PR #245, opened 2011-06-25
Added support in DBUtil class for creating tables with indexes and foreign keys.

calvinfroedge:
Tried to stay as true to Dan's form as possible.  Added some comments before functions where they were missing.  

I could not get return \implode to work for me as in other code.  Had to do return implode.

Tested and all works.  Was successful with my own migration using this code.

### DIFF
--- a/classes/dbutil.php
+++ b/classes/dbutil.php
@@ -72,19 +72,42 @@ class DBUtil {
 		return \DB::query('RENAME TABLE '.DB::quote_identifier(DB::table_prefix($table)).' TO '.DB::quote_identifier(DB::table_prefix($new_table_name)),DB::UPDATE)->execute();
 	}
 
-	public static function create_table($table, $fields, $primary_keys = array(), $if_not_exists = true, $engine = false, $charset = null)
+	/**
+	 * Creates a table
+	 *
+	 * @param	string	$table			the table you wish to create
+	 * @param	array	$fields			array of fields to be created in table
+	 * @param	array	$primary_keys	primary keys for the table
+	 * @param	bool	$if_not_exists	create table if not exists?
+	 * @param	string	$engine			Db engine to use
+	 * @param	string	$charset		A character set for the table
+	 * @param	array	$indexes		Array of indexes
+	 * @param	array	$foreign_keys	Array of foreign keys
+	 * @return  int    Number of rows affected
+	 */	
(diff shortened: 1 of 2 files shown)


classes/dbutil.php
```diff
@@ -86,6 +99,16 @@ class DBUtil {
 			$primary_keys = \DB::quote_identifier($primary_keys);
 			$sql .= ",\n\tPRIMARY KEY ".$key_name." (" . implode(', ', $primary_keys) . ")";
 		}
+		if( ! empty($indexes))
+		{
+			$sql .= ",";
+			$sql .= static::process_indexes($indexes);
+		}
+		if( ! empty($foreign_keys))
```
philsturgeon, line 107: This is close but a little off. It should be `if ( ! empty` to keep in line with the style guide.

philsturgeon:
Could you update your branch as this cannot be merged right now. Also, could you make sure it does not cross over with https://github.com/fuel/core/pull/557 as both these changes look good.

classes/dbutil.php
```diff
@@ -186,6 +219,51 @@ class DBUtil {
...
+	 * @return   string    the formated indexes string
+	 */	
+	public static function process_indexes($indexes)
+	{
+		$sql = "";
+		$index_list = array();
+		foreach($indexes as $index)
+		{
+			$index = array_change_key_case($index, CASE_UPPER);
+			($index['UNIQUE'] === true) ? $unique = 'UNIQUE ' : $unique = '';
```
ushis, line 234: Isn't it better to write:

``` php
<?php
$unique = ($index['UNIQUE'] === true) ? 'UNIQUE ' : '';
```

classes/dbutil.php
```diff
@@ -186,6 +219,51 @@ class DBUtil {
...
+			$sql = "\n\t".$unique."INDEX ".\DB::quote_identifier($index['NAME'])." (" . \DB::quote_identifier($index['FIELD']) . " " . $index['ORDER']. ")";
+			$index_list[] .= $sql;
```
ushis, line 236: I think it should be

``` php
<?php
$index_list[] = $sql;
```

without the dot.

ushis:
Damn this pull request is super old. It would be nice to have this in 1.1

classes/dbutil.php
philsturgeon, line 234: Agreed, but its pretty minor.

billmn:
agree

PR #252, opened 2011-07-01
Not using <> in MAIL FROM (smtp) caused an error if you use from("email", "name");

ipalaus:
Hello,

Using the current email.php class causes an error using SMTP if you use Email::factory()->from("example@domain.tld", "Name")...

If you use the from() with only the e-mail works fine.

Attached is how it can be solved, just adding <> in the "MAIL FROM:".

Thank you,
Isern


ipalaus:
hm... why it added commits from other people? I just forked and made the change in less than 10min... :-\

ipalaus:
hm... why it added commits from other people? I just forked and made the change in less than 10min... :-\

tomschlick:
When you send a pull request, be sure to use the "Change Commits" button on the top left of the form to select the commits you want to send. This shouldn't be a problem this time because all of those have already been merged in.

ipalaus:
I'll do it next time, thank you tomschlick!

tomschlick:
no problem. Ive done it a few times myself and I'm not aware of any way to edit a pull request files.

jschreuder:
As we'll remove the current email class before v1.0 I won't merge fixes to it anymore.

This is because of a license issue, we can't use most of the code in the current email class because it's been licensed more restrictive than ours. It should have been removed a while ago but we never got round to it.

PR #257, opened 2011-07-06
Query Builder Update - Added Join support

dlpetrie:
Added 'JOIN' support to update.php in database/query/builder/update.php.  Just copy/pasted code from the select.php file where needed.


WanWizard:
We don't accept pull requests on Master, only on the develop branch.

PR #258, opened 2011-07-07
Module Routes

andrewwheal:
Made the module name get automatically added to the front of routes loaded from a module and also added support for _root_ routes in modules

### DIFF
--- a/classes/request.php
+++ b/classes/request.php
@@ -295,7 +295,22 @@ class Request {
 			if (file_exists($modpath .= 'config/routes.php'))
 			{
 				// load and add the routes
-				\Config::load(\Fuel::load($modpath), 'routes');
+				$modname = substr($modpath, strrpos($modpath, '/', -2) + 1, -1);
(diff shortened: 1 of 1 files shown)


WanWizard:
I have mixed feelings about this. I can see this can be handy something, but I don't like magical modifications...

andrewwheal:
I wouldn't have said it was magical modifications, all it is doing is adding the module name in front of the routes which saves you having to do it yourself. Also, it allows use of _root_ in a module which IMO is a nice feature and allows a module routes file to function exactly the same as the routes file in the app/config folder.

phil-lavin:
This does seem to make sense and adds a level of consistency between the main routes config and the module specific routes config.

WanWizard:
I see where you're coming from.

But the code needs to be improved. If you prefix the route with the module name, you need to check if it isn't there already, This code breaks the current module routes file, which already has the module name prefixed.

PR #268, opened 2011-07-12
Added symbolic linking methods to \File.

ghost:
http://php.net/manual/en/function.symlink.php support added by default inside of 5.3.0. Only issue is that it doesn't work on Windows servers pre Vista. I believe we could add a hack to support legacy Windows systems, but that may not be in the spirit of progress.

I am using this as a way to solve problems with module asset serving... see http://fuelphp.com/forums/topics/view/3067

### DIFF
--- a/classes/file.php
+++ b/classes/file.php
@@ -503,6 +503,61 @@ class File {
 	}
 
 	/**
+	 * Create a new symlink
+	 *
+	 * @param	string				target of symlink
+	 * @param	string				destination of symlink
+	 * @param	bool				true for file, false for directory
+	 * @param	string|File_Area|null		file area name, object or null for base area
+	 * @return	bool
+	 */
+	public static function symlink($path, $link_path, $file = true, $area = null)
+	{
+		$path	   = rtrim(static::instance($area)->get_path($path), '\\/').DS;
+		$link_path = rtrim(static::instance($area)->get_path($link_path), '\\/').DS;
+
+		if ($file and ! is_file($path))
+		{
+			throw new \InvalidPathException('Cannot symlink: given file does not exist.');
+		}
+		if ( ! $file and ! is_dir($path))
+		{
+			throw new \InvalidPathException('Cannot symlink: given directory does not exist.');
+		}
+		elseif (file_exists($link_path))
+		{
+			throw new \FileAccessException('Cannot symlink: link path already exists.');
+		}
+
+		$return = symlink($path, $link_path);
+		return $return;
+	}
+
+	/**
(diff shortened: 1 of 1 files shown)


jschreuder:
Generally we don't have multiple names for the same method, unless there's a really clear and very good reason to. In this case I don't see any reason to allow both symlink & alias. I still have to check whether this all works enough like the other methods, but that needs to be changed at least before we consider this.

nerdsrescueme:
There you go, removed the alias methods and made the boolean file variable slightly more descriptive.

nerdsrescueme:
I had an idea for this to integrate it into Asset. perhaps a static method that creates a symbolic link in the assets folder pointing to a module's own asset folder...

I think it's an elegant way to handle assets inside of modules, and make them more compartmental.

Any thoughts?

jschreuder:
The Asset class shouldn't do file operations in such a way - if you want to use symlinks for this purpose you can do it, but that's not the right solution or even an option for everyone. Also the Asset class already supports adding multiple locations to search for the assets, which is a far more elegant solution (though symlinks might be a bit faster).

nerdsrescueme:
Wait... So you can use assets from outside the public folder with the current asset class? Please tell me I have been doing all this work for nothing lol. 

Frank Bardon Jr.

## Chief Nerd

Nerds, Rescue Me!
frank@nerdsrescue.me
http://nerdsrescue.me

jschreuder:
http://fuelphp.com/docs/classes/asset.html#method_add_path

nerdsrescueme:
Yeah it's relative to base_URL so the server couldn't access APPPATH/modules/... So models couldn't contain their own assets. I won't push the issue. I'm going to create a module extension package that way you can install these features I speak of at your discretion. Perhaps it has no place in core. 

What about symbolic linking in general though?

Frank Bardon Jr.

## Chief Nerd

Nerds, Rescue Me!
frank@nerdsrescue.me
http://nerdsrescue.me

jschreuder:
You can use modules like with any view/config/lang file, either it's just available because the current controller is in the module (in which case the module is the first one checked). Or you can prepend "Module_Namespace::" before the filename, ex. the module "test" and file "base.css" would be "test::base.css".

For that to work the module needs to be part of your `always_load` config in app/config/config.php or load it manually using `Fuel::add_module()`.

nerdsrescueme:
Thank you Jelmer, I just assumed that since it was being served relative to base_url that it would be inaccessible... Clearing that up helped a lot. Thanks again.

Issue #263, opened 2011-07-10
Notice with core/classes/security.php htmlentities()


Mitchell64:
I just downloaded the latest core code from the development branch, and code that I've been using for a while just generated the following Notice. I'm not sure yet what's triggering the notice, but I thought I'd post this in case the developer who is familiar with this code may know. I can do more investigating if necessary.

Notice!

ErrorException [ Notice ]: Object of class stdClass could not be converted to int

COREPATH/classes/security.php @ line 170:

169:    // Prevent looping & encoding twice
170:    if (in_array($value, $already_cleaned))
171:    {

jschreuder:
This is an incredibly weird error, there's no casting to int there... the first var $value is accepted as mixed and does no casting and the second var is always an array as expected.

WanWizard:
You'll get this notice if $already_cleaned is an object instead of an array. 
But I don't see how it could become an object though...

huglester:
Hello,
I must join this conversation, because I think I've met some problems with similar issue.
Until yesterday, before update (I was updating around week ago or so), so security.php also updated, and I started to get infinite loops, logs were showing this: Error - 2011-07-11 08:22:15 --> Error - Maximum function nesting level of '100' reached, aborting! in /var/www/vhosts/vibe/fuel/core/classes/database/connection.php on line 526

Later I have found, that infinite loop occurs here:

``` php
public static function dropdown()
    {
        $results = static::find('all');

        //Lets generate array now
        $return = array(0 => 'Select country');
        foreach ($results as $item)
        {
            $return[$item->id] = e($item->name);
        }

        return $return;
    }
```

As I've said, this method was working fine, until update, so later I've spotted, that doing:
$return[$item->id] = $item->name; // without escaping - solved my problem, webpage starts to load.

after going further I'm stuck now, because some insane things happen:

``` php
$return[$item->id] = e($item->name); // NOT working
$return[$item->id] = \Security::htmlentities($item->name, ENT_COMPAT, \Fuel::$encoding, false); // NOT working
$return[$item->id] = htmlentities($item->name, ENT_COMPAT, \Fuel::$encoding, false); // WORKS
$return[$item->id] = $item->name; // WORKS
```

for these tests, I'm sure I was passing the 'string' from the database

jschreuder:
@huglester I'm having a feeling there's something else going on here, the Security::htmlentities() method would never invoke the usage of the Database_Connection class. I just tested this and didn't have any problems with the e() function.

[...]

PR #273, opened 2011-07-14
Security::htmlentities() infinite loop

andrewwheal:
When you have a call to a function in a view and that function loads a child view (with data) you get an infinite loop in the Security::htmlentities() method.

Loading the parent view gets it passed through to the Security::htmlentities() method and because it is a whitelisted class it just gets added to the $already_cleaned array. When the child view with its data gets loaded by the function called in the parent view each item of data passed to the child view gets passed to the Security::htmlentities() method.

These strings are then compared to the parent view by in_array() which by its default behaviour tried to convert the view to a string so that the comparison can be made. In converting the parent view to a string it tries to render it and thus must call the function in it which contains the child view to get its value. However, to do this the child view must be loaded and its data passed through the Security::htmlentities() method. Herein lies the recursion.

Making the in_array() perform strict checking by passing it true as the 3rd parameter fixes this recursion as the parent view will not be converted to a string.

This may also have been an alternative fix to the previous issue with Security::htmlentities() - https://github.com/fuel/core/issues/263

### DIFF
--- a/classes/security.php
+++ b/classes/security.php
@@ -167,7 +167,7 @@ class Security {
 		static $already_cleaned = array();
 
 		// Nothing to escape for non-string scalars, or for already processed values
-		if (is_bool($value) or is_int($value) or is_float($value) or in_array($value, $already_cleaned))
+		if (is_bool($value) or is_int($value) or is_float($value) or in_array($value, $already_cleaned, true))
 		{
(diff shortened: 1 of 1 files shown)


jschreuder:
Thanks, good fix.

PR #276, opened 2011-07-14
Numeric helper class 

ghost:
Recreation of my original pull request due to repository issues.

The original thread can be found at: https://github.com/fuel/core/pull/239

Where did we leave off on the implementation of this?

### DIFF
--- a/classes/num.php
+++ b/classes/num.php
@@ -133,7 +133,7 @@ class Num {
 		{
 			if (doubleval($bytes) >= $mag)
 			{
-				return static::precision($bytes / $mag, $decimals).' '.$unit;
+				return sprintf('%01.'.$decimals.'f', ($bytes / $mag)).' '.$unit;
 			}
 		}
 
--- a/classes/num.php
+++ b/classes/num.php
@@ -212,7 +212,7 @@ class Num {
 		
 		while ((strlen($format) - 1) >= $fpos)
 		{
-			if (static::is_alphanumeric(substr($format, $fpos, 1)))
+			if (ctype_alnum(substr($format, $fpos, 1)))
 			{
 				$result .= substr($string, $spos, 1);
 				$spos++;
@@ -259,7 +259,7 @@ class Num {
 
 		while ((strlen($format) - 1) >= $fpos)
 		{
-			if (static::is_alphanumeric(substr($format, $fpos, 1)))
+			if (ctype_alnum(substr($format, $fpos, 1)))
 			{
 				$result .= substr($string, $spos, 1);
 				$spos++;
@@ -345,18 +345,6 @@ class Num {
 	{
 		is_null($format) and $format = static::$config['formatting']['credit_card'];
 		return static::mask_string($string, $format);
-	}
-
-	/**
-	 * Determines if a string has only alpha/numeric characters.
-	 *
-	 * @param   string     the string to check as alpha/numeric
-	 * @return  boolean
-	 * @see     ctype_alnum
-	 */
-	public static function is_alphanumeric($string)
-	{
-		return ctype_alnum($string);
 	}
 }
 
--- a/classes/num.php
+++ b/classes/num.php
@@ -122,22 +122,22 @@ class Num {
 	public static function format_bytes($bytes = 0, $decimals = 0)
 	{
 		$quant = array(
-            'TB' => 1099511627776,  // pow( 1024, 4)
-            'GB' => 1073741824,     // pow( 1024, 3)
-            'MB' => 1048576,        // pow( 1024, 2)
-            'kB' => 1024,           // pow( 1024, 1)
-            'B ' => 1,              // pow( 1024, 0)
-        );
+			'TB' => 1099511627776,  // pow( 1024, 4)
+			'GB' => 1073741824,     // pow( 1024, 3)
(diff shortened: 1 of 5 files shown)


tomschlick:
One thing this still needs is unit tests. Take a look here https://github.com/fuel/core/blob/develop/tests/str.php for an example of how to write some simple tests.

ghost:
Added the unit tests. I currently have no way to test if the file works though, as I can't install phpunit on this machine, sorry.  Let me know if there's anything else, or if the tests need fixing. :)

jschreuder:
I think there's a place for this, but as I mentioned in an earlier I'd like @dhorrigan to weigh in as well.

And I also think I already mentioned internationalization earlier: no strings should be hardcoded English (but lang files, even 'K', 'M', 'B') and patterns (like phone nr) should be configurable or in a driver. Until that's done it won't be merged.

ghost:
What do you mean by phone nr? The only thing that has presets would be the smart_phone_format (phone_format only uses a default value that is easily changed on usage), but I could make smart_format_phone configurable easily enough. Would you prefer the 'K','M','B' to be in a language file or a conf file? I haven't used much i18n with Fuel yet, so I'm not sure the best way to proceed with it.

jschreuder:
I'm sorry, you're right about the phone number - I read through it a bit too quickly. Though in that case it still would need to be configureable, something like array(strlen => pattern, strlen =>pattern). Also I'd prefer if the defaults would be null and then taken from config, and only after that fall back to American. Here's an example for this:

```
public static function format_phone($string = '', $format = null)
{
    is_null($format) and $format = \Config::get('numeric.default_phone', '(000) 000-0000');
    return static::format($string, $format);
}
```

About when to use config/lang: phone numbers aren't related to lang but to country and should go in config. Lang files are for everything that might be considered language (usage of "k" to represent thousands is among those).

classes/num.php
```diff
@@ -0,0 +1,421 @@
...
+	);
+
+	/**
+	 * Add leading zeros to a number
+	 *
+	 * @param   integer
+	 * @param   integer
+	 * @return  string
+	 */
+	public function zeroise($number = 0, $threshold = 1)
```
nerdsrescueme: Should this be static?

ghost: That it should. I'll fix that with my next set of commits around the lang files. Good catch!

ghost:
Fixed the bug that @nerdsrescueme found, and added references to config and lang files instead of default values.

Did I miss anything?

ghost:
Sorry to double post, has @dhorrigan had a chance to look at this yet? I'd love to hear your thoughts :)

ghost:
Woops, didn't mean to close.

dhrrgn:
I like most of this.  However, get rid of zeroise(), precision() and percentage().  They are severely redundant to core PHP functionality.  I know we have some redundancy in the core already, but these 3 functions are just fluff.

The reason I don't like these particular 3 methods is because the class is pretty heavy to load just for a function a developer can call themselves.  People will call Num::precision() once or twice and not use any more of the class...so they loaded 400 lines of code into memory just to call a sprintf().  Kind of pointless no?

ghost:
Yeah, I  agree with you. I just saw it in other functions in other frameworks, so I saw it somewhat ideal to include them to help those coming from other frameworks have similar functionality. Either way, there's no harm in removing them.

Anything else for you sir?

ghost:
As a note, as of the latest commits, I removed is_alphanumeric() as well, using the same logic for the removal of zeroise(), precision() and percentage(), rather than wrapping ctype_alnum with is_alphanumeric, they should just be using that function. I can put it back if you prefer, but it makes more sense to follow the same logic.

dhrrgn:
Ok, I like this, but it won't make the 1.0 cut.  Just to late in the game unfortunately.  Will definitely make it for 1.1 tho.

Thanks.

ghost:
Wish we could have gotten it in sooner... The original thread was from like a month ago.. hah :)

Oh well, just hope it can get merged in the next batch around.

jschreuder:
> The original thread was from like a month ago.. hah

Let me keep civil and just say this is not the way to get anything approved.

ghost:
Wasn't trying to offend anyone, at all. Was simply making light of not getting it into 1.0, not anything more than a joke.

jschreuder:
> Was simply making light of not getting it into 1.0, not anything more than a joke.

In that case I'll be even more specific: with "_this_" in "_this is not the way to get anything approved_" I meant: "_making sour jokes is not the way to get anything approved._"

dhrrgn:
No use in complaining that it didn't make it in time.  The review process sometimes takes longer than others, especially when changes are requested.  You honestly cannot expect me to merge in a major patch literally the day before the release.

I almost didn't merge this at all because i was angered by the comments.  However, the framework could use it.

tomschlick:
This still needs docs added. Idk if @syntaqx has them already coded up or not, if so just submit a pull request on fuel/docs.

ghost:
I apologize to anyone I may have offended through my comment, it wasn't meant to be sour, it wasn't meant to be rude.

As I said, I was just attempting to make light of something, but I will use more caution with my words in the future to make sure no misinterpretations can be made by them.

Sorry again.

PR #288, opened 2011-07-18
Format _from_csv patch

ghost:
While working with the format class, and a csv type data set from an external company, the data wasn't working correctly within the current format implementation. After playing around with it a bit, it also seems that Fuel isn't using str_getcsv which is a PHP 5.3 function to make our csv lives easier.

This patch both fixed my issue with the csv, as well as utilizes the functionality for str_getcsv. Although the delimiter, enclosure, escape and eol parameters aren't something the user can provide, I left them open so we could potentially add functionality to provide those values later.

Thoughts?

### DIFF
--- a/classes/format.php
+++ b/classes/format.php
@@ -310,63 +310,13 @@ class Format {
 	 * @param   string  $string
 	 * @return  array
 	 */
-	protected function _from_csv($string)
+	protected function _from_csv($string, $delimiter = ',', $enclosure = '"', $escape = '\\', $eol = "\n")
 	{
-		$data = array();
+		$data = str_getcsv($string, $eol, $enclosure, $escape);
 
-		// Splits
-		$rows = explode("\n", trim($string));
-
-		// TODO: This means any headers with , will be split, but this is less likley thay a value containing it
-		$headings = array_map(function($value) {
-				return trim($value, '"');
-			}, explode(',', array_shift($rows)));
-
-		$join_row = null;
-
-		foreach ($rows as $row)
+		foreach($data as &$row)
 		{
-			// Check for odd numer of double quotes
-			while (substr_count($row, '"') % 2)
-			{
-				// They have a line start to join onto
-				if ($join_row !== null)
-				{
-					// Lets stick this row onto a new line after the existing row, and see what happens
-					$row = $join_row."\n".$row;
-
-					// Did that fix it?
-					if (substr_count($row, '"') % 2)
-					{
-						// Nope, lets try adding the next line
-						continue 2;
-					}
-
-					else
-					{
-						// Yep, lets kill the join row.
-						$join_row = null;
-					}
-				}
-
-				// Lets start a new "join line"
-				else
-				{
-					$join_row = $row;
-
-					// Lets bust outta this join, and go to the next row (foreach)
-					continue 2;
-				}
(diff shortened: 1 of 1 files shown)


ghost:
http://pastecode.com/3330

Above is the example. The entire file is about 64mb's though, and contains 1 additional blank line at the end of the file. But overall, that's what you're looking at.

I realize that there was a lot of functionality inside of yours that I skimmed over, but is there any reason why we wouldn't want to implement multi-line support for str_getcsv? Or is it easier to just not?  I think one of the issues I'm seeing with the current _from_csv, is the lack of delimiter support. But I could be wrong.

Let me know what you find mate :)

Also, the additional spacing the customer provides to align the values isn't the issue here, but rather I was receiving an empty array back from the function.  I'm probably just going to trim() the space out, so I'm not expecting to have that removed.

WanWizard:
Your array is empty because:
- _from_csv() assumes the first row of your file contains csv headers
- _from_csv() assumes all fields are double-quote delimited
- only stores the parsed line if it has the same number or columns as the header row

Which means the logic is incomplete, for example in a lot of csv exports, only strings are enclosed in quotes, numbers aren't. Most csv import filter have a large array of config items, and that isn't for nothing...

ghost:
If the last line on the pastecode isn't the same count as the other two, that was a bad copypasta job on my part.

As far as the double-quote delimited, that's fine when dealing with my own csv's, but what about when dealing with customers? I don't have control over how they delim their files, so how would I account for this?

Headers are also an issue as stated above, and not having control. What do I do to tell it I don't have headers? Or to use indexes?

Basically, what would you recommend for this csv set?

WanWizard:
Well, that's exactly my point. Take for example any spreadsheet, and see what options you can set/select when you wan to import a csv file. You've got comma and tab seperated files, fxed and variable width, with and without headers, single, double or no quotes, all fields quoted, only strings, or only strings with spaces, the list goes on...

Hence my statement that the logic (in the _from_csv method) is incomplete, I think the call needs an options array so you can configure it.

nerdsrescueme:
Is there a possibility not yet thought of? Auto-detect of some sort? I guess the question becomes do you want to require strict rules for working with csv files? Or are there standards that should be followed, a spec of sorts?

ghost:
I believe there is a non-strict standard that is generally preferred, but the problem is working with outside entities (in the case of APIs, etc), where you don't have control over their formatting. We have that issue right now at my work, where we get improperly formatted csv's that don't import into excel. It's a nightmare.  That being said, we can't account for every possibility under the sun, but we should be fairly lenient.

dhrrgn:
This is one thing that will most likely never work for all situations.  We MIGHT be able to if we get some super-regex-ninja action on it.  However, I have tried before and got lost in my own regex and just made it worse.

ghost:
It might be worth examining some of the code they use in software like open office or similar software and seeing what we can emulate as far as that goes?

nerdsrescueme:
That doesn't seem like a bad idea. If you can read it :)

jschreuder:
Even if this is going to be changed, it won't be for 1.0 - changing the milestone to 1.1

jschreuder:
I'll close this as this isn't active for a while and has no current solution.

PR #295, opened 2011-07-21
Allow override of method_params from Request execute()

phil-lavin:
This facilitates the passage of objects, arrays, etc. when using Request in a HMVC context. My understanding is that currently you can only pass string data into a controller/method in a HMVC request.

### DIFF
--- a/classes/request.php
+++ b/classes/request.php
@@ -354,9 +354,10 @@ class Request {
 	 *
 	 *     $request = Request::factory('hello/world')->execute();
 	 *
+	 * @param  array|null  $method_params  An array of parameters to pass to the method being executed
 	 * @return  Request  This request object
 	 */
-	public function execute()
+	public function execute($method_params = null)
 	{
 		logger(Fuel::L_INFO, 'Called', __METHOD__);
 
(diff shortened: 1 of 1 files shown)


nerdsrescueme:
Would an array_merge be more beneficial here so you could add to, not replace the method_params property?

dhrrgn:
I like this.  Not sure about @nerdsrescueme's suggestion tho.  I think it is a "one or the other" type of situation.

jschreuder:
My only gripe with all of this is that it might get a little unclear when the URI contains params too. When the URI contains segments and there are segments passed as array it's kind of weird to just ignore the URI. After a bit of thinking I think I agree with @nerdsrescueme's suggestion, by merging them you put the URI's segments first and make the passed args the "additional arguements" instead of a replacement. You can still use only passed arguements by not passing any through the URI, but you're not overwriting anything.

Another possible pitfall is if we might need to add another first param to Request::execute(), but passing additional args seems a logical param so I don't think that's a problem.

Savageman:
Very useful in HMVC context! I like this feature. :-)

PR #297, opened 2011-07-21
Make namespace search case-insensitive as php's namespaces are case-insensitive

phil-lavin:


dhrrgn:
This breaks forward compatibility as we will be moving to using a strict PSR-0 compliant autoloader.  PSR-0 specifies that the file path and namespace match, including case.

This will not be included.  At least it didn't take you long!  Thanks.

PR #305, opened 2011-07-25
Sprintf and Lang in Exceptions

ghost:
The added changes allow for using sprintf-alike statements or Lang::line calls using Fuel_Exception. It also works for pure-strings so nothing currently in use will break.

This will make peoples lives a whole lot easier when working with i18n and having to throw errors. Currently I'm having to do the following to execute a language error:

```
throw new \Fuel_Exception(\Lang::line('payments.required', array('fields' => implode(', ', $fields))));
// or
throw new \Fuel_Exception(__('payments.required', array('fields' => implode(', ', $fields))));
```

The code I added reduces this to

```
throw new \Fuel_Exception('payments.required', array('fields' => implode(', ', $fields));
```

Additionally, the change allows you to use sprintf-alike statements

```
throw new \Fuel_Exception('This is my error %s', 'message');
```

### DIFF
--- a/classes/fuel/exception.php
+++ b/classes/fuel/exception.php
@@ -36,7 +36,7 @@ class Fuel_Exception extends \Exception {
 			{
 				$line = \Lang::line($args[0], $args[1]);
 
-				if($line != $args[1])
+				if($line !== false and $line != $args[1])
 				{
 					$args = array($line);
 				}
--- a/classes/fuel/exception.php
+++ b/classes/fuel/exception.php
@@ -12,4 +12,39 @@
 
 namespace Fuel\Core;
 
-class Fuel_Exception extends \Exception {}
+/**
+ * Default exception, mainly used for general errors. All Fuel specific
+ * exceptions extend this exception
+ *
+ * @author     Chase "Syntaqx" Hutchins <syntaqx@gmail.com>
+ * @version    1.0
+ * @package    Fuel
+ */
+class Fuel_Exception extends \Exception {
+
+	public function __construct()
+	{
+		if(($count = \func_num_args()) === 0)
+		{
+			parent::__construct('Unknown error');
+		}
+		else
+		{
+			$args = \func_get_args();
+
+			if($count == 2 and is_array($args[1]))
+			{
+				$line = \Lang::line($args[0], $args[1]);
+
+				if($line != $args[1])
+				{
+					$args = array($line);
+				}
+			}
+
+			parent::__construct(\call_user_func_array('\sprintf', $args));
+		}
+	}
+}
+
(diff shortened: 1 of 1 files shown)


jschreuder:
We talked about this earlier, and decided against it. All core exceptions will never be i18n, they aren't meant for end users and we're not going to help a developer with bad practice. Thus it makes no sense to have core errors translateable.  
There's also another issue with your code: you're changing the method signature of Exception::__construct(), this is almost never a good idea to do with a native method - especially for the most basic exception used in the core.

If you need to have translateable exceptions for other parts of your application: nothing is stopping you to use this class (though I'd suggest under a different name). But it doesn't belong in the core, certainly not as the most basic exception the core might throw.

ghost:
A few concerns I have on this:

1 - Why wouldn't you want to support i18n on core errors? People that don't speak English should be supported as well, shouldn't they?
2 - I understand not calling \Fuel_Exception() directly, but what about in cases when you have 2 or 3 exceptions for a package that should follow the same logic? Wouldn't supporting this for those exceptions, and having those exceptions extend the core be the goal?
3 - What do you mean I'm changing the method signature?

jschreuder:
1. As I said core errors are only meant for the dev, and we can't imagine any dev not having at least a rudamentary understanding of English. For the very few there are it's not worth the overhead and time it would take to keep everything translated.
2. I have no clue what you mean here, and I don't really see a reason for putting this in the core.
3. The native `Exception::__construct()` method signature is:

``` php
public function __construct($message = null, $code = 0, Exception $previous = null)
```

ghost:
To elaborate on 2 a bit:

If I have the exception:

```
class MyPackageException extends \Fuel_Exception {}
```

And I want to utilize Language within that exception, Fuel_Exception would utilize the i18n, so I wouldn't need to within my own exception.

```
throw new MyPackageException('language.line', array('whatever' => 'ohpleaseyes'));
```

Does that make sense?

jschreuder:
That's fine, but I don't see a reason to include this in the core. It's almost no work at all to create this yourself and extend it from your own exception.

ghost:
I understand that. But making everyone else's lives easier was the goal here.

jschreuder:
But then I refer back to your original proposal. You are adding a new class to save 5 characters.

``` php
throw new \Fuel_Exception(__('payments.required', array('fields' => implode(', ', $fields))));
throw new \Fuel_Exception('payments.required', array('fields' => implode(', ', $fields));
```

This isn't something that will save anyone huge amounts of time, as anyone who needs it can code it up within 5 minutes. Thus it's just a bit too trivial to go into the core. That's not to say that it's useless, just not core material.

PR #306, opened 2011-07-26
Develop PHPUnit test fix

ghost:
There is an error when running the unit tests from the core. This is due to the data variable in tests/format.php being static.

commit message:
Changed the $data variable from static to non static and put the initialization code in setUp function.

### DIFF
--- a/tests/format.php
+++ b/tests/format.php
@@ -20,10 +20,15 @@ namespace Fuel\Core;
  */
 class Test_Format extends TestCase {
 	
-	protected static $data = array(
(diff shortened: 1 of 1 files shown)


jschreuder:
Any specific reason the data array can't just be defined in the class definition?

ghost:
It was giving me an error, I don't remember which one. I will check it when I can.

dhrrgn:
This should use a data provider.  See https://github.com/fuel/core/blob/develop/tests/arr.php for how to use it.  See here for the official docs (scroll to Data Providers) http://www.phpunit.de/manual/3.2/en/writing-tests-for-phpunit.html

PR #313, opened 2011-07-26
Removes a Notice in session/cookie

Savageman:
Just a tiny fix.

### DIFF
--- a/classes/session/cookie.php
+++ b/classes/session/cookie.php
@@ -30,7 +30,7 @@ class Session_Cookie extends \Session_Driver {
 	public function __construct($config = array())
 	{
 		// merge the driver config with the global config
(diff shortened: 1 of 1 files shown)


WanWizard:
If you have this issue, you're loading a session drivers without required configuration.

Which you shouldn't. I'm more in favor of throwing an exception, who knows what else is missing?

Savageman:
Well, here it's merged with a default config. So the cookie_name is set.
If an exception is it should be done in Session_Driver::_validate_config() I think. Something like this?

if (count(array_diff(array('keys', 'to', 'be', 'existing'), array_keys($validated))) > 0) {
    throw new \Fuel_Exception("The session couldn't be loaded properly, please check your configuration.");
}

PR #314, opened 2011-07-27
JSONP support

rob-mccann:
Sorry if I've done this wrong, git is all a bit new to me!

I needed crossdomain support for my API but noticed that there wasn't built in jsonp support so I added it here.

It's just a few lines to the Format and single line to controller_rest.

The callback variable can be set by setting a variable called 'callback' in the get or post request. It defaults to 'response'.

I didn't add a from_jsonp because I don't think people will need it often.

Do let me know if I've done something wrong so I can learn :)

Rob :)

Related forum thread:
http://www.fuelphp.com/forums/topics/view/3648


philsturgeon:
Wow this is a messy pull request! I'm not sure why but you've got hundreds of commits listed. Can you try again? The changed look good other than that.

nerdsrescueme:
Can I suggets a slightly more concise syntax if this is how it will be implemented?

``` php
/**
 * To JSONP conversion
 *
 * @param mixed $data
 * @return string
 */ 
public function to_jsonp($data = null)
{
    $callback = \Input::get_post('callback', null)); 
    is_null($callback) and $callback = 'response'; 

    return $callback.'('.$this->to_json($data).')';
}
```

PR #322, opened 2011-07-28
Bug fix with Form::select() selected value

ghost:
While using some code I found in @dhorrigan 's scrapyrd, I found a bug with Form::select() when defining the selected value.

This code can be found at: APPPATH/views/scraps/view.php line 3 of scrapyrd:

```
<?php echo Form::select(array(
    'name' => 'style',
    'id'    => 'style',
    'options'   => array(
        'ascetic'   => 'Ascetic',
        'dark'      => 'Dark',
        'default'   => 'Default',
        'far'       => 'Far',
        'github'    => 'GitHub',
        'idea'      => 'Idea',
        'ir_black'  => 'IR Black',
        'magula'    => 'Magula',
        'sunburst'  => 'Sunburst',
        'vs'        => 'Visual Studio',
        'zenburn'   => 'Zen Burn',
    ),
    'selected'  => 'idea',
)); ?>
```

The `'selected' => 'idea'` portion of that code was not functioning properly, but does with this fix.

Hope this is good stuff :P I'm basing the bug off of code dan wrote himself :P

### DIFF
--- a/classes/form.php
+++ b/classes/form.php
@@ -470,7 +470,7 @@ class Form {
 		{
 			$attributes = $field;
 
-			if(!isset($attributes['selected']) or empty($attributes['selected']))
+			if ( ! isset($attributes['selected']))
 			{
 				$attributes['selected'] = ! isset($attributes['value']) ? null : $attributes['value'];
 			}
--- a/classes/form.php
+++ b/classes/form.php
@@ -469,7 +469,11 @@ class Form {
 		if (is_array($field))
 		{
 			$attributes = $field;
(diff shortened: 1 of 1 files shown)


jschreuder:
The logic is a bit off:

```
if(!isset($attributes['selected']) or empty($attributes['selected']))
```

If it's empty it's also not set, empty will supress any warnings just like isset does. But it should be possible to be empty because 0 (zero), '0' and '' are all valid values to select. This should only happen on `! isset($attributes['selected'])`.

Also it's not yet up to our coding style guide: there must be a space after the `if` and there must be spaces on both sides of the `!`.

ghost:
Kdone :)

PR #323, opened 2011-07-28
Allow passing only an array to set_attribute

utahcon:
This change allows passing of just an array of key/value pairs to set_attribute. 

Looks like it was intended this way, but for some reason was not setup to allow it.

I know this only throws a warning, but it seemed to make sense to just fix it.


jschreuder:
I merged it without paying attention while I shouldn't have: you pull-requested on master and that's not allowed. The master branch always contains the latest release and nothing more. You should edit & pull-request on the develop branch.

utahcon:
Oh yeah, sorry. Will do that.

PR #333, opened 2011-07-31
Fixing Html::img() not to include index.php.

huglester:
Fixing Html::img() not to include index.php.
Not sure if it's the best way


dhrrgn:
The best way would actually be to do it this way:

\Uri::base(false).$src;

I will make the change and give you credit.  Thanks.

PR #334, opened 2011-07-31
This change is to allow developers to pass data as objects in addition to arrays.

NateRadebaugh:
$data->test = 'value' calls get_object_vars($data), which converts it into an array. $data['test']='value'

If an object is sent, convert it into an array. If something other than
an object or array is sent, convert it to an array.

### DIFF
--- a/classes/view.php
+++ b/classes/view.php
@@ -69,8 +69,12 @@ class View {
 	 * @param   array   array of values
 	 * @return  View
 	 */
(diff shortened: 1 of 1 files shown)


jschreuder:
It's not in accordance with our [coding standards](http://fuelphp.com/docs/general/coding_standards.html) and we don't accept pull-requests on master, only on develop.

NateRadebaugh:
I hadn't seen the coding standards. Thank you for the tip!

PR #336, opened 2011-07-31
Develop - This change is to allow developers to pass data as objects in addition to arrays.

NateRadebaugh:
Fixed: In accordance with coding standards, submitting on develop branch.

This change is to allow developers to pass data as objects in addition to arrays.
Edit
$data->test = 'value' calls get_object_vars($data), which converts it into an array. $data['test']='value'

If an object is sent, convert it into an array. If something other than
an object or array is sent, convert it to an array.

### DIFF
--- a/classes/view.php
+++ b/classes/view.php
@@ -71,6 +71,14 @@ class View {
 	 */
 	public static function factory($file = null, array $data = null, $auto_encode = null)
 	{
+		if (is_object($data) === true)
+		{
+			$data = get_object_vars($data);
+		}
+		elseif (is_array($data) === false)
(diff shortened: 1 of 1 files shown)


dhrrgn:
I am going to merge this, but going to change it to throw an exception if $data is not an array or an object.

dhrrgn:
I have changed it here: https://github.com/fuel/core/commit/d3b93139832e49da309cfed18975e0b4f0e7b172

I also moved it to the constructor rather than the factory method so that non-factory syntax will work.  Also allowed for no data to be sent.

PR #342, opened 2011-08-01
Tests fail.

tarnfeld:
Num test methods are all static but use the $this variable.....


tarnfeld:
ahh why does github default to the master branch. fuck sake, sorry.

PR #343, opened 2011-08-01
Tests fail.

tarnfeld:
Num test methods are all static but use the $this variable.....


tarnfeld:
ignore this, found more problems with this test.

PR #345, opened 2011-08-02
Add support socket option for mysql/mysqli in config

SCIF:
Added support handling sockets (for mysql/mysqli drivers) defined in config/db.php


WanWizard:
You can't use is_int() to test the port number, it will fail on "3306" (which is a string).

WanWizard:
We also do not accept pull requests on the master branch. Please correct and resubmit on develop.

PR #346, opened 2011-08-02
Pre-check checkbox fields if value == true

phil-lavin:


jschreuder:
What is this about? Why are we auto checking fields when they have a value other than an empty string? This looks very wrong to me?

WanWizard:
checkboxes can only have an on or off state, and don't really have a value. This saves you from coding an array('checked' => 'checked') and pass it as an option.

I thought it was rather convenient...

jschreuder:
And what if the checkbox has an actual value (what is sent through post when it's checked) and you don't want it checked by default?

WanWizard:
Hmm... You might have a very valid point here...

phil-lavin:
This seems like desirable functionality, especially when using an Orm model with a fieldset however I think a method needs to exist to override this.

jschreuder:
I'm sorry but it just isn't. A checkbox isn't true or false, it either exists or it doesn't. It exists in post when checked, but doesn't when not checked. Thus the value of "1" (which is what true would validate to) doesn't signify the field is true as a value, it is what is passed to signify true **when the checkbox is checked**. I'm sorry, but this auto-checking business is just very wrong magic.

PR #353, opened 2011-08-03
Add 'true' parameter to MongoCursor::count() function.

aprilbriz:


WanWizard:
@springBriz @philsturgeon: Is this still relevant?

aprilbriz:
@WanWizard @philsturgeon Yes. Must set "true" for sending cursor limit and skip information.
http://www.php.net/manual/en/mongocursor.count.php

PR #354, opened 2011-08-03
'persistence' boolean value in comment must be changed.

aprilbriz:


WanWizard:
Any reason why this _must_ be changed?

aprilbriz:
(Sorry, I'm short in English.)
Because FUEL automatically reload the image when 'persistence' value is FALSE.

fuelphp/fuel/core/classes/image/gd.php
and
fuelphp/fuel/core/classes/image/imagemagick.php

These files have below code.

``` PHP
        if ($this->config['persistence'] === false)
            $this->reload();
        return $this;
```

dudeami0:
The comment was a typo, though the default value should be 'false'. If not, people will experience a 'bug' (Looks like a bug, actually not) if they create a small thumb (100x100) then a big thumb (300x300). The big thumb will just be the smaller thumb stretched due to persistence, and I would not see this as default behavior.

aprilbriz:
ok. I agree.

jschreuder:
As far as I can see this can be closed, comment if not and I'll reopen.

PR #364, opened 2011-08-06
Fatal error when running oil test

tarnfeld:
PHP Fatal error:  Call to a member function get() on a non-object in /sites/twtmore/fuel/core/classes/uri.php on line 167

### DIFF
--- a/classes/uri.php
+++ b/classes/uri.php
@@ -133,7 +133,12 @@ class Uri {
 	 */
 	public static function segment($segment, $default = null)
 	{
-		return \Request::active()->uri->get_segment($segment, $default);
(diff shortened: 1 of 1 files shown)


dhrrgn:
These should return `null` instead of `false`.

dhrrgn:
Good catch tho.

tarnfeld:
Sure, will fix. I use tests a lot, no code goes live until the tests pass so i'm fixing all the things that cause tests to fail in the core now... have you seen the state of the num.php test :S :/

dhrrgn:
Haha, ya.  Someone had those in their pull request.  Haven't had time to go through it properly.

PR #367, opened 2011-08-07
CSRF protection using HMAC of session ID

erikbrannstrom:
On the roadmap for Fuel 1.1, the "Multi-Form support for CSRF security" feature has been postponed. The suggested solution for this is to use the available Javascript functions to update the CSRF token upon form submission. I see three problems with this solution.
1. Additional effort required by the developer.
2. Cookies cannot be HTTP Only which reduces security.
3. Javascript must be enabled (a small problem, admittedly).

This patch instead uses the HMAC method described by Barth et. al. in the paper Robust Defenses for Cross-Site Request Forgery (http://seclab.stanford.edu/websec/csrf/csrf.pdf). Instead of using a new nonce for every request, the HMAC of the user's session ID is used as a token. Due to the fact that the token can be regenerated with the same result upon each request there is no need to store the token in a cookie, which in turn solves the three problems listed above.

Two potential problems that could exist with the HMAC method is that leaking the token to an external application would be worse than before, and that there could be issues when an aggresive session ID regeneration scheme is used. The first one isn't an issue unless the developer does something seriously stupid. Secondly, an aggressive regeneration shouldn't be necessary due to the fact that the ID is AES encrypted.

This solution could be further extended in accordance to the recommendations of Barth et. al. by instead using the Referer header when a site is using HTTPS. In any case, I believe this implementation is a superior alternative, so I hope you will discuss it in your team.

### DIFF
--- a/classes/crypt.php
+++ b/classes/crypt.php
@@ -138,6 +138,21 @@ class Crypt {
 
 	// --------------------------------------------------------------------
 
+	/*
+	 * generate the HMAC of a string
+	 *
+	 * @param	string	text to be hashed
+	 * @access	public
+	 * @return	string	message digest
+	 */
+	public static function hmac($value)
+	{
+		// calculate the hmac-sha256 hash of this value
+		return static::safe_b64encode(static::$hasher->hash($value));
+	}
+
+	// --------------------------------------------------------------------
+
 	private static function safe_b64encode($value)
 	{
 		$data = base64_encode($value);
--- a/classes/security.php
+++ b/classes/security.php
@@ -23,19 +23,9 @@ namespace Fuel\Core;
 class Security {
 
 	/**
-	 * @var  string  the token as submitted in the cookie from the previous request
-	 */
-	protected static $csrf_old_token = false;
-
-	/**
 	 * @var  string  the array key for cookie & post vars to check for the token
 	 */
 	protected static $csrf_token_key = false;
-
-	/**
-	 * @var  string  the token for the next request
-	 */
-	protected static $csrf_token = false;
 
 	/**
 	 * Class init
@@ -45,7 +35,6 @@ class Security {
 	public static function _init()
 	{
 		static::$csrf_token_key = \Config::get('security.csrf_token_key', 'fuel_csrf_token');
-		static::$csrf_old_token = \Input::cookie(static::$csrf_token_key, false);
 
 		if (\Config::get('security.csrf_autoload', true))
 		{
@@ -224,127 +213,17 @@ class Security {
 	{
 		$value = $value ?: \Input::post(static::$csrf_token_key, 'fail');
 
-		// always reset token once it's been checked and still the same
-		if (static::fetch_token() == static::$csrf_old_token and ! empty($value))
-		{
-			static::set_token(true);
-		}
-
-		return $value === static::$csrf_old_token;
+		return $value === static::fetch_token();
 	}
 
 	/**
-	 * Fetch CSRF Token for the next request
+	 * Get CSRF Token, which is created by generating the HMAC of the session ID 
 	 *
 	 * @return  string
 	 */
 	public static function fetch_token()
 	{
-		if (static::$csrf_token !== false)
-		{
-			return static::$csrf_token;
-		}
-
-		static::set_token();
-
-		return static::$csrf_token;
-	}
-
-	protected static function set_token($reset = false)
-	{
-		// re-use old token when found (= not expired) and expiration is used (otherwise always reset)
-		if ( ! $reset and static::$csrf_old_token and \Config::get('security.csrf_expiration', 0) > 0)
-		{
-			static::$csrf_token = static::$csrf_old_token;
-		}
-		// set new token for next session when necessary
-		else
-		{
-			static::$csrf_token = md5(uniqid().time());
-
-			$expiration = \Config::get('security.csrf_expiration', 0);
-			\Cookie::set(static::$csrf_token_key, static::$csrf_token, $expiration);
-		}
-	}
-
-	/**
-	 * JS fetch token
-	 *
-	 * Produces JavaScript fuel_csrf_token() function that will return the current
-	 * CSRF token when called. Use to fill right field on form submit for AJAX operations.
-	 *
-	 * @return string
-	 */
-	public static function js_fetch_token()
-	{
-		$output  = '<script type="text/javascript">
-	function fuel_csrf_token()
-	{
-		if (document.cookie.length > 0)
-		{
-			var c_name = "'.static::$csrf_token_key.'";
-			c_start = document.cookie.indexOf(c_name + "=");
-			if (c_start != -1)
-			{
-				c_start = c_start + c_name.length + 1;
-				c_end = document.cookie.indexOf(";" , c_start);
-				if (c_end == -1)
-				{
-					c_end=document.cookie.length;
-				}
-				return unescape(document.cookie.substring(c_start, c_end));
-			}
-		}
-		return "";
-	}'.PHP_EOL;
-		$output .= '</script>'.PHP_EOL;
-
-		return $output;
-	}
-
-	/**
-	 * JS set token
-	 *
-	 * Produces JavaScript fuel_set_csrf_token() function that will update the current
-	 * CSRF token in the form when called, based on the value of the csrf cookie
-	 *
-	 * @return string
-	 */
-	public static function js_set_token()
-	{
-		$output  = '<script type="text/javascript">
-	function fuel_set_csrf_token(form)
-	{
-		if (document.cookie.length > 0 && typeof form != undefined)
-		{
-			var c_name = "'.static::$csrf_token_key.'";
-			c_start = document.cookie.indexOf(c_name + "=");
-			if (c_start != -1)
-			{
-				c_start = c_start + c_name.length + 1;
-				c_end = document.cookie.indexOf(";" , c_start);
-				if (c_end == -1)
-				{
-					c_end=document.cookie.length;
-				}
-				value=unescape(document.cookie.substring(c_start, c_end));
-				if (value != "")
-				{
-					for(i=0; i<form.elements.length; i++)
-					{
-						if (form.elements[i].name == c_name)
-						{
-							form.elements[i].value = value;
-							break;
-						}
-					}
-				}
-			}
-		}
-	}'.PHP_EOL;
-		$output .= '</script>'.PHP_EOL;
-
-		return $output;
+		return \Crypt::hmac(\Session::key('session_id'));
 	}
 }
 


erikbrannstrom:
I understand your points and mentioned the session ID rotation in my original comment. I do however question the validity of these points.

First of all, I can't see the problem of requiring the use of sessions. The Security class will initialize the session if it doesn't already exist, so there is nothing for the developer to think about. The only potential downside I can see is that it needs to store a cookie in the user's browser, but that is no different from the original implementation with nonces and can't really count as a downside.

Secondly, the rotation of IDs. What would be the point of forcing a new session ID every five minutes? I am aware of the benefits of doing this whenever a user logs in, logs out or somehow changes his/her level of access (i.e. session fixation protection), but doing it on a fixed time interval introduces no real advantage as far as I can see.

dhrrgn:
The problem is the Session class can be a little big, so loading in all that if the developer never uses it is wasted overhead.  Not to mention that adds "magic"  we hate "magic" unless there is a very good reason for it.

erikbrannstrom:
If the Session class is bloated, fair enough. I don't really see what's so magic about loading that class though, since the developer needs to explicitly enable the CSRF auto-loading. It adds nothing to the system unless the developer asks for it, which in my book wouldn't constitute magic.

If you believe this isn't a better solution, then so be it. In my opinion though it's more robust, secure and, not to mention, sexier than keeping a (possibly infinitely long) list of tokens.

alexwhitman:
One of the benefits of not using the same token for each form is that it can prevent repeat submissions (back button or multiple submit clicks).  Once the form is submitted the token would be invalidated.  This solution uses the same token, until session ID regeneration, and removes this particular benefit of preventing repeats.

Personally I use multiple tokens marked, in the session, with a time stamp of when it was generated and do garbage collection to keep the list small.

I agree though with Erik that if the developer explicitly uses CSRF tokens, then auto-loading the Session class isn't "magic" but more of a prerequisite.

iquito:
Not being able to use the back-button or just a random possibility of a form submission to fail is really bad for usability / user experience - I think this should be really important aswell, because if users are frustrated with a website everybody loses.

CSRF attacks do not really make sense if someone is not using sessions at all - the whole point is to take over a session or do something in a session context (of somebody else), so linking CSRF to sessions makes sense.

As far as security and usability is concerned the best possibility would be if the CSRF token is connected to the session (as proposed here) and if the session ID would only be changed after an authentication instead of just randomly every 5 (or whatever) minutes. The second part is more the responsibility of the application code, but maybe it can be facilitated by the fuel session class to just give developers the possibility to only change the session ID if really necessary. Because as long as a user is not authenticated, an attacker has no gain of taking over the session.

jschreuder:
Grab the token from the cookie using javascript, that way you can do as many requests and the back button issue is non-existant.

I do think though that we need an extra measure in there to prevent cookie overwriting, the paper did convince me of that. I will need to study it further though. One thing this doesn't seem to account for is the following (and I must confess I'm 100% sure it's possible): what if an ajax request fetches the page, reeds the token and then does the real CSRF attack using the just fetched token? I gotta test this, but if it's possible it completely invalidates this protection method.

Also I would like to point out the Session class isn't "bloated", that would mean it's too big and has loads of unnecessary code. Which is both untrue and not what Dan said.

And last but not least: was this published? The first 2 guys didn't get their masters yet when they wrote this as far as I could find, thus while endorsed by their tutor I wouldn't directly make this gospel by its Stanford origin. That doesn't invalidate anything, but I just want to caution against leaving all critical thinking because some paper says so.

erikbrannstrom:
One of my points was that it is general a bad idea to allow cookies to be accessed by Javascript, since that makes a potential XSS attack worse.

I don't really understand your second paragraph. How would an attacker be able to grab your token using an AJAX request? That would only return their own unique token, not someone else's.

I honestly have no idea what the authors' qualifications are. I just believe the idea presented is a better option than the one currently in use. For real-world uses, I believe Ruby on Rails uses something along the lines of this.

jschreuder:
> One of my points was that it is general a bad idea to allow cookies to be accessed by Javascript, since that makes a potential XSS attack worse.

Doesn't really matter when it's a nonce. Also makes no difference as a real XSS attack might have access to the page contents and the cookie alike.

> I don't really understand your second paragraph. How would an attacker be able to grab your token using an AJAX request? That would only return their own unique token, not someone else's.

The attack would consist of 2 stages. Let's say I'm the attacker and you're the victim. You visit my website and I have a AJAX attack ready that grabs a page from your site and reads from it the token as presented on the page. After that a second ajax request might use the token to do any further attacks. If you read the token from the cookie and never put it in the page it would be impossible to grab it like this.

> I honestly have no idea what the authors' qualifications are. I just believe the idea presented is a better option than the one currently in use. For real-world uses, I believe Ruby on Rails uses something along the lines of this.

To make sure you didn't get me wrong: I wasn't attacking or invalidating the paper. I just wanted to warn against calling it gospel because it has Stanford in the header. We welcome any discussion, especially on security because that's a very important topic for everyone. And I do think it's a very interesting read which I'll dive further into soon to see what we can do with it. At this point I'm not convinced their prefered way is safer than a nonce, and their "downside" of the nonce was very valid but also easily fixed I think.

iquito:
If the CSRF token is connected to the session and the session id and the token change after each authentication by the user, no session fixation and CSRF vulnerabilities remain - at least none that the fuel/php-code can change so far.

The current session id rotation approach is not really secure either - if a user authenticates himself there is always a 5min-window to steal the session, so session fixation is still a risk. And changing the session id every 5min when no authentication occured is just a waste, because it was unnecessary.

jschreuder:
> If the CSRF token is connected to the session and the session id and the token change after each authentication by the user, no session fixation and CSRF vulnerabilities remain - at least none that the fuel/php-code can change so far.
> 
> The current session id rotation approach is not really secure either - if a user authenticates himself there is always a 5min-window to steal the session, so session fixation is still a risk. And changing the session id every 5min when no authentication occured is just a waste, because it was unnecessary.

So basicly you want to make the session ID a nonce? That might work for CSRF for the same reason I prefer a nonce now. But I'm not sure if you wouldn't create huge problems for Session instead. I know @WanWizard spend a long time on getting this reliable, and making session IDs nonces would probably cause huge problems.

iquito:
No the session ID wouldn't be a nonce - both the session ID and the token connected to the session would be permanent - until the user authenticates himself, at that point we change both. CSRF and session fixation both rely on some form of authentication, otherwise there is no gain for the attacker.

For the scenario you created above: ajax-requests cannot be done cross-domain, which would be a requirement for it to work - you cannot access a page of a foreign website directly from your page, so stealing the CSRF token or session ID directly is quite hard.

jschreuder:
> No the session ID wouldn't be a nonce - both the session ID and the token connected to the session would be permanent - until the user authenticates himself, at that point we change both. CSRF and session fixation both rely on some form of authentication, otherwise there is no gain for the attacker.

In that case I don't have to discuss with you separately, I will look into this further and answer you both if I come to a conclusion.

> For the scenario you created above: ajax-requests cannot be done cross-domain, which would be a requirement for it to work - you cannot access a page of a foreign website directly from your page, so stealing the CSRF token or session ID directly is quite hard.

I need to research it as I said, another scenario might be an old Flash version which used to contain huge security risks - though those should be pretty rare by now. The point is that if it's possible in any way to grab the page contents you have unlimited access and all CSRF protection is out.

iquito:
The idea about changing the session ID after every authentication to avoid session fixation (and improve CSRF protection) is also detailed on Wikipedia: http://en.wikipedia.org/wiki/Session_fixation (under the title of "Best solution: Identity Confirmation").

And sure, if you can grab the page contents then any security is lost - but that is not possible unless you can directly control the victims computer.

WanWizard:
For other apps I use the technique described by @whitman, as I'm not too happy with the fact that a valid token is stored client-side, where it can be fetched and (mis-)used without application control.

I also don't agree with the fact that using the browsers back button and resubmitting a form is desired behaviour. I currently specificly use the CSRF token to prevent that, I wouldn't be too happy if that disappears.

erikbrannstrom:
> Doesn't really matter when it's a nonce. Also makes no difference as a real XSS attack might have access to the page contents and the cookie alike.

The nonce being leaked is not the problem, it's the session cookie I'm worrying about. Probably the most common XSS attack is sending document.cookie to a remote server, making it very easy to steal someone's session.

> The attack would consist of 2 stages. Let's say I'm the attacker and you're the victim. You visit my website and I have a AJAX attack ready that grabs a page from your site and reads from it the token as presented on the page. After that a second ajax request might use the token to do any further attacks. If you read the token from the cookie and never put it in the page it would be impossible to grab it like this.

@iquito already mentioned the problem with cross-domain AJAX. If it would work, the whole idea of CSRF protection would be moot since anyone could execute AJAX requests as other users, which would include their CSRF token.

> To make sure you didn't get me wrong: I wasn't attacking or invalidating the paper. I just wanted to warn against calling it gospel because it has Stanford in the header. We welcome any discussion, especially on security because that's a very important topic for everyone. And I do think it's a very interesting read which I'll dive further into soon to see what we can do with it. At this point I'm not convinced their prefered way is safer than a nonce, and their "downside" of the nonce was very valid but also easily fixed I think.

To be honest, I hadn't thought about the Stanford logo :)

erikbrannstrom:
> I also don't agree with the fact that using the browsers back button and resubmitting a form is desired behaviour. I currently specificly use the CSRF token to prevent that, I wouldn't be too happy if that disappears.

To me, this seems like a whole other issue. It is debatable which is the desired behaviour, but CSRF protection shouldn't really have to take that into consideration. It can very well be handled by some other mechanism, for example setting up a time limit on posts.

iquito:
> I also don't agree with the fact that using the browsers back button and resubmitting a form is desired behaviour. I currently specificly use the CSRF token to prevent that, I wouldn't be too happy if that disappears.

Preventing the back button to work or preventing resubmitting a form should be part of the application logic (and only used when necessary) and not a by-product of a CSRF protection - users of a website expect the back button to work, because it's a basic feature which is widely used.

WanWizard:
@erikbrannstrom, the only way to prevent a resubmit of a form is to use an nonce that expires after submit. It might not be the reason the csrf nonce exists, but it does the job quite nicely,

You might debate on whether the correct name is used for the mechanism or not, but I want to prevent ending up with lots of different ways to protect a form (technically, an HTTP POST), we don't want to add more code then needed to do the job.

jschreuder:
@WanWizard Erik is correct, a CSRF token is **only** about preventing CSRF attacks and shouldn't add any app logic unless there's no way around that. But even then it's not a feature but a limitation. The first concern here should be only what is safe and what isn't, all the rest is not important.

WanWizard:
Who was talking about application logic?

My point is that you need a nonce to protect the HTTP POST, and you need to be able to protect multiple at the same time. And very preferably do that without any client-side intervention, or client-side storage.

I'm only stating that the fact that the nonce expires after form submission has an added benefit that I exploit, and I wouldn't be for deploying two different nonces in my forms if I can avoid it.

erikbrannstrom:
@WanWizard An added benefit for you, but the question is if that is the case for everyone. At least one other person in this pull request has stated that he wouldn't want it that way. In my opinion, a function such as this should do its job and nothing else, but this is something you'll have to decide in the Fuel team.

Anyway, I'm with @jschreuder, meaning that the discussion should be more concerned with which approach is more secure while still remaining easy to use for the developers.

dhrrgn:
Umm...A whole lot of theory jumpin around in here.  Was any decision actually made?  If there was I have no clue what it is.

This isn't rocket science folks.  All we need to do is "provide a unique identifier to verify a request as valid."  Plain, and simple.

jschreuder:
@dhorrigan I still need to study this paper fully, just haven't gotten round to it. If nothing changes I'll introduce a check to see if the token passed was tinkered with, otherwise more might change but I doubt I'll be convinced.

erikbrannstrom:
@dhorrigan Creating a framework isn't rocket science. All you need to do is "provide a code toolset that enables the developer to create secure and efficient web applications". Plain and simple. Anyone can do it.

You see what I did there? Don't be an asshole, Dan ;)

Issue #373, opened 2011-08-11
Route with underscores calls wrong controller


iquito:
When calling a page like "/pictures_deliver/favicon" the routing system of fuel tries to find "Controller_Pictures_deliver" instead of "Controller_Pictures_Deliver" - wrong case after underscores.

iquito:
Sent a pull request with my proposed change to fix this.

PR #375, opened 2011-08-11
Bugfix for routing when underscores are used

iquito:
Bugfix when controller for route contains underscores (so uppercase character is used after each underscore). See https://github.com/fuel/core/issues/373 for example/explanation

### DIFF
--- a/classes/router.php
+++ b/classes/router.php
@@ -175,7 +175,7 @@ class Router {
 		$match->directory = array_shift($segments);
 		$match->controller = count($segments) ? array_shift($segments) : $match->directory;
 
-		if (class_exists('Controller_'.ucfirst($match->directory).'_'.Inflector::classify($match->controller)))
+		if (class_exists('Controller_'.ucfirst($match->directory).'_'.str_replace(' ', '_', ucwords(str_replace('_', ' ', $match->controller)))))
 		{
 			$match->action = count($segments) ? array_shift($segments) : null;
 			$match->method_params = $segments;
@@ -189,7 +189,7 @@ class Router {
 		$match->controller = count($segments) ? array_shift($segments) : $match->directory;
 
 		// We first want to check if the controller is in a directory.
-		if (class_exists('Controller_'.Inflector::classify('_', ' ', $match->controller)))
+		if (class_exists('Controller_'.str_replace(' ', '_', ucwords(str_replace('_', ' ', $match->controller)))))
 		{
 			$match->action = count($segments) ? array_shift($segments) : null;
 			$match->method_params = $segments;
--- a/classes/router.php
+++ b/classes/router.php
@@ -175,7 +175,7 @@ class Router {
 		$match->directory = array_shift($segments);
 		$match->controller = count($segments) ? array_shift($segments) : $match->directory;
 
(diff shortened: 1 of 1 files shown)


jschreuder:
Use `Inflector::classify()` instead or use `preg_replace('/(^|_)(.)/e', "strtoupper('\\1\\2')", $class)` this is pretty unreadable.

iquito:
Using preg_replace with the /e modifier is basically using a regular expression and eval for something which can be handled by very fast php-internal functions (str_replace and ucwords). I didn't know Inflector::classify exists, but shouldn't the code there be changed to my solution so it's more efficient?

jschreuder:
The overhead this adds is so incredibly small it's not important and the code is far more readable, your code is incomprehensible and thus bad for maintanence. Which is why a completely negligible overhead is better than 3 function calls that just look bad.

You would be right if this were called in a loop or worse, but it's twice at most.

iquito:
To call my code "incomprehensible" seems really exaggerated - understanding the replacement of a character, calling ucwords and doing another replacement is easy. If you give the preg_replace code and my code to some people uninvolved with fuel I'm sure they need more time to understand what the preg_replace is doing.

I'm trying to improve on fuel here, and I think it's not really great to call code "incomprehensible" just because you think it doesn't look nice enough. I will replace it with Inflector::classify, but I don't think "preg_replace('/(^|_)(.)/e', "strtoupper('\1\2')", $class)" is a good solution - you argue it's not used often enough to have a performance impact, but who knows if Inflector::classify may be used somewhere in a loop at some stage and nobody remembers about the bad performance that brings to the table.

jschreuder:
You're right, I shouldn't have said "incomprehensible" - I just got home from holiday and am working through a massive amount of emails, bug reports, etc and was a bit blunt.

As to your point about `Inflector::classify()`, when would you ever call that in a loop or something? If you're loading that many files/classes the preg call won't be the issue.

iquito:
Hm calling Inflector::classify instead of my code generates a fatal error (without any useful information) on my end...

iquito:
Must be something in the static::singularize-call inside Inflector::classify which throws a fatal error, so Inflector::classify cannot be used after all.

In that case I'm changing it back to 

```
str_replace(' ', '_', ucwords(str_replace('_', ' ', "
```

I still think this is easier than

```
preg_replace('/(^|)(.)/e', "strtoupper('\1\2')", ", 
```

and using preg_replace with the /e modifier also does not work with suhosin, because any eval is assumed to be unsecure and is therefore blocked. Another possibility would be 

```
preg_replace_callback('/(^|_)./', function($matches) { return strtoupper($matches[0]); },
```

but is that really easier and nicer?

jschreuder:
It's still bad practice and code repetition, and you didn't apply it everywhere the controller might have underscores. I think you do have a point in efficiency to use the 3 functions, especially as I think most opcode caches slow down from using eval code. But your way still leads to repitition I don't like. Thus I'll fix up Inflector::classify() to use your way and make the Router method use the Inflector class.

jschreuder:
Forgot to reference this pull request in the commit, it's 81ea6602191fcd34efe9

iquito:
Ok, this outcome seems like the best solution, so I'm happy with it :-)

PR #376, opened 2011-08-11
No default locale instead of en_US

iquito:
If a user does not set a locale in the config file, no locale should be used by default. Setting a locale is not required by PHP, and there are some possible problems with it as described in http://fuelphp.com/forums/posts/view_reply/3557 .


jschreuder:
@dhorrigan you told him to submit this, could you take a look?

jschreuder:
You're able to switch it off, as was done already with https://github.com/fuel/core/pull/304 - that's enough.

iquito:
Why should it be necessary to switch something off? As far as the people from fuel always communicate the framework should only do the stuff you expressly want it to do and what you would expect it to do. If you remove the locale-setting from the config, you would expect that no locale is set, not that a default locale of en_US is set instead. That behavior is not even documented, and it is utterly unnecessary.

jschreuder:
Don't remove settings, that's just stupid - they're there for a reason. How the framework might act if you start removing limbs isn't something I really care about as you shouldn't do that. Most things you shouldn't do aren't documented for that same reason.

iquito:
Then why have a default locale which is _only_ set if you remove the line in the config? That speaks even more for not having a default locale.

Even for configs one should aim for the most simple solution - being able to remove a line if it is not needed makes the config shorter, and it is much more intuitive than setting the locale to "false".

jschreuder:
> Then why have a default locale which is only set if you remove the line in the config?

Because it demonstrates the expected type of value.

iquito:
Not really, how locales are named are OS-specific, and not PHP-specific. And the value can be "false" or can be a string - both are valid and could be expected. The comments can easily reflect that, and it's only relevant for people actually working on or with the fuel core.

dhrrgn:
@iquito we could have no defaults for anything at all and just make everyone configure then entire framework itself.  Now, I am sure no one wants to do that...it would take quite a while.  I find it amusing that you are not complaining about the language being defaulted to 'en', or the timezone being defaulted to 'UTC' or the encoding being defaulted to 'UTF-8'.  It's all for the same reason.  You can't expect the default config to appease everyone.

jschreuder:
I'm done with this discussion. As I've said all I care about is if it works as expected and it does. Removing settings from the main config is something you just shouldn't do, and if you do the consequences are not my problem. That's not to say I don't think the framework shouldn't be able to handle such an event, but anything resulting from it is not something I'd consider a bug. I think this looks best in the code, and yes I do give a shit what that looks like when it's edited - a great deal more than I care about what happens if people start removing core settings from their config.

You won't convince me. Ask one of the other devs, I don't care what happens in the end but I won't change it.

iquito:
The timezone needs to be set, otherwise PHP complains about it, and the encoding needs to be known too, so string operations can be made correctly (and PHP should be UTF-8 by default anyway) - so these defaults do not really matter. But the locale is totally optional and not even really helpful, so making it as easy as possible to remove it seems important to me.

This change is not really for me - I went over most of the fuel code just to find out things like this. But for everyone new to fuel it seems much more intuitive to comment out a line in the config if you do not want to use a locale, and it's easier to explain too. So my proposal is purely because I think it would be better for newbies or for someone who is just doing trial-by-error.

PR #381, opened 2011-08-11
Image class fixes. credits for original author (dudeami)

huglester:
Fixed `resize_crop()` bugs, fixed typos in config, chenged one config value as suggested per image class author; fixed imagemagick (almost did not work); fixe Imagick class - did not work at all; fixed Declaration of `Fuel\Core\Image_*::_rounded()` should be compatible with... and similar notices; and others can't remember


jschreuder:
Next time use backticks in your bugreports - otherwise its all italic because of markdown. I did it for you now.

PR #386, opened 2011-08-12
Hotfix/image preset resize

zbrox:
bugfix - when using presets for resizing or crop resizing always getting black images because of bcdiv not calculating the ratios correctly due to type cast strangeness;


WanWizard:
I think you should use (float) instead of (real). 

Real is currently an alias thay might be deprecated (settype already doesn't support it).

zbrox:
I was unsure on whether to use float or real. The docs of PHP said it all casts to float, but I decided to use real for consistency because it was used in other places of the class. Should I change it to float there and in the other occurences of real?

PR #389, opened 2011-08-13
To get the post-array, you now can use \Input::post() instead of \Input::

jeroenbrussich:
http://fuelphp.com/dev-docs/classes/input.html writes for \Input::post()

post($index, $default = null)
$index  required    The key in the $_POST array, or _null_ for the entire array

To get the post-array, you had to use \Input::post(null).
Now you can use \Input::post()

---

Fix also works for \Input::get() and \Input::get_post()

### DIFF
--- a/classes/input.php
+++ b/classes/input.php
@@ -136,7 +136,7 @@ class Input {
 	 * @param	mixed	The default value
 	 * @return	string
 	 */
-	public static function get($index, $default = null)
+	public static function get($index = null, $default = null)
 	{
 		return static::_fetch_from_array($_GET, $index, $default);
 	}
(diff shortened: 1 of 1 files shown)


jschreuder:
Like this it's inconsistent as `Input::delete()`, `Input::put()` and `Input::cookie()` would need it as well.

I'd like @dhorrigan to weigh in before accepting though, I remember it was done on purpose to only allow this only with explicit `null` input but I'm not sure.

jeroenbrussich:
Ijschreuder,

I deliberately only changed those methods of which the docs said '$index required The key in the $_POST array, or _null for the entire array_'. I did saw the other methods ;)

I also considered if this was done on purpose, but I thought we all were lazy coders here :)

philsturgeon:
The way this is suppose to work is not only should it check for null but it should also count func_num_args() == 0. This means it will not return a whole array if you are dynamically looking for a key which just so happens to be null.

We dont want to get a whole array back if we accidentally pass it a busted variable. Can you please make this change to post/get/put/etc while you're at it?

jschreuder:
I'll merge this and fix it up like @philsturgeon said.

PR #391, opened 2011-08-13
Added support for file renaming in Upload::process

frankdejonge:


WanWizard:
can't access code at the moment, so I can't check.

But wouldn't that create a problem with multiple uploads and unique filenames not enabled?

PR #402, opened 2011-08-16
Adding support for openmulti()

utahcon:
I am lazy, this is easier.

### DIFF
--- a/classes/form.php
+++ b/classes/form.php
@@ -130,7 +130,24 @@ class Form {
 
 		return \Config::get('form.'.$key, $default);
 	}
-
+	
+	/**
+	 * Create a form open tag with multipart support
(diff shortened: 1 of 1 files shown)


huglester:
I don't think such comments are accepted...
and be sure you pull to develop, because now you are pulling to the master which for sure won't be  accepted :)

philsturgeon:
No we decided against this because it is creating abstraction for no reason. When you do this in HTML you create a form and send it an extra attribute, which is what you do now. It's more natural to developers this way.

Seeing as this had a no vote last time I will action the no vote this time. Plus this was made to the wrong branch which would be an automatic "nope" anyway.

Sorry buddy, please don't let this put you off committing again! :)

PR #406, opened 2011-08-17
Multiple CSRF tokens in Session

alexwhitman:
As a previous CSRF pull request was denied I thought I'd have a go.

Random tokens get generated and stored in the session.  Tokens can have an expiry specified by security.csrf_expiration in seconds.  If set to 0 tokens are perpetual until they are used.  I'm also going to send a pull request for fuel/fuel with changes to config.php for this pull request.

### DIFF
--- a/classes/security.php
+++ b/classes/security.php
@@ -246,7 +246,7 @@ class Security {
 		$tokens = \Session::get(static::$csrf_tokens_session_key, array());
 		$new_token = md5(uniqid().time());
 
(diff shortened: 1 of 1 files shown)


jschreuder:
Tokens need to be accessible from javascript, if you want to use them like this you're free to do so by extension but this won't go into the core.

Multiple tokens are being considered and while at it we might make the way its stored configurable (cookie or Session), but not like this.

PR #408, opened 2011-08-20
Patch 1

ghost:
added method get_one.


jschreuder:
We don't accept pull-requests on master

PR #410, opened 2011-08-21
Fixes #407

### LINKED ISSUE #407: Shutdown events in reverse order mean session data cannot be changed on shutdown
Is this strictly necessary? The reason I ask is this causes the session save to happen before anything user code has added. This means that a user defined shutdown function cannot modify session data as it's already been saved.

phil-lavin:


dhrrgn:
Merged here https://github.com/fuel/core/commit/823731ac686cbca3c1baee28a3165f48755de6bb

PR #422, opened 2011-08-24
Closures as routes

erikbrannstrom:
Route objects now have a property called 'closure' which is null by default. If the value of the route is an anonymous function, that property is set to the closure. The Request class checks this property to decide if a controller/action search is necessary or not.

The closure can have one parameter, which will be passed the Request object. Named parameters will therefore be available, while "regular" parameters (using $1, $2 etc) are not, since their use is more about re-routing than passing values. They could probably be implemented, but then you guys would have to make a decision on how they should work with closures.

I've tested this with a closure throwing a 404 exception, returning only a string and returning a full View object, and all seems to work fine, but I suppose you might have some more tests you would like to run.

Finally, it might seem like the Request class has gotten a full makeover, but you'll see that I've mainly just moved some things around to make it all work smoothly.

### DIFF
--- a/classes/request.php
+++ b/classes/request.php
@@ -339,83 +339,99 @@ class Request {
 			throw new \Request404Exception();
 		}
 
-		$controller_prefix = '\\'.($this->module ? ucfirst($this->module).'\\' : '').'Controller_';
-		$method_prefix = 'action_';
-
-		$class = $controller_prefix.($this->directory ? ucfirst($this->directory).'_' : '').ucfirst($this->controller);
-
-		// If the class doesn't exist then 404
-		if ( ! class_exists($class))
+		if ($this->route->closure !== null)
 		{
-			static::reset_request();
-			throw new \Request404Exception();
-		}
-
-		logger(Fuel::L_INFO, 'Loading controller '.$class, __METHOD__);
-		$this->controller_instance = $controller = new $class($this, new \Response);
-
-		$this->action = $this->action ?: (property_exists($controller, 'default_action') ? $controller->default_action : 'index');
-		$method = $method_prefix.$this->action;
-
-		// Allow override of method params from execute
-		if (is_array($method_params))
-		{
-			$this->method_params = array_merge($this->method_params, $method_params);
-		}
-
-		// Allow to do in controller routing if method router(action, params) exists
-		if (method_exists($controller, 'router'))
-		{
-			$method = 'router';
-			$this->method_params = array($this->action, $this->method_params);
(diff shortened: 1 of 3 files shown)


jschreuder:
Not too sure if this is the right way, it feels a bit "hacky".

For discussion (@dhorrigan): how about moving the class finding & method detection to the Route object and have the Route object always return something callable? (either `array($controller, $method)` or Closure)

Edit: only question on how to deal with before()/after() methods, possibly by checking if the callable is an array and checking those methods on the object (index 0 of the array).

dhrrgn:
@jschreuder I don't think this is hacky.  It makes sense to me.

Moving the class finding to the route object and returning a callable in the Route then returning that?  That seems way more hacky.

I am a +1 on this.

jschreuder:
@dhorrigan why more hacky? That would allow you to return anything from a route to call, far more flexible: this hardcodes support for only our Controllers and closures; my way makes pretty much everything possible.

dhrrgn:
@jschreuder The way you are proposing, like I said, would force us to instantiate the controller inside the Route object, then return the callable of `array($controller, $method)`.  How would we get the controller instance back to the request?  `Request::active()->set_controller($controller);`?  That is way more hacky.  A request can only be made to a controller or a closure...why would we need to route to anything else...the point is to NOT have anything else routable.

jschreuder:
@dhorrigan you'd return the callable as array($controller_instance, $method), as we call the method on an instantiated controller. Thus the controller would be instantiated by the route and returned to it, or the controller would be a property of the Route object which is set as controller_instance during the execute method (remaining `null` for a closure).

I don't see a reason to limit it to Controllers & Closures, I probably wouldn't use something else but I don't see a reason for the limitation.

jschreuder:
@erikbrannstorm we talked it over, my idea is a bit too big of a change pre-2.0 so we'll merge your way. We'll just rename ->closure to ->callable to signal that any type of callable is fine when you're writing your own Route object (only Closures supported by default)

erikbrannstrom:
@jschreuder Awesome, and that change makes sense. Quick question though, shouldn't `if ($this->translation instanceof \Closure)` on line 103 in Route be `if (is_callable($this->translation))` then also? Or is that what you mean by only supporting Closures by default?

jschreuder:
Copy paste from IRC for anyone reading along: but basicly the reason is that arrays are already used for other types of advanced routing - trying to add callable support to that would get very messy (same goes for strings that are also function names).

PR #426, opened 2011-08-25
load_saved() method in image driver to return saved image

ronan-gloo:
Maybe an idea to return the new created image instead of call Image::load() again ?

i.e of use:

$thumb = Image::load('test.png')
  ->crop_resize(50, 50)
  ->save('test_thumb')
  ->load_saved();

// $thumb->sizes() returns:

stdClass Object
(
    [width] => 50
    [height] => 50
)

### DIFF
--- a/classes/image/driver.php
+++ b/classes/image/driver.php
@@ -20,6 +20,7 @@ abstract class Image_Driver {
 	protected $image_directory = null;
 	protected $image_filename  = null;
 	protected $image_extension = null;
+	protected $new_fullpath		 = null;
 	protected $new_extension   = null;
 	protected $config          = array();
 	protected $queued_actions  = array();
@@ -607,7 +608,9 @@ abstract class Image_Driver {
 		{
 			throw new \Fuel_Exception("Could not set permissions on the file.");
 		}
-
+		
+		$this->new_fullpath = $filename;
+		
 		$this->debug("", "Saving image as <code>$filename</code>");
 		return array(
 			'filename' => $filename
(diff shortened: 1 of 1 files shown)


jschreuder:
This looks wrong to me, no object should change what it represents. If it represents the original image it shouldn't be able to change its identity to the new image, if you add such a method it should return a new object representing the new image not change the identity of the old. But I don't think this is the right way either way, the save() method shouldn't return `$this` but should return an instance representing the image that was saved.

ronan-gloo:
The driver already has a public method named load(), which allow you to load any file you want with the same instance. (reload() also allow you to get modified image)
I understand your point, but considering how class is already designed, this make sense to my opinion, until you ask explicitly to load modified image, whenever if it's a new file or not.

dhrrgn:
I mostly agree with @jschreuder.  This is pointless...just do `Image::load` again.  Where I disagree is that save() should return a new object.  It returning `$this` has helped me a lot.  Example (actual code from live app...path name changed):

``` php
    $img = Image::load(TMPPATH.Input::post('filename'))
                ->crop(Input::post('x'), Input::post('y'), Input::post('x2'), Input::post('y2'))
                ->resize(640, 480)->save(PHOTO_PATH.$filename)
                ->resize(174)->save(PHOTO_PATH.$thumb_filename);
```

Makes for super easy multiple saves.

jschreuder:
@dhorrigan but isn't this just as good, and maybe even more logical?

```
$img = Image::load(TMPPATH.Input::post('filename'));
$img->crop(Input::post('x'), Input::post('y'), Input::post('x2'), Input::post('y2'))->resize(640, 480)->save(PHOTO_PATH.$filename);
$img->resize(174)->save(PHOTO_PATH.$thumb_filename);
```

I think it's more logical to have the save() method return what it actually saved, instead of its origin. Also if the Image lib is currently capable of changing an instance's identity we should fix that asap.

bproctor:
I tend to agree with @jschreuder.  Having the save() method return what was actually saved seems like it would in general be more useful.  I think his code example makes a little more sense as well because each save() is in it's own statement.

jschreuder:
@dudeami as you wrote most of the Image class, could you weigh in?

dudeami0:
The persistence config option is what I think your looking for. It will allow you to keep modifying the image after its been saved.

PR #431, opened 2011-08-28
add Asset::from to load remote javascript (or css)

ghost:


philsturgeon:
I don't get it. Asset::css() should cover this. If it doesn't, thats a problem.

bencorlett:
I use Asset::js('https://ajax.googleapis.com/ajax/libs/jquery/1.6.2/jquery.min.js'); all the time? It runs \Uri::create() on the path which obviously has support for external links.

jschreuder:
@virtualpain you'll have to add more than a commit or we'll close this, Ben & Phil pointed out that this seems to be duplicate functionality.

Also, if this would be something new you should be using the `html_tag()` function like all the other Asset methods.

Issue #437, opened 2011-08-31
Uri::current(), Uri::main() says http instead of https


bproctor:
Uri::current() and Uri::main() both return http:// when visiting called from an https:// site.

bproctor:
a couple of things I should have mentioned.  I'm running nginx and in my config, base_url = null.  I'm looking into what might be causing this...

jschreuder:
This is how it is detected: https://github.com/fuel/core/blob/1.1%2Fdevelop/classes/input.php#L191 - might help you figure out why it fails for you.

bproctor:
Ah yes, nginx doesn't set $_SERVER['HTTPS'] by default.  Thank you.

jschreuder:
Closing this as you have a pull-request on the same issue.

PR #438, opened 2011-08-31
HTTPS detection fix for nginx server

bproctor:
This allows Input::protocol() to function properly on nginx.

nginx does not set $_SERVER['HTTPS'] by default.  You have to check the server port to see if it's HTTPS or HTTP.  The best solution would be to just set the HTTPS variable in the nginx config file, but I don't think most people would know to do that.


jschreuder:
Related: https://github.com/fuel/core/issues/437

PR #447, opened 2011-09-01
Inline error reporting for Form

litemotiv:
Adds simple configurable inline error reporting to the Form class, consisting of a css-class added to invalid fields and a span containing the error message.


jschreuder:
Basicly the idea is good, but needs some changes before I'll merge it. The error message method does nothing and can be removed. Inline errors can be fetched from Validation and need not be set as properties. Also they shouldn't have a default markup, that's what the templates are for. And they shouldn't be included in the default template.

litemotiv:
Ok i addressed the earlier problems and simplified the procedure, the error_msg method is gone and error messages are no longer set as attributes but rendered inline using field->error(). I changed the inline_errors config setting in the template to False so that it isn't run by default, but left the {error_msg} template tag in place. This way users only need to set inline_errors to True for working inline error reporting.

jschreuder:
I'll merge it now, but some things will change:
- I'll add an error message template config, it shouldn't be put into anything by default (span tag or whatever)
- I don't think it's best to give the error class to the field, one might want it in the label instead or whatever - it's better as a template value that is either empty or filled and you can give it to a wrapping tag and use css selection to style the label or the field or everything.

jschreuder:
The changes: ac080ec086b9a4515df6d8af8b89a90cd749a06c (and a small one after that but that didn't involve error messages)

jschreuder:
And almost forgot this change: aa34e2e59646c007bfc81ea9a679b0bc3576768a

litemotiv:
I see there is a new hardcoded table tag introduced in form.php, i'm not sure if that's supposed to be there?

jschreuder:
Defaults for templates aren't hardcoded, they're defaults. I introduced the table tag here because the old template was useless and this at least can be output directly for testing purposes.

litemotiv:
I mean in /classes/form.php line 600, perhaps the table tags there were meant to become template fields as well?

jschreuder:
First param of `Form::get_config()` is the key it'll fetch, second is the default when that key isn't found.

litemotiv:
Ah ok, got it. Thanks. :-)

PR #449, opened 2011-09-02
Fixed Oil returning HTML 404 message when there an error exception

crynobone:
Fixed Oil returning HTML 404 message when there an error exception on production environment
### Sample on production

```
$ php oil r migrate
<!DOCTYPE html>
<html>
<head>
    <meta charset="utf-8">
    <title>Fuel PHP Framework</title>
    <style type="text/css">
        * { margin: 0; padding: 0; }
        body { background-color: #EEE; font-family: sans-serif; font-size: 16px; line-height: 20px; margin: 40px; }
        #wrapper { padding: 30px; background: #fff; color: #333; margin: 0 auto; width: 800px; }
        h1 { color: #000; font-size: 55px; padding: 0 0 25px; line-height: 1em; }
        .intro { font-size: 22px; line-height: 30px; font-family: georgia, serif; color: #555; padding: 29px 0 20px; border-top: 1px solid #CCC; }
        p { margin: 0 0 15px; line-height: 22px;}
    </style>
</head>
<body>
    <div id="wrapper">
        <h1>Oops!</h1>
        <p class="intro">An unexpected error has occurred.</p>
    </div>
</body> 
```

I don't think it would be relevant to show a generic error message for oil, as the oil should and would only be accessible from developer or server admin

### DIFF
--- a/classes/error.php
+++ b/classes/error.php
@@ -53,7 +53,7 @@ class Error {
 			logger(Fuel::L_ERROR, $severity.' - '.$last_error['message'].' in '.$last_error['file'].' on line '.$last_error['line']);
 
 			$error = new \ErrorException($last_error['message'], $last_error['type'], 0, $last_error['file'], $last_error['line']);
-			if (\Fuel::$env == \Fuel::PRODUCTION AND !\Fuel::$is_cli)
+			if (\Fuel::$env == \Fuel::PRODUCTION AND ! \Fuel::$is_cli)
 			{	
 				static::show_production_error($error);
 			}
@@ -82,7 +82,7 @@ class Error {
 		$severity = ( ! isset(static::$levels[$e->getCode()])) ? $e->getCode() : static::$levels[$e->getCode()];
 		logger(Fuel::L_ERROR, $severity.' - '.$e->getMessage().' in '.$e->getFile().' on line '.$e->getLine());
 
-		if (\Fuel::$env == \Fuel::PRODUCTION AND !\Fuel::$is_cli)
+		if (\Fuel::$env == \Fuel::PRODUCTION AND ! \Fuel::$is_cli)
 		{
 			static::show_production_error($e);
 		}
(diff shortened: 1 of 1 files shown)


WanWizard:
That command displays a stack trace and an error message here, so I assume this has been taken care of in the oil package?

crynobone:
This happen during shutdown_event, after Exception has occur.

The code I modify to ensure called from CLI would return php_error instead of production_error

jschreuder:
Real classy, @WanWizard asks you to confirm and you reply by whining on Twitter. Remember we're doing this in our free time just like you, mistakes can happen and non-critical stuff will have to wait till we get round to testing it. When we ask to confirm it's just to allow you to point out where we may have gone wrong.

crynobone:
If you want a verification, why close it? If you want to wait to test it, why close it? This is not a race to get a beta release by hiding possible bug.

jschreuder:
@crynobone not all bugs get closed immediately and it happens often they get solved by a solution for something else. When we get closer to a release we start testing and verifying bugs, you gave a very specific case in which it would fail, WanWizard tested it and didn't get the error - in which case I'd have assumed the same thing. Closing isn't the same as forgetting, we'll reopen (like I just did) if we get a reply to a verification-question but we'll clean up in advance when we don't (which also happens).

In any case, my point was: you could have just answered the question posed. This is helping nobody.

WanWizard:
@crynobone if that was the problem, then the issue wasn't very descriptive. 

Your sample had nothing to do with an error in a shutdown event. I closed it because the issue in the ticket had been solved. i can't smell that there might be more to it.

crynobone:
Test case/Reproducing step: http://scrp.at/aCA
1. Set environment to PRODUCTION in app/bootstrap.php
2. Run refine/migrate/generate using php oil that generate an Exception
3. You get a HTML in your Terminal.

Next time I'll try to use better description, maybe "Fixed Oil returning HTML 404 message when there an error exception on **PRODUCTION** environment"

WanWizard:
This should do:

```
/data/www/virtual/fueltest  (on 1.1/develop)
[hverton@catwoman] $ php oil r foo
Error - I'm fucked in APPPATH/tasks/foo.php on line 8
```

PR #452, opened 2011-09-02
Related to 74 of ORM - Temp callables

phil-lavin:
Related to 74 of ORM - Allows callables to be removed and allows temp callables to be passed to run()

### DIFF
--- a/classes/validation.php
+++ b/classes/validation.php
@@ -238,6 +238,24 @@ class Validation {
 		return $this;
 	}
 
+	/*
+	 * Remove Callable
+	 *
+	 * Removes an object from the callables array
+	 *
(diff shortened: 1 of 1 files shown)


phil-lavin:
This is associated with it:

https://github.com/fuel/orm/pull/80

jschreuder:
This is pretty much what I had in mind, but I want to think on your solution for a bit. An improvement might be to add something like this when checking if a class is already in the callables:

```
foreach ($this->callables as $key => $c)
{
    if (is_string($temp_callable) and is_a($temp_callable, $c))
    {
        unset($this->callables[$key]);
    }
}
```

[edit:] The remove_callable() method is a good addition, but not strictly necessary to use here in the class when done like above.

PR #457, opened 2011-09-06
Add ability to chain session methods

bencorlett:
Session::set('foo', 'bar')->('userid', $userid);

### DIFF
--- a/classes/session/cookie.php
+++ b/classes/session/cookie.php
@@ -30,7 +30,7 @@ class Session_Cookie extends \Session_Driver {
 	public function __construct($config = array())
 	{
 		// merge the driver config with the global config
-		$this->config = array_merge($config, isset($config['cookie']) && is_array($config['cookie']) ? $config['cookie'] : static::$_defaults);
+		$this->config = array_merge($config, isset($config['cookie']) and is_array($config['cookie']) ? $config['cookie'] : static::$_defaults);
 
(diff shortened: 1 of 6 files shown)


WanWizard:
Not a bad idea. However, it is not complete.
- the Session class docblocks needs to be modified accordingly
- some of the driver methods are overloaded by the specific backend drivers, they need to be modified too

bencorlett:
Hi @WanWizard,

What do you mean by
- The Session class docblocks needs to be modified accordingly

I've changed what the function returns, will I need to do more than that? I haven't really changed anything aside from what it returns so I thought that http://d.pr/YMDI would be enough?

Looking forward to hearing from you.

Ben

WanWizard:
The session driver's methods can be called through the static methods of the Session class. The docblocks there need to be updated to reflect the correct return type (like you did for the driver).

Also, several of the generic driver methods are overloaded in the specific backend drivers, like for example write(). You haven't changed those, so now it's inconsistent, not every driver behaves the same on all methods.

bencorlett:
Hey, 

I fixed the specific driver overloads this morning, unless there is more I missed?

I'll do those dicblocks tonight when I get home :)

jschreuder:
Next time test a little bit better: https://github.com/fuel/core/commit/2281813509ac01f9dee6d3a5c71a97da70b16ad7

Issue #432, opened 2011-08-29
Format::_from_xml() doesn't work properly with recursion


kriansa:
When you parse some data from xml using Format::factory($data, 'xml')->to_array(), if in your xml there's more than one level, the sub-levels will be parsed as a empty array.

ex:

``` xml
<xml>
    <servername>test</servername>
    <user>
        <name>John</name>
        <address>CA</address>
    </user>
</xml>
```

Will be parsed as:

``` php
array(2){
    'servername' => 'test',
    'user' => array(2){
        'name' => array(0){},
        'address' => array(0){},
    }
}
```

That's because the simplexml_load_string function convert all the input string in objects/sub-objects of SimpleXMLElement class, so casting the main element as array doesn't work well.

digitales:
I'm still new to git, but I have written a fix and tested it locally. Hopefully it will be accepted into the development branch.

kriansa:
I don't think this is the best way to fix it. You should do it from the "source" of the problem (cuz Format::_from_xml() isn't converting xml to array recursively). If you use Format::factory($data, 'xml')->to_json() for example, it's still going to break the sub-levels. I fixed it, i'm just trying to use this git to commit it... (yeah digitales, i'm so newbie as you in git :P)

PR #463, opened 2011-09-07
Fix for #432 - Format::_from_xml() not working with recursion

digitales:
A simple fix for the issue #432 : Format::_from_xml() doesn't work
properly with recursion  The change checks for an object and then converts it on each recursion through the XML DOM.

### DIFF
--- a/classes/format.php
+++ b/classes/format.php
@@ -96,6 +96,11 @@ class Format {
 
 		$array = array();
 
+		if (is_object ( $data))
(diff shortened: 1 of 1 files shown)


jschreuder:
@philsturgeon take note of correct code style also, or fix it yourself after merging.

philsturgeon:
Didn't spot it on a tiny screen. Sorry boss.

PR #469, opened 2011-09-09
Fixed Format::_from_xml() & missing static keyword in DB::count_last_query & Ftp::mirror bug

kriansa:
Commit ddccc2c16fbabcf1d80ff52844ac9a6035286bcf: Real fix to issue #432.

I don't think that digitale's commit (33ea3217a6e86f6827bdd49aecad79737a617f64) is the best way to fix it. You should do it from the "source" of the problem (cuz Format::_from_xml() isn't converting xml to array recursively). If you use Format::factory($data, 'xml')->to_json() for example, it's still going to break the sub-levels.

(Sorry, i'm newbie in git...)

commit 3bd6ccbc18f7f0efcfb6b4422737b5517582ca16: Added a missing static keyword in DB::count_last_query that was throwing a E_STRICT.

commit 2c5635aba0a656ec4459a5fbd30cbe226d855f3c: Added support to set a "inactive" CSS class in << Previous and Next >> links on pagination class, that way it's easier to make custom CSS for these disabled links

commit ad8013d9ab8ca56a3e10d3efa17a8bedb6e95a33: fixed the bug of a missing method on Ftp::mirror


kriansa:
I'm sorry for this bunch of commits, but I didn't know that every push on my fork updates my pull request too.

I updated my first post.

philsturgeon:
Sorry I'm going to have to cancel this pull request as it contains too many other changes. Please can you make a new branch for each and every change you make to avoid this happening?

Also if you could show off some examples that would be really handy. Show what is broken and what is working, maybe sending in some unit tests to show it off?

Thank you! :)

PR #475, opened 2011-09-12
Make field being validated accessible from validation rules

andrewwheal:
Store the field currently being validated in a $_current_validation_field property on the class containing the rule being run.

I originally thought to pass it through as a parameter but this would mean all current validation rules would break. The idea to store it in a parameter came from waldmeister on irc.

To use it with Orm Models the property will either need to be added to the Orm\Model class, or to every model that wants to use custom validation. Custom validation classes would also need the property adding to them.

### DIFF
--- a/classes/validation.php
+++ b/classes/validation.php
@@ -112,6 +112,11 @@ class Validation {
 	 * @var  array  contains validation error messages, will overwrite those from lang files
 	 */
 	protected $error_messages = array();
+	
+	/**
+	 * @var  mixed  contains the field currently being validated
+	 */
+	protected $_current_validation_field = null;
 
 	protected function __construct($fieldset)
 	{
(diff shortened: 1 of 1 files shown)


jschreuder:
I understand what you're trying to do, but this would break any usage of a static class as a callable.

I also think it's the wrong place: it should be set in the run() method based on `$field->name` inside the try-catch block.

The best way is probably a companion method for Validation::active() that give you the fieldname (or `Fieldset_Field` object), something like `Validation::active_field()`.

andrewwheal:
Good point on the static callables - re-working my code, so will have another try when I get onto the validation.

I always have believed in 3rd time lucky ;)

andrewwheal:
Implemented your suggestion @jschreuder, didn't even realise there was a Validation::active() before but it is a much nicer way to do it.

andrewwheal:
@jschreuder, anything wrong with this now or can it be pulled?

jschreuder:
@SicoAnimal merged, can you also add the doc explanations for your new methods?

PR #477, opened 2011-09-12
Added to_url function, creates a string suitable for use in URLs

zoe-edwards:
Accepts a string, removes everything that isn’t a number or letter, converts spaces into hyphens. Optionally change hyphens for something else as part of the function.

I use this in almost every single project I work on, so I’d obviously love to see it as part of Fuel.

### DIFF
--- a/classes/str.php
+++ b/classes/str.php
@@ -285,6 +285,30 @@ class Str {
 			return $args[($next ? $i++ : $i) % count($args)];
 		};
 	}
+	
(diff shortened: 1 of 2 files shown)


jschreuder:
There's already the [Inflector::friendly_title()](http://fuelphp.com/docs/classes/inflector.html#method_friendly_title) method.

zoe-edwards:
Dammit.

PR #478, opened 2011-09-12
Added Arr::key_exists to support dot-notated keys for array_key_exists.

frankdejonge:


dhrrgn:
Where are the Unit Tests?  :P

PR #479, opened 2011-09-12
A new Model_Crud Class

dhrrgn:
I find this class really useful and wanted to throw it out there to see what people thought.

This is an optional Model class which allows all the basic CRUD methods as well as handy finder methods.  The find methods work the same as in the Orm.

Model Objects of whatever that model is are returned for the rows.  So if you have a `Model_Post` class then all the objects that the finder methods for that Model return are of type `Model_Post`.

There is validation in here as well, brought to you by @FrenkyNet.

Everyone please have a look and let me know what you think.

**Note: This is NOT a full Orm, it does NOT do relationships for you.  You would need to write your own methods for that.**
### The example Model_Post class

``` php

<?php

class Model_Post extends Model_Crud {
    protected static $_table_name = 'posts';

    // This is not required and defaults to 'id'
    protected static $_primary_key = 'id';

}
```
### Find by Primary Key:

``` php
<?php

$post = Model_Post::find_by_pk($post_id);
echo $post->content;
```
### Find all by arbitrary column:

``` php
<?php

$posts = Model_Post::find_by('author_id', $author_id);
foreach ($posts as $post)
{
    echo $post->content;
}
```
### Find one by arbitrary column:

``` php
<?php

$post = Model_Post::find_one_by('author_id', $author_id);
echo $post->content;
```
### Magic Methods (same as above basically):

``` php
<?php

$posts = Model_Post::find_by_author_id($author_id);
$post = Model_Post::find_one_by_author_id($author_id);
```
### Updating a record:

``` php
<?php

$post = Model_Post::find_by_id($post_id);
$post->content = Input::post('content');
$post->save();
```
### Creating a New Record:

``` php
<?php

$post = new Model_Post(array(
    'content' => 'fooooooooo',
));
$post->save();
```
### The set() method:

``` php
<?php

$post = Model_Post::find_by_id($post_id);
$post->set(array(
    'content' => Input::post('content'),
    'author_id' => Session::get('user_id'),
));
$post->save();
```
### The Model

``` php
<?php

class Model_User extends Model_Crud {
    protected static $_table_name = 'userss';

    protected static $_rules = array(
        'name'     => 'required|min_length[4]',
        'surname'  => 'max_length[20]',
        'password' => 'required|min_length[8]',
    );

    protected static $_labels = array(
        'password' => 'Super Secret Password',
        'name'     => 'Name',
    );
}
```
### The Controller Action

``` php
<?php

public function action_edit($id)
{
    // Get the user
    $user = Model_User::find_by_id($id);

    $user->name = 'Johnny';
    $user->password = 'short'; // This is to short

    if ( ! $user->save())
    {
        $all_errors = $user->validation()->error();
        $password_error = $user->validation()->error('password');
    }
    else
    {
        // do something on success.
    }
}
```


jschreuder:
Looks good, I would prefer to have the static properties in line with the Orm as well though: `$_table_name` and `$_primary_key` (possibly with multiple pk support?).

Also on having table property defined by default has the risk of "falling through" to the base model property when using `static::$_table`. Example changing it using `static::$table` for `Model_Example` and setting it again in `Model_Another` after which the first model's table is reset as well because neither had the $table property defined in its class and both are using the base Model's property. Failing is better than such problems I think.

dhrrgn:
@jschreuder Updated the variables to be the same and commented $_table_name out to prevent the fall-through issue.  @FrenkyNet added validation which I like how he did it.  May need to comment those variables out as well and check if it isset?

PR #483, opened 2011-09-13
Reverse the logic for Upload::is_valid() status

andrewwheal:
Changed the determination of the validate status to presume invalid until a valid upload is found, rather than the opposite.

### DIFF
--- a/classes/upload.php
+++ b/classes/upload.php
@@ -364,12 +364,12 @@ class Upload {
 		}
 
 		// determine the validate status
-		$valid = true;
+		$valid = false;
 		foreach(static::$files as $key => $value)
 		{
-			if ($value['error'] !== 0)
+			if ($value['error'] == 0)
 			{
-				$valid = false;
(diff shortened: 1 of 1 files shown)


WanWizard:
It's about time a decision is made on how this should operate.

This is the 4th time this logic changes, the last time to exactly the opposite, with the comments that is_valid() should return true if there is nothing to validate. This will revert that back to false, which can have side effects.

andrewwheal:
I have not thoroughly tested the code in my pull request, but from using it it seems right.

I have 2 file uploads...
no files: http://snipr.it/~hl
2 valid files: http://snipr.it/~hm
1 valid file: http://snipr.it/~hn
1 valid, 1 invalid file: http://snipr.it/~ho
2 invalid files: http://snipr.it/~hp

Seems good to me... how about other people?

WanWizard:
I can live with this. I'll check the documentation for updates related to this.

PR #486, opened 2011-09-14
Fixed incorrect query for field modification in DBUtil

frankdejonge:


frankdejonge:
Added other types as suggested by @WanWizard.

PR #494, opened 2011-09-16
[Mongo] Allow select to take single values for includes and excludes rather than 

bproctor:
Allow select to take single values for includes and excludes rather than having to have an arrays. ex: $mongodb->select('foo')->get('foobar');


tomschlick:
Looks good to me.

PR #503, opened 2011-09-19
Allow adding a Fieldset_Field to a Fieldset via add()

kbanman:
Added support for creating a Fieldset_Field instance outside of the Fieldset, then adding it to the Fieldset via the `add()` method.

I have a use case that requires this ability, and although it may not be a common use case, I don't think it interferes with much. Here's what I'm doing: (And please let me know if there's a better way of accomplishing this)

My models tend to share many properties, so I have separated common fields into "sub-models" from which the parent model inherits fields. Currently I have the "sub-models" `Address` and `Person`, and actually `Address` is also a "sub-model" of `Person`, because most people have addresses.

Using Fieldset to generate the forms for this setup poses a challenge; Simply using `add_model()` only adds the fields that are directly in the current model, and doesn't include the fields that are to be inherited from "sub-models" like `Address`. I went the route of implementing my own `set_form_fields()` in my models, which adds the fields for the current model, looks for any sub-models to inherit from, and adds those fields as well, but adds the name to an HTML form array so that the end result is HTML that looks something that looks like this:

```
<input name="core_field1" />
<input name="core_field2" />
<input name="address[addr_field1]" />
```

That way I can later retrieve the fields from POST, separated into the different models ready for validation, and if field names are shared between models, they don't conflict.

So far, so good. Problem is, in my `set_form_fields()`, the line `$instance and $form->populate($instance);` only populates the core fields. My solution to that was to create a separate `Fieldset` for the sub-model, add the fields for the submodel, populate it, then add the fields from that Fieldset into the main Fieldset. This pull request adds the ability to add those fields to another fieldset.

### DIFF
--- a/classes/fieldset.php
+++ b/classes/fieldset.php
@@ -194,6 +194,17 @@ class Fieldset
 	 */
 	public function add($name, $label = '', array $attributes = array(), array $rules = array())
 	{
+		if (is_a($name, 'Fieldset_Field'))
+		{
+			// Make sure it doesn't conflict with existing fields
+			if ($existing_field = static::field($name->name))
+			{
+				\Error::notice('Field with this name exists already, cannot be overwritten through add().');
+				return $existing_field;
+			}
+			$this->fields[$name->name] = $name;
+			return $name;
+		}
 		if (empty($name) || (is_array($name) and empty($name['name'])))
 		{
 			throw new \InvalidArgumentException('Cannot create field without name.');


jschreuder:
Each field can only belong to one Fieldset, that's one of the basic assumptions made in the whole thing and not something I intend to break. While I don't have anything against allowing you to forge your own Fieldset_Field instances which you could input into `Fieldset::add()` I didn't see any use for it as that method is practically already the factory/forge method for Fieldset_Field no reason to add complexity. But you want to move fields from one fieldset to the other, that's not something I think is a very good idea. Just create separate fieldsets for each model and have additional fields added to each fieldset instance.

Your use case I'd personally approach differently. From the basic approach I'm not sure if inheritance is the right way to go, what if you want something to inherit from two branches? My suggestion would be to create a new Model parent class like this (this won't work with property auto-detection, but I'm assuming you don't use that):

``` php
<?php
class Model extends Orm\Model  {
    public function _init()
    {
        if (property_exists(get_called_class(), '_partials'))
        {
            foreach (static::$_partials as $part)
            {
                static::$_properties = $part::load(static::$_properties);
            }
        }
    }
}
```

And have your model like this:

``` php
<?php
class Model_Person extends Model {
    protected static $_properties = array('id', 'first_name' => array('data_type' => 'varchar'), 'last_name' => array('data_type' => 'varchar'));
    protected static $_partials = array('Model_Partial_Address');
}
```

And have the partials like this (first base class, then implemented example):

``` php
<?php
class Model_Partial {
    public function load($properties)
    {
        foreach (static::$_properties as $prop => $settings)
        {
            is_int($prop) ? $properties[$settings] = array() : $properties[$prop] = $settings;

        }
        return $properties;
    }
}

class Model_Partial_Address extends Model_Partial {
    protected static $_properties = array('address_1', 'address_2', 'postal_code', 'state');
}
```

A bit of a basterdized Decorator pattern, but it should work beautifully.

kbanman:
Thanks for taking the time to provide such a thorough response. Hope I can bother you just a tiny bit more:

So you wouldn't put addresses in a separate table? 
From my understanding of your code here, you're simply merging the `$_properties` from the partial to the main model, correct? 
My motivation for separating the address fields into a separate model was to keep from having to add the same 6 fields to every model's table, validation, etc. I could define helper methods, like formatting, in the case of `Address` once, and use them from the various models.
I suppose with some tweaking of `__call()` I could make that work still.

jschreuder:
> So you wouldn't put addresses in a separate table? 

Depends on whether it has any meaning to separate it out. When I put companies and people in a DB both will have addresses but I have no logical reason to put their addresses in different tables, thus doing so would only add complexity. On the other hand, if I ever needed a list of addresses with both people & organizations combined it would make sense to separate them into their own model.

But I think I didn't get what you are doing right now. As I understood it you were using inheritance to share properties, but you are using belongst-to/has-one relations? In that case you could just use `->add_model('Person')->add_model('Address')` on a fieldset and the native `set_form_fields()` method should add the fields of both without trouble.

PR #508, opened 2011-09-20
Add Format::to_html()

kbanman:
Intelligently formats data as an HTML table (or as a list, if there are no keys)

One thing I find to be lacking is an option of ordered vs. unordered list.

### DIFF
--- a/classes/format.php
+++ b/classes/format.php
@@ -293,6 +293,80 @@ class Format {
 	}
 
 	/**
+	 * Convert to HTML (table or unordered list)
+	 *
+	 * @param   mixed   $data
+	 * @param   bool    $humanize_headings
+	 * @return  string
+	 */
+	public function to_html($data = null, $humanize_headings = true)
+	{
+		if ($data === null)
+		{
+			$data = $this->_data;
+		}
+
+		// Force it to be something useful
+		if ( ! is_array($data) AND ! is_object($data))
+		{
+			$data = (array) $data;
+		}
+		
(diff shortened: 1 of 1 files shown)


philsturgeon:
I think it would be better off always formatting this as a table even with no header. Otherwise you could just format as an array and send it to Html::ul()

frankdejonge:
The way I see it, the Format class it for parsing data or converting to another data format. Html is not a data format, therefor not belonging in Format.

philsturgeon:
It's pretty handy for dumping out quick AJAH content. I had it in the CI class but took it out when converting to Fuel as we didn't have a "Table" class. I want it, but not as a ul.

frankdejonge:
I can definitely see the use for it. Just think it would have a better place in the Html class.

philsturgeon:
True, we support to_array() so if the HTML class handles the table output we're good to go.

The main reason for its location in format is that the Rest library can easily output HTML then, which is fairly important for Restful services.

philsturgeon:
Talking to @FrenkyNet we've agreed that Yes this is a useful feature, but probably not for format.

The logic should be in Html class and an extra bit of logic added to the rest class:

```
 elseif (method_exists($this, '_to_'.$this->format))
```

This means we can do some formatting in the rest class when they dont have to be specifically in the Format library. 

Basically we've just asked you to totally recode everything, so what I'll do is close this one and if you do feel motivated to do it as we've asked then please send another pull request, if not... fair enough :)

PR #509, opened 2011-09-21
Added replace functionality to the Arr class

axelitus:
Added some replace functionality that sets item values only if they already exist in the array.
This functions are helpful when some default array is used but also an options array is given and one does not want to clutter the resulting array with irrelevant keys.

### DIFF
--- a/classes/fuel.php
+++ b/classes/fuel.php
@@ -260,9 +260,9 @@ class Fuel {
 	public static function find_file($directory, $file, $ext = '.php', $multiple = false, $cache = true)
 	{
 		// absolute path requested?
-		if (strpos($file, '/') === 0 or strpos($file, ':') === 1)
+		if ($file == realpath($file))
 		{
-			return is_file($file) ? $file : false;
+			return $file;
 		}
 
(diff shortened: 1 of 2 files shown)


ghost:
It might be worth adding the ability to use ::replace_items within ::replace_item, but simply passing a multi dimensional array. Not sure if this is something than can be done, but it's something that would be convenient :)

More or less, it would just be the same logic as ::replace_items, within ::replace_item, but have a check for whether $key and $value are arrays that are the same length.

Regardless of if you do it or not, it's something that I would personally like, and it would make usage easier, while also saving an extra function to load into memory.

axelitus:
Actually, as the functions are right now it won't work right away... the main reason is that the replace_items() function takes the keys and the values in the same array as an assoc array. This could be easily fixed by checking the key and value parameters to be arrays of the same length.

But maybe you're right and it would be better to integrate both functions. I'll work on it and push some commits when done...

axelitus:
Because of some git problems I had this has been changed to Pull Request [548](https://github.com/fuel/core/pull/548)

PR #518, opened 2011-09-26
Add support for full (including regex) reverse routes.

canton7:
Assume we have a route of the form:

```
'blog/:date/(:segment)/(:any)' => 'blog/view/$2/$1'
```

When the Route object for this route is instantiated, create two new
variables, reverse_search and reverse_translation.

In this case, reverse_search would take the form:

```
'blog/view/(.+)/([^/]*)'
```

while reverse_translation will look like:

```
'blog/:date/$2/$1'
```

When Router::get is called, it calls Route#match_reverse for each route,
passing it the destination it has been asked to reverse route.
match_reverse checks this against reverse_search. If it matches, this means
that the current route is capable of pointing to the given destination.
If then uses reverse_search and reverse_translation to construct a uri that
would point towards the given destination.

Router::get has been modified to first check for routes whose name matches
the given destination, then check for reverse routes, then assume the
given destination is a reverse route. This gives named routes priority over
reverse routes.

Named parameters are dealt with by being left alone: Router::get uses
Uri::create, which takes care of them.

This patch does not handle cases where translation is a closure -- since
this is not yet documented, I'm not sure how it should behave. Is it even
possible to construct reverse routes whose translations are closured?

### DIFF
--- a/classes/route.php
+++ b/classes/route.php
@@ -36,22 +36,81 @@ class Route {
 
 	protected $search = null;
 
+	protected $reverse_search = null;
+
+	protected $reverse_translation = null;
+
 	public function __construct($path, $translation = null)
 	{
 		$this->path = $path;
 		$this->translation = ($translation === null) ? $path : $translation;
-		$this->search = ($translation == stripslashes($path)) ? $path : $this->compile();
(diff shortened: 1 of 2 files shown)


canton7:
Actually, this has issues when the route doesn't start replaces at $1. I'll get an update pushed soon

PR #523, opened 2011-09-28
Allow global render() function to pass data and turn off auto sanitizatio

phil-lavin:
Allow global render() function to pass data and turn off auto sanitization

### DIFF
--- a/base.php
+++ b/base.php
@@ -145,13 +145,14 @@ if ( ! function_exists('get_object_public_vars'))
  *
  * @param	string	The view name/path
  * @param	array	The data for the view
+ * @param       bool    Auto sanitize data
  * @return	string
  */
 if ( ! function_exists('render'))
 {
-	function render($view, $data = array())
+	function render($view, $data = array(), $auto_sanitize = true)
 	{
-		return \View::forge($view, $data)->render();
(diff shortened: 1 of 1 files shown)


base.php
```diff
@@ -145,13 +145,14 @@ if ( ! function_exists('get_object_public_vars'))
  *
  * @param	string	The view name/path
  * @param	array	The data for the view
+ * @param       bool    Auto sanitize data
  * @return	string
  */
 if ( ! function_exists('render'))
 {
-	function render($view, $data = array())
+	function render($view, $data = array(), $auto_sanitize = true)
```
frankdejonge: The default values should be the same as those from View::forge, both should be null.

base.php
```diff
@@ -145,13 +145,14 @@ if ( ! function_exists('get_object_public_vars'))
...
 	{
-		return \View::forge($view, $data)->render();
+		return \View::forge($view, $data, $auto_sanitize)->render();
```
frankdejonge: Could you rename $auto_sanitize to $auto_filter to keep it inline with the other param naming?

phil-lavin:
That ok?

frankdejonge:
Merged, made $data default to null too (same as in View::forge).
Thanks!

PR #525, opened 2011-09-29
Fixed #499: Add disabled state and use all available pagination template

### LINKED ISSUE #499: disabled state for pagination
Would it be possible to get a disabled state added to the previous and next buttons? As for that to be controlled by the templates like 'active_state' & 'active_end' are?

crynobone:


crynobone:
There's two similar pull request now, developer really want this feature :P

WanWizard:
Pagination class functionality will be reviewed after v1.1.

Issue #320, opened 2011-07-28
Make fieldsets nesteable and allow custom wrapping tags


jschreuder:
Allow one fieldset instance to be a field in another. For output this also needs support for `<fieldset>` wrapper tags instead of `<form>` and should also allow to use none.

nerdsrescueme:
I really like this idea. Could it be possible simply have it function with "groups", kind of like "areas" in the File class? And perhaps when the output is created it could detect whether it is a child or a root instance and automatically output <fieldset> instead of <form> ?

jschreuder:
I have a feeling you're overthinking this, but thing will become more clear when I design this - which won't be for at least another 3 weeks. This is slated for v1.1.

nerdsrescueme:
Quite possibly, I have a tendency to do that. Can't wait to see what you come up with.

WanWizard:
@jschreuder: status update?

jschreuder:
Already did some preperation and let Dan know I wouldn't get round to this till the weekend.

PR #527, opened 2011-09-30
html fieldset support, preliminary

CoolGoose:
I'm open to improvements :) It's a bit hack-ish right now.

### DIFF
--- a/classes/form.php
+++ b/classes/form.php
@@ -140,17 +140,21 @@ class Form {
 	/**
 	 * Create a fieldset open tag
 	 *
-	 * @param	string	string for the fieldset legend
 	 * @param	string|array	action string or array with more tag attribute settings
+	 * @param	string	string for the fieldset legend
 	 * @return	string
 	 */
-	public static function fieldset_open($legend = null, $attributes = array())
+	public static function fieldset_open($attributes = array(), $legend = null)
 	{
 		$fieldset_open = '<fieldset ' . array_to_attr($attributes) . ' >';
 
-		if (! empty($legend))
+		if ( ! is_null($legend))
 		{
-			$fieldset_open.= "\n<legend>{$legend}</legend>";
+			$fieldset_open.= "\n<legend>".$legend."</legend>";
+		} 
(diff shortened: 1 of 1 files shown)


jschreuder:
If you remove it from the `build_field()` method I'll merge it, `fieldset_open` and `fieldset_close` aren't fields and shouldn't be used like that (I have other plans for that: #320).

CoolGoose:
Should i keep it to be used with $form->add or do you want it like $form->fieldset_open ?

jschreuder:
@Coolgoose read the issue I referred to, it'll be possible to nest Fieldset instances into Fieldset instances and after that the nested Fieldset instances will use the `<fieldset />` tags and it'll be possible to specify form or fieldset tag for the "root" Fieldset instance.

Thus `Form::fieldset_open()` for now (though $form->fieldset_open() would work as well to get the code), real OO support will have to wait and will certainly never be done with the add() method.

CoolGoose:
Like so ?

jschreuder:
Yeah, but I'm not yet sure about the $legend param. It would feel better if this method was more like the `Form::open()` method. But where Form::open() either takes attributes or the action, it would feel wrong to allow here the "legend" instead of the attributes (action and legend are too different). I think it would be best to swap the params and allow "legend" as a key in attributes as well:

``` php
<?php
public static funnction fieldset_open(array $attributes = array(), $legend = null)
{
    ! is_null($legend) and $attributes['legend'] = $legend;
    // etc... (and use $attributes['legend'] after this instead of $legend)
}
```

Also prefix the `!` with a space on line 150, the code guidelines say that they must have spaces on both sides.

CoolGoose:
Better ? :D

jschreuder:
I'll fix up the last things myself: next time use my code as it would have prevented the double code for the `<legend>` tag and `elseif` is one word.

jschreuder:
@Coolgoose Can you also add your methods to docs?

CoolGoose:
Asap i'm going to do a pull on the 1.1-dev docs (today in any case)

PR #530, opened 2011-09-30
Multiple bugs with Arr and Str TestCase

crynobone:
# Arr

```
1) Fuel\Core\Tests_Arr::test_element_when_array_is_not_an_array
InvalidArgumentException: First parameter must be an array or ArrayAccess object.
```

Arr::element() **now** expect the first argument to be an array, InvalidArgumentException will be thrown when set otherwise

```
1) Fuel\Core\Tests_Arr::test_elements_throws_exception_when_keys_is_not_an_array with data set #0 (array('Jack', '21', 200, array('Pittsburgh', 'PA', 'US')))
Expected exception InvalidArgumentException
```

Arr:element() **now** accept $keys as either array or string.

```
2) Fuel\Core\Tests_Arr::test_prepend
Failed asserting that <null> is equal to 
Array
(
    [one] => 1
    [two] => 2
    [three] => 3
)
```

Since Arr::prepend just reference the array and not return it, the output return NULL.
# Str

```
5) Fuel\Core\Tests_Str::test_truncate_not_html with data set #0 (15, 'Lorem ipsum dolor sit amet, consectetur adipiscing elit.')
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-<h1>Lorem ipsum dol...</h1>
+<h1>Lorem ipsum dol...</>

6) Fuel\Core\Tests_Str::test_truncate_is_html with data set #0 (15, 'Lorem ipsum dolor sit amet, consectetur adipiscing elit.')
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-<h1>Lorem ipsum dol...</h1>
+<h1>Lorem ipsum dol...</>

7) Fuel\Core\Tests_Str::test_truncate_multiple_tags with data set #0 (15, 'Lorem ipsum dolor sit amet, consectetur adipiscing elit.')
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-<p><strong>Lorem ipsum dolor sit amet, consectetur adipiscing elit.</strong></p>
+<p><strong>Lorem ipsum dolor sit amet, consectetur adipiscing elit.</strong></p></></></></>
```

Problem cause by mb_substr will return '' when $length = null (Reference: http://www.php.net/manual/en/function.mb-substr.php#77515)

### DIFF
--- a/classes/arr.php
+++ b/classes/arr.php
@@ -587,7 +587,7 @@ class Arr {
 	 */
 	public static function prepend(&$arr, $key, $value = null)
 	{
-		$arr = (is_array($key) ? $key : array($key => $value)) + $arr;
+		return $arr = (is_array($key) ? $key : array($key => $value)) + $arr;
 	}
 
 }
--- a/classes/arr.php
+++ b/classes/arr.php
@@ -587,7 +587,7 @@ class Arr {
 	 */
 	public static function prepend(&$arr, $key, $value = null)
 	{
-		return $arr = (is_array($key) ? $key : array($key => $value)) + $arr;
+		$arr = (is_array($key) ? $key : array($key => $value)) + $arr;
 	}
 
 }
--- a/classes/str.php
+++ b/classes/str.php
@@ -88,7 +88,7 @@ class Str {
 		$encoding or $encoding = \Fuel::$encoding;
 
 		return function_exists('mb_substr')
-			? mb_substr($str, $start, $length, $encoding)
+			? mb_substr($str, $start, (is_null($length) ? strlen($str) : $length), $encoding)
 			: (is_null($length) ? substr($str, $start) : substr($str, $start, $length));
 	}
 
--- a/classes/str.php
+++ b/classes/str.php
@@ -88,7 +88,7 @@ class Str {
 		$encoding or $encoding = \Fuel::$encoding;
 
 		return function_exists('mb_substr')
-			? mb_substr($str, $start, (is_null($length) ? strlen($str) : $length), $encoding)
+			? mb_substr($str, $start, (is_null($length) ? mb_strlen($str, $encoding) : $length), $encoding)
 			: (is_null($length) ? substr($str, $start) : substr($str, $start, $length));
 	}
 
--- a/tests/arr.php
+++ b/tests/arr.php
@@ -555,8 +555,8 @@ class Tests_Arr extends TestCase {
 			'two' => 2,
 			'three' => 3,
 		);
-		$output = Arr::prepend($arr, 'one', 1);
-		$this->assertEquals($expected, $output);
+		Arr::prepend($arr, 'one', 1);
+		$this->assertEquals($expected, $arr);
 	}
 	
 	/**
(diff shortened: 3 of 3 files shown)


classes/arr.php
```diff
@@ -587,7 +587,7 @@ class Arr {
 	 */
 	public static function prepend(&$arr, $key, $value = null)
 	{
-		$arr = (is_array($key) ? $key : array($key => $value)) + $arr;
+		return $arr = (is_array($key) ? $key : array($key => $value)) + $arr;
```
jschreuder: This seems wrong, it should either work by reference or return something - not do both.

classes/str.php
```diff
@@ -88,7 +88,7 @@ class Str {
 		$encoding or $encoding = \Fuel::$encoding;
 
 		return function_exists('mb_substr')
-			? mb_substr($str, $start, $length, $encoding)
+			? mb_substr($str, $start, (is_null($length) ? strlen($str) : $length), $encoding)
```
jschreuder: This also seems wrong: it solves one very specific invalid input type but ignores all others. Seems random and plain wrong.

EDIT: sorry missed the explanation just now, it's a bit hacky though in its current implementation. Wouldn't something like `$lenght = is_null($length) ? mb_strlen($str) - $start : $length` be better? And it should go on a line of its own with a comment explaining it, otherwise it won't make sense to anyone looking at it.

classes/arr.php
crynobone: Always open for alternative. Having fail Unit Testing in core make it harder to test in app/package. 

I'm not sure why reference was prefered in the first place but abviously `$this->assertEquals($expected, \Arr::prepend($arr, array('one' => 1));` return NULL instead the referenced array.

jschreuder: When a unit test fails you must always check whether its a bug in the code or whether the unit test is wrong, in this case the unit test doesn't match the code which is clearly designed to work by reference. Thus the test should be fixed to match the code it is supposed to test and use prepend outside the assertEquals function and test the array afterwards.

crynobone:
Updated based on comments.

tests/arr.php
```diff
@@ -555,7 +555,8 @@ class Tests_Arr extends TestCase {
 			'two' => 2,
 			'three' => 3,
 		);
-		$this->assertEquals($expected, Arr::prepend($arr, 'one', 1));
+		$output = Arr::prepend($arr, 'one', 1);
```
jschreuder: won't work, $output will get null.

tests/arr.php
```diff
@@ -574,7 +575,8 @@ class Tests_Arr extends TestCase {
 			'two' => 2,
 			'three' => 3,
 		);
-		$this->assertEquals($expected, Arr::prepend($arr, array('one' => 1)));
+		$output = Arr::prepend($arr, array('one' => 1));
```
jschreuder: won't work, $output will get null.

classes/str.php
```diff
@@ -88,7 +88,7 @@ class Str {
 		$encoding or $encoding = \Fuel::$encoding;
 
 		return function_exists('mb_substr')
-			? mb_substr($str, $start, $length, $encoding)
+			? mb_substr($str, $start, (is_null($length) ? mb_strlen($str, $encoding) : $length), $encoding)
```
jschreuder, line 91: This should still go on its own line with an explanatory comment and is wrong because it requests more length than is ever available. Subtract the $start because that way the code makes more sense, even if it doesn't matter for the output.

PR #533, opened 2011-10-01
Several test related changes.

sheeep:
`$> oil test` does not fail anymore.
- One test has been removed, as it has been performed already, another two has been altered so they fit the Arr::prepend/element interface.
- A bug has been fixed in Str::sub().


jschreuder:
Could you check this against the current state? I merged a pull-request today that was similar and fixed at least partially the same thing.

PR #536, opened 2011-10-02
Add closure fix to fuel.php. If closure stored in config file arguments...

leeovery:
I dont particularly like this and Im sure it can be done much better but its the best way I could come up with / find.

Essentially, if a closure is assigned as a value in a config file, after passing through the 'value' method in core/fuel.php all parameters were trashed and lost.

This meant closures couldn't be stored in config files if you needed to define arguments to the function. Without arguments it worked fine.

With this pull request the closure is recreated using the Reflection class and the SplFileObject class. All the arguments are re-attached to the closure.

Sorry if Ive done this pull request wrong too. Still getting used to git!

Thanks,
Lee (kudoslee)

### DIFF
--- a/classes/fuel.php
+++ b/classes/fuel.php
@@ -662,15 +662,60 @@ class Fuel {
 
 	/**
 	 * Takes a value and checks if it is a Closure or not, if it is it
-	 * will return the result of the closure, if not, it will simply return the
-	 * value.
+	 * will return the result of the closure with any function arguments
+	 * preserved. If not, it will simply return the value.
 	 *
 	 * @param   mixed  $var  The value to get
 	 * @return  mixed
 	 */
 	public static function value($var)
 	{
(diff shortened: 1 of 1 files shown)


jschreuder:
This was by design: closures in config are meant to be executed in order to get the actual value, it wasn't meant to be able to pass along closures. Your "fix" isn't that, it actually adds functionality but more importantly it adds ambiguity: closures without args are always executed but add a single var and it will be a weirdly regenerated closure. That's not a good thing.

We made the call to only allow closures in config that are meant to dynamicly assign a value, not closures as config values.

leeovery:
Ah I see. Now you've explained I can see what's going on!

Fair enough. I'll work around it. Shame though as that functionality would be a positive one but that's only my opinion. 

Perhaps this could be documented somewhere at some point as I know a few people on IRC were confused about this. 

Thanks for replying. 

Lee

PR #539, opened 2011-10-03
Fixed #538:

### LINKED ISSUE #538: Response Status Header doesn't work with \Controller_Template
Examine the following code

```
public function action_404()
{
    $this->response->status = 404; //doesn't seem to be sending 404 response
    $this->template->title = '404 - Page Not Found';
    $this->template->content = View::forge('app/error')->set('title', '404 - Page Not Found');
}
```

The problem with this code related with the changes to 1.1/develop _The $this->response Response object is now deprecated. Your action methods should return either a string, View object, ViewModel object or a Response object_. To make it right, we need to provide a second parameter for \Response::forge() in https://github.com/fuel/core/blob/1.1/develop/classes/controller/template.php#L54 to `$this->response->status` (Controller and Controller_Rest work this way). 

```
// After controller method has run output the template
public function after($response)
{
    // If the response is a Response object, we don't want to create a new one
    if ($this->auto_render === true and ! $response instanceof \Response)
    {
        $response = \Response::forge($this->template, $this->response->status);
    }

    return parent::after($response);
}
```

It should also be the root cause of this issue https://github.com/fuel/core/issues/515

But that changeset also cause another issue since the way it check the condition is wrong. normal web server should return `HTTP 1/1 404 File not found` and not `Status: 404 File not found` and vise versa.

crynobone:
- Allow Response Header Status to be set from Controller_Rest since using auto_render, the final response object is only set from after() method.
- Normal web server should return HTTP 1/1 404 File not found and not Status: 404 File not found and vise versa.

### DIFF
--- a/classes/controller/template.php
+++ b/classes/controller/template.php
@@ -51,7 +51,14 @@ abstract class Controller_Template extends \Controller {
 		// If the response is a Response object, we don't want to create a new one
 		if ($this->auto_render === true and ! $response instanceof \Response)
 		{
-			$response = \Response::forge($this->template);
+			$status = null; // it should set to 200 when given null
+
+			if ($this->response instanceof \Response)
+			{
+				$status = $this->response->status;
+			}
+
+			$response = \Response::forge($this->template, $status);
 		}
 
 		return parent::after($response);
--- a/classes/controller/template.php
+++ b/classes/controller/template.php
@@ -51,14 +51,8 @@ abstract class Controller_Template extends \Controller {
 		// If the response is a Response object, we don't want to create a new one
 		if ($this->auto_render === true and ! $response instanceof \Response)
 		{
-			$status = null; // it should set to 200 when given null
-
-			if ($this->response instanceof \Response)
-			{
-				$status = $this->response->status;
-			}
-
-			$response = \Response::forge($this->template, $status);
+			$response = $this->response; 
+			$response->body = $this->template;
 		}
 
 		return parent::after($response);
--- a/classes/response.php
+++ b/classes/response.php
@@ -202,7 +202,7 @@ class Response {
 		if ( ! headers_sent())
 		{
 			// Send the protocol/status line first, FCGI servers need different status header
-			if (!empty($_SERVER['FCGI_SERVER_VERSION']))
+			if ( ! empty($_SERVER['FCGI_SERVER_VERSION']))
 			{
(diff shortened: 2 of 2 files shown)


classes/response.php
```diff
@@ -202,7 +202,7 @@ class Response {
 		if ( ! headers_sent())
 		{
 			// Send the protocol/status line first, FCGI servers need different status header
-			if (empty($_SERVER['FCGI_SERVER_VERSION']))
+			if (!empty($_SERVER['FCGI_SERVER_VERSION']))
```
jschreuder: There needs to be a space both before and after a `!` according to the coding guidelines.

classes/controller/template.php
```diff
@@ -51,7 +51,14 @@ abstract class Controller_Template extends \Controller {
 		// If the response is a Response object, we don't want to create a new one
 		if ($this->auto_render === true and ! $response instanceof \Response)
 		{
-			$response = \Response::forge($this->template);
+			$status = null; // it should set to 200 when given null
+
+			if ($this->response instanceof \Response)
```
jschreuder: is `$this->response` ever something else?

classes/controller/template.php
```diff
@@ -51,7 +51,14 @@ abstract class Controller_Template extends \Controller {
...
+			{
+				$status = $this->response->status;
+			}
+
+			$response = \Response::forge($this->template, $status);
```
jschreuder: I think it's  better to reuse the existing `$this->response` here: `$response = $this->response; $response->body = $this->template;`

classes/controller/template.php
crynobone: Practically it's shouldn't happen. A quick way of doing it is by assuming $this->response is always an instanceof \Response but it could be overwritten somewhere between __construct to after. Should be ok if $this->response->status is a valid response status but \Response class never really check if status is a valid response code or even a number.

classes/controller/template.php
crynobone: The current code is based on https://github.com/fuel/core/commit/13c6101a2a24eaff1a6d44d6dd5b3a8ab7b62c03 made by @dhorrigan. I would leave that to you guys to decide :)

jschreuder: My suggestion is consistent with how he solved `Controller_Rest` (if no Response was returned, $this->response is assumed to be a Response object).

crynobone: Refixed as suggested

PR #548, opened 2011-10-05
Added replace functionality to the Arr class

axelitus:
Added some replace functionality that sets item values only if they already exist in the array.
This functions are helpful when some default array is used but also an options array is given and one does not want to clutter the resulting array with irrelevant keys.

This was previously Pull Request [509](https://github.com/fuel/core/pull/509)

### DIFF
--- a/classes/arr.php
+++ b/classes/arr.php
@@ -481,6 +481,73 @@ class Arr {
 
 		return (array_sum($array) / $count);
 	}
+	
+	/**
+	 * Sets the value of the given item only if the item exists in the given array.
+	 * The key must be dot-notated. 
+	 *
+	 * @param   array  the array to fetch from
+	 * @param   mixed  the key for the item to be replaced in the array
+	 * @param   mixed  the value to be set if the key exists
+	 * @return  bool   true: element replaced | false: element not in array
+	 */
+	public static function replace_item(array &$array, $key, $value)
+	{
+		if(empty($array))
+		{
+			return false;
+		}
+		
+		if (!is_string($key) && !is_numeric($key))
+		{
+			throw new \InvalidArgumentException('Second parameter must be a string or numeric.');
+		}
+		
+		$keys = explode('.', $key);
+
+		while (count($keys) > 1)
+		{
+			$key = array_shift($keys);
+			
+			if(!isset($array[$key])){
+				return false;
+			}
+			
+			$array =& $array[$key];
+		}
+		
+		$key = array_shift($keys);
+		if(!isset($array[$key])){
+			return false;
+		}
+		
+		$array[$key] = $value;
+		return true;
+	}
+	
+	/**
+	 * Sets the values of the given items only if the items exists in the given array.
+	 * The keys must be dot-notated. 
+	 *
+	 * @param   array  the array to fetch from
+	 * @param   array  associative array with the dot-notated keys as indexes and the values to be replaced for
+	 * @return  int    the count of the replaced elements
+	 */
+	public static function replace_items(array &$array, array $keys_values)
+	{
+		if(empty($array))
+		{
+			return 0;
+		}
+		
+		$return = 0;
+		foreach($keys_values as $key=>$value)
+		{
+			$return += (int)(static::replace_item($array, $key, $value));
+		}
+		
+		return $return;
+	}
 
 	/**
 	 * Alias for replace_key for backwards compatibility.
--- a/classes/arr.php
+++ b/classes/arr.php
@@ -643,6 +643,59 @@ class Arr {
 
 		return $array;
 	}
+	
+	/**
+	 * Merge 2 arrays recursively, differs from \Arr::merge() in:
+	 * - The values are only merged if they exist in the left-most array
+	 * 
+	 * The value of the right-most array has precedence.
+	 * Note that if a branch-end-item (a.k.a. an item that is not an array) is to be
+	 * replaced, it will only be replaced by a non-array value.
+	 *
+	 * @param   array  multiple variables all of which must be arrays
+	 * @return  array
+	 * @throws  \InvalidArgumentException
+	 */
+	public static function merge_replace()
+	{
+		$array  = func_get_arg(0);
+		$arrays = array_slice(func_get_args(), 1);
+		
+		if ( ! is_array($array))
+		{
+			throw new \InvalidArgumentException('Arr::merge_replace() - all arguments must be arrays.');
+		}
+		
+		foreach ($arrays as $arr)
+		{
+			if ( ! is_array($arr))
+			{
+				throw new \InvalidArgumentException('Arr::merge_replace() - all arguments must be arrays.');
+			}
+			
+			foreach ($arr as $key => $value)
+			{
+				if(isset($array[$key]))
+				{
+					if(is_array($array[$key]))
+					{
+						if(is_array($value))
+						{
+							$array[$key] = static::merge_replace($array[$key], $value);
+						}
+					}
+					else
+					{
+						if(!is_array($value)){
+							$array[$key] = $value;
+						}
+					}
+				}
+			}
+		}
+		
+		return $array;
+	}
 
 }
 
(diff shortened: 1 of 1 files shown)


jschreuder:
While usefull I think these are a bit too specific to really belong in the core, don't have a strong opinion on this though so I'll leave it to @dhorrigan, @WanWizard, @philsturgeon & @FrenkyNet to comment/decide further.

philsturgeon:
Got some examples or use cases?

WanWizard:
This pull request contains both bugfixes and new functionality.

Could you split the two, now the discussion about the new methods is blocking the merge of the bugfixes?

jschreuder:
@WanWizard check the diff, it contains 2 new methods.

axelitus:
As an example I am using this method in my History package. As I said this is useful for config arrays, but it's not limited to. The intended use is for when one needs to change an array's value but only if the key exists previously in the first array.

History Package example: [History Package L113](https://github.com/axelitus/fuel-pkg-history/blob/1.0/master/classes/history.php#L113)

jschreuder:
You just described the method again and then showed an example where you use `Arr::merge()` as alternative when `Arr::merge_replace()` is not present, this didn't explain why it's necessary it mostly showed that it's not really necessary.

axelitus:
I use the `Arr::merge()` method as a fallback because as of now the core does not have that functionality, so the package would be useless to anybody that downloads it. The `Arr::merge_replace()` method actually mimics what `Arr:merge()` does, so it can be safely interchanged. The difference is that when `Arr::merge_replace()` is called no new keys are added if they do not exist in the left-most array.

The next thing I'd like to say is that this function is more of a convenience rather than a 'must-have necessity' as I see it.

Now I'll go with an example (it's gonna be a bit long). It's a fictious example and it may sound really silly also, but hopefully would explain the convenience such a function can bring:

Imagine one has an external source (such as a WebService) that would return data like the following (for reasons unknown) and one could not change this (I'll just make a value assignment here):

```
$original_arr = array(
    0 => array(
        'name' => 'Name 0',
        'email' => 'email_0 [at] domain [dot] com',
        'alternate_email' => 'alternate_email_0 [at] domain [dot] com',
        'age' => 'Age: 23'
    ),
    1 => array(
        'name' => 'Name 1',
        'email' => 'email_1 [at] domain [dot] com',
        'alternate_email' => 'alternate_email_1 [at] domain [dot] com'
    ),
    2 => array(
        'name' => 'Name 2',
        'email' => 'email_2 [at] domain [dot] com',
        'age' => 'Age 23'
    ),
    3 => array(
        'name' => 'Name 3',
        'email' => 'email_3 [at] domain [dot] com'
    )
);
```

In our App we would want to know how many emails the user has and his/her age (as a number). We will be using that info as follows (perhaps there's a better way but we'll stick with this):

For age:

```
echo "The user is " . \Arr::get($row, 'age', '?') . " years old.";
```

For email quantity:

```
function email_quantity($row)
{
    $count = 0;
    isset($row['email']) and $count++;
    isset($row['alternate_email']) and $count++;

    return $count;
}

echo "The user has " . email_quantity($row) . " emails.";
```

## Example 1

First we need to 'filter' what we received. I'll do three different examples here. The first one is by using just the `Arr::set()` method:

```
foreach ($original_arr as &$row)
{
    \Arr::set($row, 'email', clean_email(\Arr::get($row, 'email', '')));
    \Arr::set($row, 'alternate_email', clean_email(\Arr::get($row, 'alternate_email', '')));
    \Arr::set($row, 'age', (int)\Arr::get($row, 'age'));
}
```

Modified array:

```
array(
    0 => array(
        'name' => 'Name 0',
        'email' => 'email_0@domain.com',
        'alternate_email' => 'alternate_email_0@domain.com',
        'age' => 23
    ),
    1 => array(
        'name' => 'Name 1',
        'email' => 'email_1@domain.com',
        'alternate_email' => 'alternate_email_1@domain.com',
        'age' => 0 
    ),
    2 => array(
        'name' => 'Name 2',
        'email' => 'email_2@domain.com',
        'alternate_email' => '',
        'age' => 23
    ),
    3 => array(
        'name' => 'Name 3',
        'email' => 'email_3@domain.com',
        'alternate_email' => '',
        'age' => 0
    )
)
```

The output for this example would be:

```
Name 0
The user is 23 years old.
The user has 2 emails.

Name 1
The user is 0 years old.
The user has 2 emails.

Name 2
The user is 23 years old.
The user has 2 emails.

Name 3
The user is 0 years old.
The user has 2 emails.
```

Which obviously is not what we intended to display. With this method we will need to introduce some more conditions...

## Example 2

We could easily fix this if we would introduce a new condition:

```
foreach ($original_arr as &$row)
{
    \Arr::key_exists($row, 'email') and \Arr::set($row, 'email', clean_email(\Arr::get($row, 'email', '')));
    \Arr::key_exists($row, 'alternate_email') and \Arr::set($row, 'alternate_email', clean_email(\Arr::get($row, 'alternate_email', '')));
    \Arr::key_exists($row, age') and \Arr::set($row, 'age', (int)\Arr::get($row, 'age'));
}
```

Modified array:

```
array(
    0 => array(
        'name' => 'Name 0',
        'email' => 'email_0@domain.com',
        'alternate_email' => 'alternate_email_0@domain.com',
        'age' => 23
    ),
    1 => array(
        'name' => 'Name 1',
        'email' => 'email_1@domain.com',
        'alternate_email' => 'alternate_email_1@domain.com'
    ),
    2 => array(
        'name' => 'Name 2',
        'email' => 'email_2@domain.com',
        'age' => 23
    ),
    3 => array(
        'name' => 'Name 3',
        'email' => 'email_3@domain.com'
    )
)
```

The output for this example would be:

```
Name 0
The user is 23 years old.
The user has 2 emails.

Name 1
The user is ? years old.
The user has 2 emails.

Name 2
The user is 23 years old.
The user has 1 emails.

Name 3
The user is ? years old.
The user has 1 emails.
```

Which is what we wanted to output

## Example 3

The last example is using the `Arr::replace_item()` method, which handles the `Arr::key_exists()` condition automatically:

```
foreach ($original_arr as &$row)
{
    \Arr::replace_item($row, 'email', clean_email(\Arr::get($row, 'email', '')));
    \Arr::replace_item($row, 'alternate_email', clean_email(\Arr::get($row, 'alternate_email', '')));
    \Arr::replace_item($row, 'age', (int)\Arr::get($row, 'age'));
}
```

Which will give us an identical array as Example 2, thus displaying the info correctly.

## Conclusion

As I said earlier, this functions are merely convenient, they are by no means necessary as we could manage to get around the lack of them. I don't know if they do belong to the core though as that is for you to decide. I do find them useful as one has lesser code to write later on.

WanWizard:
@jschreuder: that's what I wrote. We can't merge this pull request until a descision has been made about the new methods, which means we also can't pull the bugfixes included in this request.

frankdejonge:
@wanwizard: aren't the bugfixes just commits for the new features, I don't see any other bugs being fixed in the diff.

axelitus:
Yes, the bugfixes are just commits about the two new functions (previously three). No other methods were touched in the Arr class, I did only merge new pushes from the fuel/core/1.1/develop branch + my own two functions.

WanWizard:
Ok. Bear with me while I'm trying to find my glasses... ;)

jschreuder:
Ok, I get its use now but I still don't think it belongs in the core.

Also I don't get why you're using Arr::set(), in the case below you can also eliminate Arr::get() which will make the whole statement a lot faster as its all native PHP. Something that can matter when dealing with a lot of rows.

```
array_key_exists('email', $row) and $row['email'] = clean_email($row['email']);
```

Which I would argue is actually easier to understand than:

```
\Arr::replace_item($row, 'email', clean_email(\Arr::get($row, 'email', '')));
```

@WanWizard, @FrenkyNet reopen if you disagree.

axelitus:
Indeed in this case you're right, there are other more optimized ways to achieve this... I wasn't aiming at optimization either as it was just an example of usage...

Perhaps you're right and the replace_item() method is altogether not necessary, but what about the merge_replace() method for merging arrays only by replacing? This is to have a 'clean' array at the end, with no unnecessary items/keys in it.

PR #549, opened 2011-10-05
Fixed an issue where the get_browser() function would return an object instead

axelitus:
Fixed an issue where the get_browser() function would return an object instead of an array. The array keys are now lower-cased as the returned array may return an array with lower-cased keys.

Quoting the PHP.net docs:
"The information is returned in an object or an array which will contain various data elements representing, for instance, the browser's major and minor version numbers and ID string; TRUE/FALSE values for features such as frames, JavaScript, and cookies; and so forth."

This issue popped up in Windows7 64bit using xampp1.7.3

### DIFF
--- a/classes/agent.php
+++ b/classes/agent.php
@@ -33,62 +33,62 @@ class Agent {
 	 * @var  array  information about the current browser
 	 */
 	protected static $properties = array(
-		'Browser'             => 'unknown',
-		'Version'             => 0,
-		'MajorVer'            => 0,
-		'MinorVer'            => 0,
-		'Platform'            => 'unknown',
-		'Alpha'               => false,
-		'Beta'                => false,
-		'Win16'               => false,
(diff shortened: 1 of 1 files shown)


WanWizard:
_cough_

Sorry, placed an irrelevant remark. I'll have a look at the code.

axelitus:
Haha :) that happens... a lot... at least to me hehe...

I tried to touch as little as possible... if using browscap, the array keys get lower-cased using the array_change_key_case() so this should be no problem at all (in fact, every time, whether is through get_browser() or with browscap() the keys get lowered)

It worked for me, but it should be tested using browscap too...

WanWizard:
The only thing I was a bit concerned about was that apps could break that use the data directly. But as they should use the property() method, and that converts the key to lowercase, that should not be a problem. I'll merge.

PR #551, opened 2011-10-07
As with $msg, $open and $close of validation_error get_message() should override the config defaults

phil-lavin:


jschreuder:
You're right but the code isn't yet up to the coding standards: operators like ?: need spaces on both sides, thus:

```
$open   = empty($open) ? \Config::get('validation.open_single_error', $open) : $open;
```

Otherwise it's unreadable.

phil-lavin:
Is that ok?

PR #557, opened 2011-10-10
Added create_index and drop_index functions to DBUtil class

zoe-edwards:
I couldn’t find this anywhere and it’s often something you add as the project grows, so it’s probably better off being completely separate to `create_table()`.

Added two functions:

```
create_index($table, $field, [ $type ] )
```

`$type` might be `unique` or `fulltext`.

```
drop_index($table, $field)
```

Pretty self explanatory.

It gives the index name the same name as the field.

### DIFF
--- a/classes/dbutil.php
+++ b/classes/dbutil.php
@@ -305,6 +305,55 @@ class DBUtil {
 		}
 		return false;
 	}
+	
+	/**
+	 * Creates an index on that table.
+	 * 
+	 * @access	public
+	 * @static
+	 * @param	string	$table
+	 * @param	string	$index_name
+	 * @param	string	$index (should be 'unique' or 'fulltext')
+	 * @return	bool
+	 * @author	Thomas Edwards
+	 */
+	public static function create_index($table, $index_name, $index = '')
+	{
+		$sql = 'CREATE ';
+		
+		$accepted_index = array('unique', 'fulltext');
+		
+		if ($index !== '')
+		{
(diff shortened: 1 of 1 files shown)


crynobone:
Would be nice if we can combine two or more fields to create an INDEX

zoe-edwards:
Yeah good point, I’ll edit it if nobody beats me to it.

philsturgeon:
I've noticed this never made it into the documentation. Could you add some examples and an explanation, then send a new pull request please?

zoe-edwards:
Sorry I was going to expand on it to allow the combination of two or more fields. I’ll write the Docs now and do that later!

philsturgeon:
Nono this pull request is in, done, history. Just make a new pull request on the docs repo :)

zoe-edwards:
Ah nice, cheers!

WanWizard:
Those two are documented as well: 411e925715c0249d207ae2e2522501a3fd898658

philsturgeon:
I'll use grep over TextMate next time, they are indeed there:

```
grep -R "create_index" .
./classes/database/dbutil.html: <h4 class="method" id="method_create_index">create_index($table, $index_columns, $index_name, $index = '')</h4>
./classes/database/dbutil.html: <p>The <strong>create_index</strong> method allows you to create secondary indexes on a table.</p>
./classes/database/dbutil.html: <pre class="php"><code>\DBUtil::create_index('table', 'name');
```

PR #562, opened 2011-10-10
Added jsonp support

rob-mccann:
As per pull request 317.

Added jsonp support to the rest controller.


philsturgeon:
Wahey, nice!

PR #563, opened 2011-10-10
Fix&Test to Format::_from_xml() and Better suport to customize links in Pagination class

kriansa:
Well, I'm still newbie in git, so forgive-me if I'm doing something wrong.

I made some fixes on my last pull request, but I screw everything up with a bunch of unwanted commits. Now I'm doing as requested by @philsturgeon.

The first is a real fix to issue #432. The actual function Format::_from_xml() convert the XML and its children as SimpleXMLElement objects, instead arrays (as expected).

In my test, I use to_php() to show you clearly that it's really unexpected.

And the commit b41b62f90a5a9fceaf80fe719557d21461d577b2 adds a better support to customize the unclickable links << Previous and Next>> when they're inactive(when you're in first or last page), so we can set a custom CSS class on them.

Thanks.

### DIFF
--- a/classes/pagination.php
+++ b/classes/pagination.php
@@ -51,6 +51,8 @@ class Pagination {
 		'next_mark'      => ' &raquo;',
 		'active_start'   => '<span class="active"> ',
 		'active_end'     => ' </span>',
+		'inactive_start'   => '<span class="inactive"> ',
+		'inactive_end'     => ' </span>',
 	);
 
 	/**
@@ -226,7 +228,7 @@ class Pagination {
 
 		if (static::$current_page == static::$total_pages)
 		{
-			return $value.static::$template['next_mark'];
+			return static::$template['inactive_start'].$value.static::$template['next_mark'].static::$template['inactive_end'];
 		}
 		else
 		{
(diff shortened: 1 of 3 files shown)


classes/pagination.php
```diff
@@ -226,7 +228,7 @@ class Pagination {
 
 		if (static::$current_page == static::$total_pages)
 		{
-			return $value.static::$template['next_mark'];
```
philsturgeon, line 229: Can you do something about these changes showing up?

philsturgeon, line 229: Oh sorry I hadn't noticed that this was an intended change. Generally we ask people to make sure only one "change" happens per pull request.

kriansa, line 229: I see. Is there any way to accept just one diff of a pull request or should I make another one, or two?

philsturgeon, line 229: Thanks for making the XML file in a new pull request, this has been merged to develop. Please make the validation pull request fresh too.

kriansa:
Okay... Now I get it!

https://github.com/fuel/core/pull/578

and

https://github.com/fuel/core/pull/577

Thanks!

PR #564, opened 2011-10-10
Environment based config

tarnfeld:
I have been playing about with this idea for a while and with these new config updates it makes doing this a whole lot easier!

Though, I am not sure how the `Config_File::load()` method should work now.. what is the idea behind the `$override` option? Also, simply using `array_merge` won't allow for full extention, eg: in production I have loads of database config but in development I just have a change of db host and password... I looked into using `array_merge_recursive` but from some tests I did if you have two conflicting keys it wont write over, it will append so you end up with both values - not like `array_merge` behaves!

### DIFF
--- a/classes/config/file.php
+++ b/classes/config/file.php
@@ -94,7 +94,7 @@ abstract class Config_File implements Config_Interface
 			return array_reverse($paths);
 		}
 
-		return array();
+		throw new \ConfigException(sprintf('File "%s" does not exist.', $this->file));
 	}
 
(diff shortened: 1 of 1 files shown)


dhrrgn:
Basically, the `$overwrite` option tells it whether to do just an array_merge (overwrite the entire config group) or an array_merge_recursive (but we use our custom Arr::merge() method to overwrite keys).

tarnfeld:
Here ye go.

PR #569, opened 2011-10-11
Changes to the Validation class to allow Validation of input arrays

saf33r:
I altered the Validation run method to allow validation of input arrays. Basically it checks if the form input is an array, if so it runs through each of the array values and validates according to set rules.

### DIFF
--- a/classes/validation.php
+++ b/classes/validation.php
@@ -347,25 +347,53 @@ class Validation
 		{
 			static::set_active_field($field);
 
-			$value = $this->input($field->name);
-			if (($allow_partial === true and $value === null)
-				or (is_array($allow_partial) and ! in_array($field->name, $allow_partial)))
+			$field_name = str_replace('[]', '', $field->name);
+			$is_array = is_array($this->input($field_name));
+			$values = array();
+			$validated_array = array();
+
+			if($is_array)
 			{
-				continue;
-			}
-			try
-			{
-				foreach ($field->rules as $rule)
+				foreach($this->input($field_name) as $input_field)
 				{
-					$callback  = $rule[0];
-					$params    = $rule[1];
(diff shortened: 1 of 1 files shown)


mjsilva:
With the current version this was the only way I found to use Validate input arrays without touching/extending the core. http://manueljoaosilva.com/2011/10/fuelphp-input-arrays-validation/

saf33r although I appreciate your effort in trying to make this work I don't think this is the best implementation since a user can manipulate the input to make it look like an array and change the flow of the validation.

jschreuder:
@saf33r the point made by @mjsilva is indeed why this solution won't be accepted. Validation should determine from the fieldset whether it's dealing with an array or not, if you detect that from the input you create a potential security hole because you make it possible for an attacker to change how the input is validated.

It also looks like your code can only deal with one specific type of array field input and won't support deeper arrays or named keys, those would also run the risk of being overwritten with the way you handle the input later.

My main problem with all of this remains that it's wrong from a conceptual point of view: an array in the input should still be considered one field instead of many. As a complex field there's no build-in validation for it and you'd have to write a specific rule for it.

I understand where it can be usefull, and I'll think about support but any support for this should never overrule the above principle.

saf33r:
@mjsilva and @jschreuder, thanks for the feedback and sorry for the delay in replying, just been too busy.

Yes this does only support basic arrays and it actually needs recursion to allow for nested arrays.

I'm not quite sure whether it opens a security hole as if you give it an array when a value is expected, then it would treat that field as an array. Which would in turn throw an exception, at least that's how I read the code to work and what I saw from experience.

Would it be possible to discuss further how you envision a array validation solution to work? That way I or someone reading can tackle it as some stage?

PR #571, opened 2011-10-12
Added in missing previous / next link template markup

shodanuk:


crynobone:
This pull request is similar to #525

shodanuk:
Indeed it is, I'll close this one. I've left a comment on your commit which you might like to consider :)

PR #573, opened 2011-10-12
Migrations - added Module and Package Support

dlpetrie:
Added module and package support for migrations.

Modules accessible by adding --modules option for all modules or --modules=name1,name2 for specific modules.

Packages work the same as modules.

adding --modules or --packages disabled the default migration unless you add in the --default option.

Works for all commands, though specifying a version only allows you to update 1 item per command.

Edit:

removed edit - methods now default to the apps default migration folder

### DIFF
--- a/classes/migrate.php
+++ b/classes/migrate.php
@@ -28,6 +28,12 @@ class Migrate
 
 	protected static $table = 'migration';
 
+	protected static $table_definition = array(
+		'name' => array('type' => 'varchar', 'constraint' => 50),
+		'type' => array('type' => 'varchar', 'constraint' => 25),
+		'version' => array('type' => 'int', 'constraint' => 11, 'null' => false, 'default' => 0),
+	);
+
 	public static function _init()
 	{
 		logger(Fuel::L_DEBUG, 'Migrate class initialized');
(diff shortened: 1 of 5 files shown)


classes/migrate.php
```diff
@@ -37,21 +37,22 @@ class Migrate
 		static::$table = \Config::get('migrations.table', static::$table);
 
 		\DBUtil::create_table(static::$table, array(
```
dhrrgn: Should detect if table already exists, and if it does, then modify it with the new columns and add the app entry to it.

classes/migrate.php
```diff
@@ -61,9 +60,9 @@ class Migrate
 	 * @access	public
 	 * @return	mixed	true if already latest, false if failed, int if upgraded
 	 */
-	public static function latest()
+	public static function latest($name, $type)
```
philsturgeon: This stuff needs to be optional. It should work identical to how it works now if you mention nothing, then upgrade a specific module if you pass it the arguments.

tasks/migrate.php
```diff
@@ -52,97 +174,149 @@ class Migrate {
...
 		}
 
 		// Just go to the latest
 		else
 		{
-			if (($result = \Migrate::latest()) === false)
+			if (($result = \Migrate::latest($name, $type)) === false)
 			{
-				throw new \Oil\Exception("Could not migrate to latest version, still using {$current_version}.");
+				\Cli::write('Already on latest migration for '.$type.':'.$name.'.');
```
philsturgeon, line 203: Why are we moving away from throwing Oil Exceptions? Anything that runs as part of Oil should be throwing exceptions like this. The task cannot run without oil, so it's not a dependency issue. Please can you change this back.

classes/migrate.php
philsturgeon: Agreed, put it in and mark it as deprecated or some shiz, so we remove this extra check in v1.2.

tasks/migrate.php
dlpetrie, line 203: The reason I change those out to \Cli::write was to allow for the multiple migrations at once.  If they stayed as Exceptions, once an Exception was thrown, it would completely stop the process rather than going to the next item to migrate.

If someone does: php oil refine migrate --modules and the 2nd of 5 modules is already on the latest version, it will stop the execution at the 2nd module and the last 3 won't migrate.

dlpetrie:
Forgot to mention in commit cce568f that i made params optional for public methods - it now defaults to the main apps migration.

Should be up to date with all request changes except for the oil exceptions vs cli::write

PR #578, opened 2011-10-14
Adds a inactive support to inactive links in Pagination class

kriansa:
Adds a better support to customize the unclickable links << Previous and Next>> when they're inactive(when you're in first or last page), so we can set a custom CSS class on them.

### DIFF
--- a/classes/pagination.php
+++ b/classes/pagination.php
@@ -52,6 +52,8 @@ class Pagination
 		'next_mark'      => ' &raquo;',
 		'active_start'   => '<span class="active"> ',
 		'active_end'     => ' </span>',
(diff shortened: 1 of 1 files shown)


crynobone:
Similar to pull request #525

WanWizard:
Pagination class functionality will be reviewed after v1.1.

kriansa:
May I ask why? There will be another class with its funcionality?

WanWizard:
We're not happy with the current class, it's bloated, and contains to much stuff that belongs in the app, not in a core class.
As we're in feature freeze for v1.1. now, nothing can be done about it untill that is released.

I created a new issue for this so it stays on the radar for the next milestone.

PR #579, opened 2011-10-15
Fixed default behaviour when using ::find_one_by should return single obj

crynobone:
Fixed default behaviour when using ::find_one_by should return single object, not an array


dhrrgn:
@FrenkyNet You said you wouldn't break it...you lied!

frankdejonge:
@dhorrigan Ha! And you broke scaffolding! ... seriously, you did. Model_Crud model is namespaced. And Orm model is Model_ prefixed. Views don't know the difference and just use the namespaced. I think it's better to go with Model_ prefix for now, as all other things do that atm.

PR #581, opened 2011-10-16
Fixed Lang::get() - now returns original line, if nothing found.

ushis:
Now Lang::get() is more usable. If you write something like

``` php
<h1><?php echo __('All Users'); ?></h1>
```

and "All Users" is not translated, it will return 'All Users' instead of nothing...


ushis:
Oh no, that wasnt pretty awesome. Sorry. These changes are better: https://github.com/fuel/core/pull/582

PR #582, opened 2011-10-16
Fixed __() - now returns original line, if nothing found.

ushis:
Now `__()` is more usable. If you write something like

``` php
<h1><?php echo __('All Users'); ?></h1>
```

and "All Users" is not translated, it will return 'All Users' instead of nothing...

### DIFF
--- a/base.php
+++ b/base.php
@@ -170,7 +170,7 @@ if ( ! function_exists('__'))
 	{
 		$default === false and $default = $string;
 		
-		return \Lang::get($string, $params, $string);
+		return \Lang::get($string, $params, $default);
 	}
 }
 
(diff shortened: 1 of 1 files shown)


WanWizard:
I don't agree with removing the default.

If you have a habit of using language string codes instead of a gettext() style of language strings, you can no longer pass a default

```
$lang  = __('allstars', array(), 'All Stars');
```

Perhaps it's better to use

```
$default === false and $default = $string;
return \Lang::get($string, $params, $default);
```

so you can switch to using the input string when you pass 'false' as default value.

ushis:
Hmm, i think that's a good compromise.

ushis:
Nope, these changes are silly. It only makes sense to put this one line into Lang::get(). Further i think that calling

``` php
<?php
echo __('Hey Ho', array(), false);
```

is not that great. So people, who'd like to use the gettext() style, can write their own method for their app. I'll close this.

Issue #594, opened 2011-10-21
Module specific 404 routing not working


darklow:
In case of 404 error framework tries to load app default 404 route, not the one specified in modules/MODULE/config/routes.php
It looks like in the HttpNotFoundException of index.php file Config class doesn't know yet about Modules routes config.

WanWizard:
That depends. 

Like all other module items, a 404 route defined at the module level is only available within the request for that module.

darklow:
Didn't quite understand your comment, but let me explain by real example
I have module named **admin**

Configs are:
app/config/routes.php: `'_404_'   => 'web/404'`
app/modules/admin/config/routes.php: `'_404_'   => 'admin/404'`

Now when i go to address:
/admin/non_existing_dir/
Framework show's nothing but blank page.

When i tried to dump `Config::$items` in public/index.php i can see that Config doesn't contain any routes config from admin module, but only from app config.

`$route = Config::get('routes._404_')` Variable $route in this case is: "web/404" and not "admin/404" as i think it should be

Therefore following script tries to load web/404 controller in the admin module, which i think is wrong.

``` php
catch (HttpNotFoundException $e)
{

    if ($route = Config::get('routes._404_'))
    {
        $response = Request::forge($route)->execute()->response();
    }
    else
    {
        throw $e;
    }
}
```

WanWizard:
I was refering to the moment the module routes are loaded. But your 404 is triggered from a module, so that's not an issue.

I don't think the catch should use Config information there, as that doesn't contain the current route list.

Can you dump \Router::$routes at that point, and see if that does correctly lists "admin/404" as the `_404_` route? (can't test at the moment).

darklow:
Yes, i can confirm, `\Router::$routes` contains correct `_404_` route as specified in module config.

WanWizard:
Ok. Does this work?

```
catch (HttpNotFoundException $e)
{
    $route = array_key_exist('_404_', Router::$routes) ? Router::$routes['_404_'] : Config::get('routes._404_');
    if ($route)
    {
        $response = Request::forge($route)->execute()->response();
    }
    else
    {
        throw $e;
    }
}
```

darklow:
Yes, now it routes correctly. 
Just with a small fix: `Router:$routes['_404_']` is object(Fuel\Core\Route), so it works when i used `->translation` key:

``` php
catch (HttpNotFoundException $e)
{ 
[...]

PR #596, opened 2011-10-21
Override default 404 route.

OscaTutenchamon:
If `Controller` has it's own `action_404()` and method that doesn't exists inside controller will be called `Controller->action_404()` method will be executed instead default `_404_` route.
I don't know Fuel very well so no idea if that's the best way of doing that.

### DIFF
--- a/classes/request.php
+++ b/classes/request.php
@@ -458,8 +458,15 @@ class Request
 			}
 			else
 			{
-				static::reset_request();
-				throw new \HttpNotFoundException();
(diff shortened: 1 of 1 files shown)


dhrrgn:
The _404_ route is deprecated in 1.1, as such, there is no point in adding functionality that is based on it.

WanWizard:
@dhorrigan, does that mean `_404_` is going to be removed from the routes.php? And if so, how are you going to deal with module specific 404 routes? See also issue #594.

PR #602, opened 2011-10-23
Added a referer() function alongside referrer() to reflect the spelling of

jesseobrien:
Added a referer() function alongside referrer() to reflect the spelling of PHP HTTP_REFERER. The spelling difference will likely get people annoyed when they're used to "referer" (it did me).

### DIFF
--- a/classes/input.php
+++ b/classes/input.php
@@ -211,11 +211,21 @@ class Input
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


jschreuder:
Even though it's a common mispelling we don't do aliases, there should always be but 1 function for something - anything else ends in a mess.

PR #611, opened 2011-10-25
Fixed "Fatal error: Class 'FuelException' not found"

Savageman:
It may be deprecated, but the current with the default config doesn't work. Autoloader is not registered yet so we can't use \FuelException from the global namespace here, it won't be found.

### DIFF
--- a/classes/fuel.php
+++ b/classes/fuel.php
@@ -20,7 +20,7 @@ class FuelException extends \Exception {}
 /**
  * @deprecated  Keep until v1.2
  */
-class Fuel_Exception extends \FuelException {}
+class Fuel_Exception extends FuelException {}
 
 /**
  * The core of the framework.


Savageman:
Side Note. I was catching the old \Fuel_Exception somewhere in my app and it's not BC with the actual code from core.

Because \Fuel_Exception extends \FuelException and \FuelException is now thrown so it's not caught. So I think it should be the other way to be BC until it's removed.

jschreuder:
It is registered with the autoloader: https://github.com/fuel/core/blob/1.1/develop/bootstrap.php#L106 so I'm not sure what the problem is there?

The other point is valid though, it'd be better for BC to have FuelException extend Fuel_Exception. @dhorrigan agree?

Savageman:
Yes, you're right first point is invalid, it's not a problem with the default setup.

I extended Fuel and with the previous setup, it was impossible to extend the Fuel class itself without loading it by hand as it's the case with the Autoloader class today.

Now the Autoloader class is loaded first, I can extend Fuel like any other core class, first problem solved for me.

jschreuder:
@dhorrigan this was your change so I'll leave the fix up to you.

WanWizard:
@dhorrigan: fix or close?

PR #612, opened 2011-10-25
Feature: Allow sub-directories in config files

Savageman:
This disappeared since the Config driver refactoring. Not sure why.

### DIFF
--- a/classes/config.php
+++ b/classes/config.php
@@ -53,7 +53,6 @@ class Config
 		{
 			$info = pathinfo($file);
 			$type = isset($info['extension']) ? $info['extension'] : 'php';
-			$file = $info['filename'];
 			$class = '\\Config_'.ucfirst($type);
 
 			if (class_exists($class))
--- a/classes/config/file.php
+++ b/classes/config/file.php
@@ -58,7 +58,7 @@ abstract class Config_File implements Config_Interface
 	 */
 	public function group()
 	{
-		return pathinfo($this->file, PATHINFO_FILENAME);
+		return $this->file;
 	}
 
 	/**


dhrrgn:
How were you trying to load it?  It works.  I have tested it multiple times and it works every time.

Savageman:
Sorry, this "Comment & Close" just next to another "Comment" button is the works design ever. :s

This way:

``` php
\Config::load('mymodule::myconfig', true);
$config = \Config::get('mymodule::myconfig', array()); // returns empty array
```

Without the patch, the namespace (module name) gets stripped, leaving only the file name part.

Also when loading the "form" config file from a specific module, I don't want every "form" config files from every loaded modules and from the core to be returned! ^^

huglester:
I think you should do smht like this:
\Config::load('mymodule::myconfig', 'configname');

and later: echo \Config::get('configname.key');

Savageman:
Yeah, that could work I guess. But this seems counter intuitive as the special value "true" for the 2nd argument becomes pointless.

Would be nice if it works with modules too!

huglester:
well... I'm not sure how it should work then? 
how the 'alias' should be named?
'modulename_configname' ? 
because if we only take the 'configname', it could lead to problems with overwriting the  default configs etc...

I'll lead that to devs :)

Savageman:
I dug some more into this and have more information to share !

`\Config::load('module::configname', true);` works ok.

What isn't working is `\Config::load('module::folder/configname');`

It's because the filename is computed using `pathinfo()` and
`pathinfo('module::configname', PATHINFO_FILENAME) == 'module::configname');` but
`pathinfo('module::folder/configname', PATHINFO_FILENAME) == 'configname');`

I think that makes more sense now.

So the title of this pull request should be: allow sub-directories for config files. :-)

jschreuder:
This has gone beyond the pull-request, create a new feature request for this - won't make it for 1.1 though as we're in feature freeze.

Savageman:
From my point of view it's a regression since the config refactoring, because it worked fine in 1.0 and I was using it. But I'll leave it up to you. ;)

@jschreuder, you better want a new ticket for this? Isn't it better to let this pull request opened and close it when it can be merged after 1.1 ?

jschreuder:
If it's a bugfix it can still go into 1.1 but if it's new functionality it can't. As I understand it this is a problem with using `true` for the group when loading a config file from a subdirectory in a specific module, this pull-request doesn't solve that but it changes the behavior. We were using the filename as the groupname (no matter the subdir) and this makes it use the full path always.

Savageman:
Not quite. Without this patch, it's impossible to load a config file from any sub-directory (it's not module-related), because the directory is stripped with the pathinfo() thing.

You're right it changes the behavior (which we need to be careful about) by using the full path instead of just the file name, but that aligns with the old behavior before the refactoring.

jschreuder:
Ahh, you changed the topic without noting it in the original post, I though @dhorrigan replied to the subdirs that he tested it and it works.

Anyway the pull-request is still bad and what I said still stand: any bugfix will be accepted and a functional change will be denied.

Savageman:
Yes, sorry for changing the title and not telling it in the original post. :'(

I cannot decide whether it's a bug-fix or a new feature.
If it should be possible to load config files from subdirectories (like it is for the views), then it's a bug-fix. If it wasn't planned at first then it's a new feature. I don't know who has the intel on this, but I don't. ;)

jschreuder:
It was possible before and I don't think it wasthe intention to disable it, though disabling it wouldn't be illogical in the long term with environment configs being put in subdirs.

For now I will accept any pull-request that enables loading it from dirs like before without changing any other behavior.

dhrrgn:
Fixed here https://github.com/fuel/core/commit/8a288b57c7b3a337d487724f4a34e220573d4fab

dhrrgn:
On another note:  The group name will by default still only be the actual config file name if only `true` is sent as the group.  So `Config::load('foo/bar', true);` will result in a group name of `bar`.  If you wish to have a different group name you can specify it by `Config::load('foo/bar', 'foo_bar')` or whatever.

jschreuder:
@dhorrigan doesn't this cause a difference between files in subdirs from modules and files without subdirs from modules? The subdir-ed files get the `module::` prefixes stripped, but the non-subdir-ed won't get the module prefix stripped.

Savageman:
Cross-posting, I'm doing it again. ^^

Awesome, thanks for resolving this!

Still have the change of behavior since 1.0 with module config files loaded using the :: syntax, but I can live with it.
Not sure which behavior is the best since when loading a config file without :: the module from the current request will be searched and the config file found anyway.
I just feel it's a bit weird to merge all the config with the same name (from every module) in the same group by default, but it's up to me. ^^

Savageman:
@jschreuder: I think you're right, but it's even worse than you say I think.

i didn't actually run the code, but if I calculate the group in my mind, I get this:
`Config::load('name', true); // group = name, ok`
`Config::load('subdir/name', true); // group = name, not ok, should be subdir/name`
`Config::load('module::name', true); // group = moddule::name, not ok? should be name?`
`Config::load('module::subdir/name', true); // group = name, not ok, should be subdir/name?`

The `Config_File->group()` method should return `Inflector::demodulize($this->file);` ?

dhrrgn:
Should be fixed...guess we should have just merged the pull request :P

Savageman:
And now we're back to exactly my original pull request... Which I'm really happy with since it aligns with the old behavior from 1.0. :)

But now all config files with the same name are not merge in the same group when loading them with the module prefix, which @jschreuder didn't look happy about...

jschreuder:
@dhorrigan I care too little to change it, but I thought the idea was that the groupname was always the filename. Which is why I denied the pull-request.

dhrrgn:
Ya, I dont particularly care at the moment, nor do I have the time.

jschreuder:
At which time it's better most of the time to leave something open then to fix it up without thinking it through...

jschreuder:
Ok, I checked the 1.0 branch - as @Savageman said it used to work like this so while I think it'd be better to make the filename the auto-group, this was the right way for BC.

Savageman:
Now that the code is exactly what I suggested, it's just too bad I don't appear as contributor for this in the commit. :'(

But I'm really happy this is in the trunk now. :)

PR #614, opened 2011-10-26
Added support to BMP in Image class

kriansa:
I added GD functions to handle BMP images, as it's still a image format very used. Imagick and Imagemagick drivers has native support.

I don't see a point excluding BMP support just because GD doesn't support it.

### DIFF
--- a/classes/image/gd.php
+++ b/classes/image/gd.php
@@ -18,7 +18,7 @@ class Image_Gd extends \Image_Driver
 {
 
 	private $image_data = null;
-	protected $accepted_extensions = array('png', 'gif', 'jpg', 'jpeg');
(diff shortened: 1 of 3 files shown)


dhrrgn:
Did you get this code from somewhere, or is it 100% original?  If you got it from somewhere...where?

kriansa:
I made some adjustments for the function imagebmp from phpThumb and imagefrombmp is from http://www.phpro.org/examples/Convert-BMP-to-JPG.html

I never meant to hide the original credits, but I thought it wouldn't be necessary.

dhrrgn:
We can't use this then, as phpThumb is licensed under the GPL.  GPL code cannot be included in software released under the MIT license.

PR #615, opened 2011-10-26
Several bugfixes / documented but missing features in Image class

kriansa:
I made a few fixes in Image and Drivers classes, nothing new, just what should be.

Added docblocks to Image::reload()
Changed visibility of image drivers vars to protected, allowing extensions/plugins
Added real support to JPEG quality to Imagemagick and Imagick drivers
Removed a duplicated checking on exec command in ImageMagick::exec()
Removed unused var in ImageMagick::exec()
Improved create_color code (all drivers uses the same function, why this was written 3 times?)
Little fix on Imagick::add_background: it was suposed to add a bg only if bgcolor is set (like other drivers)
Fixed a mispell in Imagic::create_color docblock

### DIFF
--- a/classes/image.php
+++ b/classes/image.php
@@ -274,6 +274,11 @@ class Image
 		return static::instance()->sizes($filename);
 	}
 
+	/**
+	 * Reloads the image.
+	 *
+	 * @return  Image_Driver
+	 */
 	public static function reload()
 	{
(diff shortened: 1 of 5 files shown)


WanWizard:
@dhorrigan: can this be merged?

PR #627, opened 2011-10-30
Updated curl to allow for headers as array

rob-mccann:
Hi all,

Curl wasn't setting the headers as I expected it to so I made these changes, it'll need reviewing by someone who knows more about the class than me :)

Rob :)


rob-mccann:
looks like this has been fixed since I last pulled :)

PR #629, opened 2011-10-30
Respect the common strategy to signify a SSL reverse proxying

krtek4:
I'm using FuelPHP on an host behind a proxy. Users connect with HTTPS on the proxy which then relays the request with HTTP. A classic SSL reverse proxy situation.

The problem is that `Input::protocol()` returns `http` where, in my sense, it should return `https`. The following patch resolve the problem by using two of the more common strategies used in these kind of proxying. An issues opened against Drupal explains this a little better : http://drupal.org/node/313145

(My original problem was that the actions of the forms where pointing to the http version of the site, and I traced down this "issue" to this.)

### DIFF
--- a/classes/input.php
+++ b/classes/input.php
@@ -192,7 +192,9 @@ class Input
 	public static function protocol()
 	{
 		if ((static::server('HTTPS') !== null and static::server('HTTPS') != 'off')
-			or (static::server('HTTPS') === null and static::server('SERVER_PORT') == 443))
+			or (static::server('HTTPS') === null and static::server('SERVER_PORT') == 443)
+			or (static::server('HTTP_X_FORWARDED_PROTO') === 'https')
+			or (static::server('HTTP_FRONT_END_HTTPS') === 'on'))
 		{
 			return 'https';
 		}


WanWizard:
I prefer not to have this in the core code.

It is something an end-user can inject, and can lead to security issues. You can easily work around this in your application by using

```
if(isset($_SERVER['HTTP_X_FORWARDED_PROTO'])) $_SERVER['HTTPS'] = 'on';
```

krtek4:
I totally understand to security argument and I must admit I overlooked that.

Maybe provide some kind of "force SSL" config value ? I'm sure I'm not the only one having this issue, and it will probably be useful for some people.

Taking this a little further, a configuration option on the Form class to always use HTTPS instead of HTTP ? Some people have their "main" pages on HTTP and only redirect the forms actions to HTTPS.

PS: Sorry if a pull request is not the place to discuss such things.

WanWizard:
Please add an issue for this, so we can have a look. It's not for v1.1 though, as we are a feature freeze at the moment.

PR #630, opened 2011-10-31
Better locale initialization

Savageman:
More reliable locale initialization + adding an error message when it's not installed.

### DIFF
--- a/classes/fuel.php
+++ b/classes/fuel.php
@@ -210,7 +210,11 @@ class Fuel
 		\Router::add(\Config::get('routes'));
 
 		// Set  locale
-		static::$locale and setlocale(LC_ALL, static::$locale);
+		if (!empty(static::$locale))
+		{
+			setlocale(LC_ALL, static::$locale, static::$locale.'.'.strtolower(str_replace('-', '', static::$encoding)))
+			or logger(\Fuel::L_WARNING, 'The configured locale '.static::$locale.' is not installed on your system.', __METHOD__);
+		}
 
 		static::$initialized = true;
 


jschreuder:
What is exactly the meaning of this: `static::$locale.'.'.strtolower(str_replace('-', '', static::$encoding))`? I know localizations as `aa_AA` and just `aa` (where `aa` is a 2 letter country abbreviation) not sure what type of localization this is supposed to provide?

Savageman:
On my dev environment (which is on Ubuntu), the locale aa_AA doesn't exist, I need to use aa_AA.utf8

A locale is defined like this: language_COUNTRY.encoding@variant
For example, in France you can have fr_FR.iso885815@EURO
But Country, encoding & variant are optional.

I think it's better for initialization to test both the locale alone (aa_AA) and the locale with encoding (aa_AA.utf8)

jschreuder:
The problem I have with this is that it's kinda random, the most universal way is the aa_AA and many other systems may have their own way of setting this. The question is whether we should add fallbacks for all of those, I think not in this case unless this is a very widely used alternative. Some things are so incredibly non-universal that by adding support for 1 specific system we open the floodgates for people requesting support for all their specific preferences. In cases like these it may be best to require people to set their localization system specificly, because the spagetti code resulting from specific support is much worse.

Savageman:
Yeah Ubuntu may not be the most universal system for checking this.

I just had a look on Cent OS and Debian, and the aa_AA locales are there indeed (aa_AA.utf8 are there too).

I also checked all the core files and it seems Fuel::$locale is never used anywhere (for now?). So could another solution would be to put the encoding in the app config file, like this: `'locale' => 'fr_FR.utf8'`. If yes, a word on this in the documentation would be nice.

Otherwise, I think the logger is still a good thing to warn the user?

jschreuder:
Good point about the docs, I'll add a note about that right now to the configuration page.

I'm thinking about the logger, maybe we should even make it throw an exception when setting the locale fails - that might have far reaching consequences for the app.

@dhorrigan what do you think?

jschreuder:
nvm, the logger is better - as long as we have locale set to en_US by default it shouldn't throw an exception when setting it fails.

Savageman:
Thanks, I'm happy with this. :-)

PR #635, opened 2011-11-01
Fix display of fatal error in modules

kbanman:
When a fatal error is encountered in a module, the error page's reference to the `Fuel` class is using a relative namespace, and thus failing.

### DIFF
--- a/views/errors/php_fatal_error.php
+++ b/views/errors/php_fatal_error.php
@@ -52,7 +52,7 @@
 				$debug_lines = \Debug::file_lines($trace['file'], $trace['line']);
 		?>
 			<li>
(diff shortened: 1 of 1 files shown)


jschreuder:
Ok, this is just weird: any file loaded will be in global unless explicitly namespaced itself - no matter the namespace it was taken from. This shouldn't fix anything...

crynobone:
Did happen to face this issue once, from package but was not able to reproduce it consistently.

jschreuder:
@kbanman: I just tested this, it doesn't make a difference. The fatal error from module is shown correctly with or without the backslash. I won't undo it as it doesn't make any difference, but you should realize you may be masking a more serious error.

WanWizard:
Just to be clear, a non-namespaced class request is relative to the current namespace, and not global. 

What is relevant is the location of the code, so in this case the namespace of the error handler, not of the module class that triggered the error.

Issue #636, opened 2011-11-01
Fatal error when adding a namespaced class name to Fieldset::add_model


kbanman:
I'm getting this error:
`ErrorException [ Compile Error ]: Cannot redeclare class Clients\Model_Client_Contact`

From this code:

```
\Fieldset::forge('client_contact')->add_model('\\Clients\\Model_Client_Contact');
```

Which is inside a module's controller (`Clients`).
Also note, the following line does not work either:

```
\Fieldset::forge('client_contact')->add_model('\\Clients\\Model_Client');
```

But instead of an error message I just get a blank page, and nothing in the logs either.
Also, unlike the previous line, replacing the `\\Clients\\Model_Contact` string with an instance of `Model_Client` works for this example. The first one gives the error `InvalidArgumentException [ Error ]: Input for add_callable is not a valid object or class.`.

jschreuder:
Could you try it without the prefixed backslash, classes referenced by string will always be taken from global context and don't need the backslash prefix. If that's the problem the solution for this is as simple as adding `ltrim($class, '/')` to `add_model()`

Btw: I'm using namespaced models and adding them without trouble.

kbanman:
Yes, removing the prefixed backslash fixes the problem. I've submitted a [pull request](https://github.com/fuel/core/pull/638).

jschreuder:
Closing this and discussion can continue on #638

PR #638, opened 2011-11-01
Fix namespace prefixing in Fieldset::add_model()

kbanman:
See [Issue 636](https://github.com/fuel/core/issues/636)


jschreuder:
The fix isn't complete as it just applies to `add_model()` and does nothing about `add_callable()`.

kbanman:
I messed up my fork's branches. Will re-submit.

PR #640, opened 2011-11-02
BMP support in Image class

kriansa:
A few days ago I created this pull #614 but it was closed due a GPL licensed code incompatible with MIT license.

I'm re-pulling with some improvements since my previous code and a Public Domain function (not GPL), with all the source links as requested by @dhorrigan.

Like I said before, I can't see a point leaving behind BMP support just because GD doesn't support it, while other driver's can handle it just fine. Of course is not a web format, neither should be, but many people still use it.

Thanks.

### DIFF
--- a/classes/image/gd.php
+++ b/classes/image/gd.php
@@ -18,7 +18,7 @@ class Image_Gd extends \Image_Driver
 {
 
 	private $image_data = null;
-	protected $accepted_extensions = array('png', 'gif', 'jpg', 'jpeg');
+	protected $accepted_extensions = array('png', 'gif', 'jpg', 'jpeg', 'bmp');
 	protected $gdresizefunc = "imagecopyresampled";
 
 	public function load($filename, $return_data = false)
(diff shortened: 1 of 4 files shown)


huglester:
I'm the one of those - who needs this. thank you :)

dhrrgn:
Please update the code to comply with our coding standards and I will merge the request.

kriansa:
Done.

huglester:
@dhorrigan, @WanWizard or whoever... can this be merged in? It's really needed support. thanks

WanWizard:
No. "This pull request cannot be automatically merged.".

It's need an update before it can be merged.

huglester:
Hello again :)
@kriansa maybe you can help a little?

I wanted to addopt this pull request, so it gets into fuel 1.1/develop, but after I tried to do some test - I failed with an error:
http://d.pr/C7uC

I have created test suite to test things up with different drivers, located here: http://scrp.at/baS

Installation is easy, you put this controller to controller/ dir, create a directory: 'bundle_images' in public/ folder, and put there at least one photo.

Enter the controller via site.com/bundle and after a while you'llsee the results with different drivers.

Thanks

kriansa:
I tested this code on my core fork and it's working fine (except imagick that doesn't work well on windows) but the bmp rendering by GD is ok.

Take a look at my code gd.php -> https://github.com/kriansa/fuel-core/blob/me/develop/classes/image/gd.php

huglester:
strange. I tried to replace my gd.php file with yours with same results. Maybe it works 'better' on windows.. not sure here.
Can you share your BMP image example somewhere? 

Just in case my image if somehow different or so (those 16,24,32 bits etc)

huglester:
Just tested, 24 bits BMP works ok, 16 and 32 bit fails under Linux

kriansa:
This is odd. It definitely doesn't work better on windows, but it's true that it wasn't tested on linux. It was supose to work, since it doesn't uses any os-specific functions...

edit
8 bits images doesn't seems to work either, even under windows.

edit
Oh crap, my commit also added another unwanted file... I will make another pull request later.

huglester:
I've read somewhere that http://www.programmierer-forum.de/function-imagecreatefrombmp-laeuft-mit-allen-bitraten-t143137.htm this one works with 16 bits too, etc. but so far tried to adopt it with fuel with no luck

kriansa:
@huglester
I don't intend to update this to support all these formats. I just made this pull because I know that bmp support is disabled by fuel just because GD doesn't support it, so instead creating a simple error message like "sorry, GD driver doesn't support BMP images", I added this simple function found somewhere in web. Imagemagick is a fantastic lib and should be used for production when handling with BMP images, cuz this function is very slow. It took 7 secs to renderize a 800x600 BMP file, while Imagemagick took just 1.8sec. However, feel free to modify as you wish.

huglester:
hello again @kriansa ,
seems you are right. BMP support in GD is a bit poor, and other drivers should be used. 

I have done more testings, but could not get this working. I think at least we can throw a notice, that 'Only 24bit bmp images' are supported, untill somebody find the sollution for them to work too

WanWizard:
@kriansa @huglester What are you going to to with this pull request. Close it, as it doesn't seem to be the solution?

huglester:
I have tried diffent ways to integrate 8/16/32 bit support but still with no luck. I have loooked into phpThumb() they seem to handle this, but we just can't take the code from them :(

kriansa:
I think we should use your idea and throw an exception when handling with bmp images with gd. What do you think? At least we aren't blocking bmp support in Imagemagick.

huglester:
kriasna, I think it would be the best way for now. At least we'll be told what is happening and why GD is not accepting BMPs...

WanWizard:
Ok. Can someone code up a new pull request for this? And some docs explaining this?

frankdejonge:
We've talked about this, but supporting BMP formats is way out of date, and shouldn't be encouraged any longer. Closing pull-requests.

Issue #601, opened 2011-10-23
Pagination class


WanWizard:
The status of the current Pagination class needs to be reviewed.

Does a bit of application-related code have a place in the core? If so, can we make it less bloated, and template driven so we can get rid of the HTML now generated by the class?

jschreuder:
I'm in full agreement on this, less features - anything complex can go into extensions or the template.

dhrrgn:
What if Pagination ::link() (I think that is the method). Took a callback which generates the links?  We call the callback for every link that needs to be generated.

WanWizard:
Could be an option, but that requires quite a bit of logic that you need to store somewhere in your app, maybe we can do something clever with placeholders, and just pass a string?

And all HTML stuff has to go into a template, it doesn't belong in the class.

Savageman:
Don't know if it will help, but I had to rewrite the pagination class to fit my needs.
My main concern was to not use a URI "segment" to store the page number.

The callback function is exactly what I did. The only parameter it takes is the current page number to generate the link for.

This way I could implement my app logic as I wanted:
- Store the page number somewhere in the URL in a way that fits our URL rewriting mechanism (not as an URI segment, nor as a GET parameter)
- Remove the page number from the URL for the first page

sergiigrebeniuk:
I think paginator is a widget. We can place it into core/modules

WanWizard:
Added to the 2.0 roadmap

PR #642, opened 2011-11-02
Pagination - use template for prev and next links

ushis:
You have defined `static::$template['next_start']`, but did not use it. And it should be possible to have no anchor value and use next/prev mark only.


jschreuder:
This won't go into v1.1 (feature freeze has been in place since RC1) and for the next version the entire Pagination class will be reviewed and possibly rewritten: #601

ushis:
allright

PR #644, opened 2011-11-03
Response: Made a few changes to forge() and __construct() so that headers can also 

kapv89:
Made a few changes to forge() and __construct() so that headers can also be supplied while forging a response. This allows to do stuff like: 

return \Response::forge(json_encode($data), 200, array('Content-type'=>'application/json'));r

### DIFF
--- a/classes/response.php
+++ b/classes/response.php
@@ -68,9 +68,9 @@ class Response
 		509 => 'Bandwidth Limit Exceeded'
 	);
 
-	public static function forge($body = null, $status = 200)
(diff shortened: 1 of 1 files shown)


jschreuder:
I'm not opposed to this, but parenthises are required as per our [coding standards](http://fuelphp.com/docs/general/coding_standards.html).

Also this is a new feature and we're officially in feature freeze, thus this will probably be bumped to the next version.

kapv89:
Hey .. i'm closing this pull request as i made another one with those paranthesis (new to github .. didn't click on the right places :P)

PR #645, opened 2011-11-03
Repsonse: allow headers to be set when creating the request

kapv89:
Changed forge() and __construct() to allow headers to be set while forging a request

### DIFF
--- a/classes/response.php
+++ b/classes/response.php
@@ -68,9 +68,9 @@ class Response
 		509 => 'Bandwidth Limit Exceeded'
 	);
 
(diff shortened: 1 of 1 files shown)


jschreuder:
And still not fully up to spec, I probably missed this the last time: you must have spaces between `$k=>$v`.

Btw if you add a commit fixing this to the same branch it should just be added to the pull-request (or at least it used to be).

kapv89:
closed this one as another one is open which does the same thing and matches fuel's coding standards

PR #650, opened 2011-11-06
Added pagination module (probably for v1.1.1?)

sergiigrebeniuk:
Signed-off-by: delmot grebenjuk.s@gmail.com


dhrrgn:
Why do we need this?  We already have the Pagination class.

sergiigrebeniuk:
There is the issue #601 connected with this class

dhrrgn:
This is not the answer to the problem...this is a ton of application logic in the core, which we want to avoid.

PR #652, opened 2011-11-08
HTTP_ACCEPT is always checked to detect REST format by default

crynobone:
Config::get('rest.ignore_http_accept') by default is not set, which would return null, therefore the condition would return true and additional checking is require. We should add a default value to avoid overhead with the additional checking unless specifically requested

### DIFF
--- a/classes/controller/rest.php
+++ b/classes/controller/rest.php
@@ -162,7 +162,7 @@ abstract class Controller_Rest extends \Controller
 		}
 
 		// Otherwise, check the HTTP_ACCEPT (if it exists and we are allowed)
-		if (\Config::get('rest.ignore_http_accept', false) !== false and \Input::server('HTTP_ACCEPT'))
+		if (\Config::get('rest.ignore_http_accept') === false and \Input::server('HTTP_ACCEPT'))
 		{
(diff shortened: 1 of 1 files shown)


philsturgeon:
It should be set by default, which is probably a better problem to fix. 

Also, that condition would be better suited as === true instead of defaulting to true and !== false

crynobone:
It was changed at f332919 from === false to !== false which cause the it to always run format detection via HTTP_ACCEPT

philsturgeon:
Right now you are saying "if config is null, make it false, then check for false". Just check if its true or not instead of faffing.

crynobone:
Updated based on suggestion, so the original code is right all along...

philsturgeon:
Is it? Surely it should be `=== true`,  which is what I've been suggesting.

crynobone:
The config description suggest 'ignore_http_accept', from my understanding of it

ignore_http_accept === true (shouldn't check HTTP_ACCEPT)
ignore_http_accept === false (HTTP_ACCEPT shouldn't be ignore, check it)

I might be wrong though.

philsturgeon:
This has been fixed in 1.1/develop, thank you for pointing out the issue!

PR #656, opened 2011-11-09
Ensure that the variables are strings and prevents injections on the Mongo query

ipalaus:
According to that document http://www.idontplaydarts.com/2010/07/mongodb-is-vulnerable-to-sql-injection-in-php-at-least/

And this references in php.net: http://php.net/manual/en/mongo.security.php

This is the way to prevent attacks of user input.


ipalaus:
@dhorrigan I need to Sign-off this to being accepted?

philsturgeon:
Thank you, this was never even considered!

PR #658, opened 2011-11-10
Add chainable set_param() to Request

kbanman:
Supports setting named parameters in a chain, for example

```
$request = Request::forge('foo/bar')->set_param('foo', 'bar')->execute();
```

Maybe there should be support for passing in an array of parameters this way?

### DIFF
--- a/classes/controller.php
+++ b/classes/controller.php
@@ -50,6 +50,17 @@ abstract class Controller
 	{
 		return $response;
 	}
+	/**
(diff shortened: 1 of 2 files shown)


jschreuder:
Aren't params already a public property? That makes this a bit unnecessary. Also I don't actually think it's good practice, the request params are passed through the URI - there's better places to collect non-URI params which you could always merge with the request params in some other way.

PR #667, opened 2011-11-14
Change radio/checkbox value comparison to ===

andrewwheal:
Changed to === when comparing values of radio/checkbox fields to determine whether they need to be checked

### DIFF
--- a/classes/fieldset/field.php
+++ b/classes/fieldset/field.php
@@ -390,7 +390,7 @@ class Fieldset_Field
 						$attributes['value'] = $value;
 						$attributes['label'] = $label;
 
-						if (is_array($this->value) ? in_array($value, $this->value) : $value == $this->value)
+						if (is_array($this->value) ? in_array($value, $this->value) : $value === $this->value)
 						{
 							$attributes['checked'] = 'checked';
 						}


jschreuder:
This wasn't done by accident, the problem may arise in different ways - if the original values are set as ints `'5'` won't match `5` with this. The problem with the current solution is that all false-values match each other and true matches any non-false value. Thus this doesn't solve anything, it only changes the location of the problem.

andrewwheal:
Ah, didn't think of that... however I want Yes/No options with values of 1 and 0 respectively and there is currently no way of doing this without having the No option checked.

The other problem of having `'5'` and `5` would at least be solvable (always make sure you are working with the same type), but is probably more common a problem.

jschreuder:
I am open to other solutions, as it is a problem. I just don't accept a solution that creates a new problem somewhere else. Also remember that `in_array()` uses `==` comparisson without the 3rd param being set to `true` (which is an inconsistency in your pull-request).

andrewwheal:
I also didn't look through the rest of the code for anywhere else that a `==` might need to be changed to a `===` for consistency...

I will put some thought into how this might be possible as I would prefer my 'No' not to be auto selected when the page loads.

PR #673, opened 2011-11-17
Added field messages to fieldsets

andrewwheal:
Added the ability to set messages on fieldset fields which can then be displayed next to the field using the configurable templates when the fieldset is built

### DIFF
--- a/classes/fieldset/field.php
+++ b/classes/fieldset/field.php
@@ -477,7 +477,7 @@ class Fieldset_Field
 		$error_msg = ($form->get_config('inline_errors') && $this->error()) ? str_replace('{error_msg}', $this->error(), $error_template) : '';
 		$error_class = $this->error() ? $form->get_config('error_class') : '';
 		$message_template = $form->get_config('message_template', "");
-		$field_msg = ($form->get_config('error_replaces_message') && $this->error()) ? '' : str_replace('{field_msg}', $this->message, $message_template);
+		$field_msg = ($this->message && !($form->get_config('error_replaces_message') && $this->error())) ? str_replace('{field_msg}', $this->message, $message_template) : '';
 
 		if (is_array($build_field))
 		{
--- a/classes/fieldset/field.php
+++ b/classes/fieldset/field.php
@@ -50,6 +50,11 @@ class Fieldset_Field
 	protected $value;
 
 	/**
+	 * @var  string  Text that will be displayed next to the field
+	 */
+	protected $message = '';
+
+	/**
 	 * @var  array  Rules for validation
 	 */
 	protected $rules = array();
@@ -183,6 +188,19 @@ class Fieldset_Field
 
 		$this->value = $value;
 		$this->set_attribute('value', $value);
+
+		return $this;
+	}
+
+	/**
+	 * Change the field's message
+	 *
+	 * @param   string
+	 * @return  Fieldset_Field  this, to allow chaining
+	 */
+	public function set_message($message)
+	{
+		$this->message = (string)$message;
 
 		return $this;
 	}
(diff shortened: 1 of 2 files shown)


WanWizard:
@jschreuder: any comments?

darklow:
This could be very useful. I use field help messages almost for all the fields all the time. I think this is a "musthave" feature and hope it will be approved.

andrewwheal:
(As a reminder to myself) I am going to look at adding a value to the $_properties array on an Orm\Model which can then be read into this so that $fieldset->build() can be used.

jschreuder:
While I like the thoughts behind it I'm not happy with the implementation.
You have a switch for 'error_replaces_message', but it's actually 2 different types of messages: when it's on it's an error message overwrite and should just replace the error message. When it's off it's probably an additional instruction for the field or some other description. Those are 2 very different things and shouldn't be handled as the same thing. Best solution here would be to remove the switch and split it up in a 'error_overwrite' and a 'description' (or something similar for additional text with the field).

Also it's not fully up to coding standards yet, you need a few more spaces - for example line 203 after `(string)` and a `!` also gets a space after it on line 480. Line 480 has gotten way to complex for an inline condition anyway, something that complex should be a normal if for readability.

andrewwheal:
I can't think of any instances when you would want a separate description and error_override, and I purely added the error_replaces_message flag as an afterthought as I thought that people might like to be able to do this.

I will fix the coding standards when I do the Orm\Model $_properties.

jschreuder:
The point is that it's different usages, either it's just an error overwrite (in which case I'd ask you to rename it to that and not make it a separate value for the templates) or it adds a description field (which I think may be helpfull to many). But it shouldn't be both at the same time. I don't care if it's an afterthought or fully intentional: the bottom line is that it's in there.

andrewwheal:
Just working on this and re-read your comments, I think you may have misunderstood the 'error_replaces_message' flag.

It is for only showing the message when there is no error message to be displayed, not a message to display instead of an error message.

jschreuder:
Ah, I did indeed misunderstand it. But I don't really like this either, it should be an error message overwrite - that is much clearer. This overcomplicates it a bit and that's not a good thing, when people want more they'll have to add that themselves but I doubt it's ever really necessary.

If you rename $message/set_message() to $error_overwrite/set_error_overwrite() I'll accept it. In the build() method it should just overwrite the error, thus also no {field_msg} and related logic.

andrewwheal:
@darklow Thanks for spotting that, I will fix it when I get time.

@jschreuder:
I will remove the 'error_replaces_message' flag if you want, but I am not removing the whole message stuff... this was the point in the feature. I will however add error overwrite.

What is the best way to implement a 'field_message' item in the Orm/Model $_properties array that automatically gets pulled into the fieldset? I was looking at it last night and thought I would add it to where the fieldset is generated, but this seems to be in the Orm\Validation_Observer which seems an odd place for messages to be added?

PWesterdale:
I've thought the ability to only label fields has been a limitation for a while now. Whilst I agree that also an overwriting error message or specific error message per item is also useful, I'm of the opinion that if you have the ability (through messages like this pull request sets out) to tell the user what you want, you probably won't need a specific error message.

jschreuder:
@SicoAnimal
From the start I understood the whole point of this to allow for an error message overwrite, which is what I have been commenting on. If this "message" is meant for another goal (additional message shown only during errors?) you'll have to explain it better and give it a more descriptive name, 'message' is a far too generic name and doesn't cover its function in any way.
The flag has to go in any case, that I won't accept and I've gone into why. Everything should have a clear meaning and function, the flag makes it ambiguous and the name 'message' doesn't cover its meaning.

As to the Validation_Observer stuff, it is true that the form generation code currently resides there - that needs fixing and was the result of it growing beyond it's initial goals. You can pass any additional form attributes through a form key: [it's shown in the docs](http://docs.fuelphp.com/packages/orm/creating_models.html#/propperties)

@PWesterdale 
You can always extend the Fieldset_Field class and add your own stuff, which you can always pull-request later if you think it'll benefit others. We're very extendable just so you can work in your own demands easily.

andrewwheal:
The point of this is not to allow an overwrite for the error message but to allow the addition of a general message to a field which generally will always display regardless of whether the form had been submitted or there were any errors.

I then added a switch to allow the message to be hidden when an inline error message was being displayed incase people would find that useful (I have seen it on sites before).

Sorry for the confusion that this seems to have caused, I called it 'message' as it is a general message for a field (to, for example, tell the user about a character limit).

I know that form attributes can be set in the $_properties array, I was trying to find the best place to add code for pulling the value out and setting up the field messages. Shall I add it in the Validation_Observer with the other related code for it to be moved elsewhere when the whole lot is fixed, or wait until the other parts are fixed and add it then?

As an extra note, this did start life as an extension of Fieldset_Field which I then wrote into the core file for this pull request as some other developers I work with thought it would be useful.

jschreuder:
**The message field**
I think description is a better choice here because it contrasts better with "error message" which prevents confusion. With the switch removed and renamed to 'description' I'll accept it (I'm open to other name options as well).
Let's limit this pull-request to just that, it makes things easier.

**The "switch"**
This all has in part to do with a problem I've noticed earlier but for which we don't have a solution yet: template strings in Fuel. In the fieldset class it uses a limited mustache-like variable-replacement, Lang strings use Ruby-symbols and none allow for conditionality. We'll need to standarize this in Fuel and probably allow for full parser/templating libs to be used to allow conditionality in the templates.

**The error overwrite**
This was what caused the confusion with me because this pull-request reminded me of a recent feature request for error message overwrites regardless of the actual error thrown. This is a very interesting addition, possibly necessary but should be separate. If you're willing to look into it I'm grateful though ofc ;)

**The validation observer**
I'm not sure what you mean here, adding a key to the `'form' => array('message/description' => 'The actual message')` in the `$_properties` array should do shouldn't it?

jschreuder:
Closed in favor of #838

PR #674, opened 2011-11-18
Added support for before and after attributes in the Fieldset class, e.g...

0xD3ADBE3F:
$login_form->add('cookie', '', array('type' => 'checkbox', 'after' => 'Remember me for 2 weeks'));

So you could now have something like this for checkboxes/radio's:

Hello    |     Hi [ x ] Bye

### DIFF
--- a/classes/form.php
+++ b/classes/form.php
@@ -207,7 +207,17 @@ class Form
 			$attributes['id'] = \Config::get('form.auto_id_prefix', 'form_').$attributes['name'];
 		}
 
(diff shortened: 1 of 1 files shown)


WanWizard:
@jschreuder: any comments?

jschreuder:
Fieldset fields already support adding a specialized template for a field, which is a more flexible solution than this. Also it's weird/wrong to wrap the field itself inside a label element.

PR #675, opened 2011-11-18
Allows for DB::expressions with order_by

dlpetrie:
Allows for DB::expr() in order_by so you can return random results by doing

'order_by' => DB::expr('RAND()')

or

'order_by' => 'id' (which would default to ASC)

### DIFF
--- a/classes/model/crud.php
+++ b/classes/model/crud.php
@@ -183,9 +183,16 @@ class Model_Crud extends \Model implements \Iterator, \ArrayAccess {
 				$query->where($where);
 			}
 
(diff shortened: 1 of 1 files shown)


fabioluciano:
When you add a related model to a query using the order_by with DB::expr('RAND()'), the result doesnt work as expected. The query returns the first row.

jschreuder:
That seems to be about the ORM, this is about the Model_Crud class. The Orm has it's own Repo: [fuel/orm](https://github.com/fuel/orm)

PR #677, opened 2011-11-20
Fix for passing rest.valid_logins array by value rather than reference in rest controller error

slivero:
When using auth in rest controllers the following error is given:

ErrorException [ Runtime Notice ]: Only variables should be assigned by reference

COREPATH/classes/controller/rest.php @ line 252:

This is due to an array being passed by reference, this pull request fixes that.

### DIFF
--- a/classes/controller/rest.php
+++ b/classes/controller/rest.php
@@ -249,7 +249,7 @@ abstract class Controller_Rest extends \Controller
 			return false;
 		}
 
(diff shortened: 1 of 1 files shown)


philsturgeon:
Thanks for posting that. It was a trivial 0.000001 second time saver that breaks in PHP 5.3. Forgot to remove it :)

PR #689, opened 2011-11-24
Changed templating to make it a little more useful and consistent

peter-fritchley:
Changed template so that it includes the previous_start, previous_end, next_start and next_end template variables in the output. Also wrapped each of the page links to make it more consistent (and make for easier styling).

### DIFF
--- a/classes/pagination.php
+++ b/classes/pagination.php
@@ -42,7 +42,7 @@ class Pagination
 	public static $template = array(
 		'wrapper_start'  => '<div class="pagination"> ',
 		'wrapper_end'    => ' </div>',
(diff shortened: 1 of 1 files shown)


jschreuder:
@WanWizard I thought we weren't going to update the Pagination class anymore but replace it entirely for the next version?

WanWizard:
That is correct. 

There should be quite a few issues and pull requests open for this class, that are not accepted for that reason.

The reference above is an issue about this subject that I closed.

jschreuder:
I know and I would have closed it for that reason, but you referenced this pull-request from another.

WanWizard:
I did at that time because it was a duplicate, so I could close the other one.

PR #690, opened 2011-11-25
1.1/master

joshuarubin:
I am using fuel with postgresql. I know my changes probably break other db support and are incomplete. However, they should help give an understanding about areas of code that need to be adjusted for postgresql support.

These changes are my first attempt at porting my previous work in fuel 1.0. It is not well tested, but I wanted to give you something in case you find it useful before releasing fuel 1.1 final.

### DIFF
--- a/classes/controller/rest.php
+++ b/classes/controller/rest.php
@@ -249,7 +249,7 @@ abstract class Controller_Rest extends \Controller
 			return false;
 		}
 
-		$valid_logins = & \Config::get('rest.valid_logins');
+		$valid_logins = \Config::get('rest.valid_logins');
 
 		if (!array_key_exists($username, $valid_logins))
 		{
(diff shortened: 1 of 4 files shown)


frankdejonge:
Hey Joshua,

Just a few things. Pull request are made on develop, not on master. Separating pull requests helps is to get a clearer picture of the thing you want to get merged, making is more likely to get merged also. And a descriptive pull request message would also be handy. What did you do? What did you fix? And we expect people to test their code before requesting a pull. So I'm closing this one now.

Cheers,

frankdejonge:
and removing the per table charset will break mysql stuff. not the way to go.

crynobone:
Time to make DBUtil driver based?

frankdejonge:
it will be part of Query Builder 2.0

PR #691, opened 2011-11-25
Finder::search() with multiple = true not returning an array on absolute files

Savageman:
I don't know why github wants to push 6 commits into this.

I just updated my local repos and made 1 commit to submit it to you. In fact ther's only 1 file changed, which is good. :'( If anyone knows how to avoid this, it would help me.


WanWizard:
$multiple doesn't have any meaning with absolute files (which can only be one), so why pass a TRUE value?

Savageman:
It's for loading config files. There's an array_merge in the load() method. So it needs the array.

WanWizard:
Ok. I'll see if I can cherry-pick...

PR #695, opened 2011-11-28
Remove public folder from URL

dbpolito:
fuel.php changed to remove public folder from url based on config.php flag

### DIFF
--- a/classes/fuel.php
+++ b/classes/fuel.php
@@ -313,6 +313,12 @@ class Fuel
 
 			// Add a slash if it is missing
 			$base_url = rtrim($base_url, '/').'/';
+			
+			// Remove public folder
+			if (\Config::get('public_folder', false) === true)
+			{
+				$base_url = str_replace('public/', '', $base_url);
+			}
 		}
 		return $base_url;
 	}


dbpolito:
Pull Request related
https://github.com/fuel/fuel/pull/256

WanWizard:
What exactly are you trying to achieve here?

If you use rewriting, the public folder is not part of the URI. If you don't, the public folder must be part of the URI.

dbpolito:
Well, i'm using rewriting the the it's getting public here even the url have not. So here the url is localhost/fuel/posts and the generated link are localhost/fuel/public/post.

Thoughts?

johnnyfreeman:
The "public" folder is the public document root folder. The "fuel" folder is above root. See http://docs.fuelphp.com/installation/instructions.html#/manual

WanWizard:
And according to that same page (http://docs.fuelphp.com/installation/instructions.html#/install_inside_root) you should use a rewritebase to remove /public when rewriting. RTFM.

dbpolito:
I don't know if it's only happening here but i'm using this .htaccess and the URL is working properly but when i use an Html::anchor('posts/create', 'Add new Post'); the link generated is fuel/public/posts/create and give me error, then i have to change the url manually to fuel/posts/create and it works. Do you know what i mean?

WanWizard:
I you use the recommendation in the docs, with the rewritebase, the URI passed to the frontcontroller shouldn't contain /public, so no URI generated by FuelPHP will.

What does `echo $_SERVER['script_name'];` contain? If it contains /public, you have a confguration error.

dbpolito:
The page http://localhost/fuel/ has the SCRIPT_NAME /fuel/public/index.php

Is it only here? I'm using  PHP Version 5.3.6-13ubuntu3.2 with the default configuration. Here is my phpinfo page http://pastebin.com/t1S5V0xy

WanWizard:
Fuel can not detect whether or not the rewriting is configured properly. If you have a non-standard installation (which you have), you can fix this in the standard config file:
- set your index.php value to false
- set the base_url to `\Input::protocol().'://'.\Input::server('http_host').'/',`

No need to add something to the core code for this special case.

PR #696, opened 2011-11-28
Fix Arr::get \ArrayAccess bug

tomschlick:
Fixes a case where you cant access an \ArrayAccess object's data through Arr::get() because of a bug in array_key_exists

### DIFF
--- a/classes/arr.php
+++ b/classes/arr.php
@@ -55,13 +55,12 @@ class Arr
 
 		foreach (explode('.', $key) as $key_part)
 		{
-			if($array instanceof \ArrayAccess and isset($array[$key_part]))
+			if( ($array instanceof \ArrayAccess and isset($array[$key_part])) === false )
 			{
-				return $array[$key_part];
-			}
-			elseif ( ! is_array($array) or ! array_key_exists($key_part, $array))
-			{
-				return \Fuel::value($default);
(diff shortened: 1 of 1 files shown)


tomschlick:
btw unit tests ran clean

WanWizard:
This is an issue with all code that accepts an array(), and uses one of the array_... functions, as none support ArrayAccess.

@dhorrigan: any rules on what will get a workaround, and what not (trying to avoid non-consistent behaviour)?

dhrrgn:
Can't believe we didn't catch this before.  Anywhere we allow ArrayAccess should get a similar fix I suppose.

@tomschlick You have code formatting issues. `if( ($array instanceof \ArrayAccess and isset($array[$key_part])) === false )`  should be `if (($array instanceof \ArrayAccess and isset($array[$key_part])) === false)`.

WanWizard:
You mean anywhere we allow an array as argument? Hmm... Going to be a long night then...

dhrrgn:
No, just anywhere that we explicitly allow ArrayAccess.  I dont think it is many places really...I think a lot of time we have compensated for this (I know we do in the Security class).

WanWizard:
Which is nowhere. There are only a few classes that explicitly allow ArrayAccess, and none of them use array_ functions on the argument passed as far as I can see.

PR #701, opened 2011-11-30
Removed typecasting of values to strings because it makes querying using MongoIds impossible

alexbilbie:


alexbilbie:
FYI, bug was noticed by @tomschlick

tomschlick:
Tried this on my local copy and it seemed to fix the problem.

WanWizard:
I thought this was introduced as a security measure, to avoid Mongo's version of No-SQL injection?

See http://www.idontplaydarts.com/2010/07/mongodb-is-vulnerable-to-sql-injection-in-php-at-least/

philsturgeon:
Agreed, is this not intended to escape stuff? How else can we avoid the injection issue.

alexbilbie:
I always typecast before I call the library functions, e.g.

$this->mongo_db->where(array('username' => (string)$username))->get('users');

The problem with assuming strings is that you then can't use arrays, ints, floats, etc.

alexbilbie:
Just checked the PHP documentation [1] and it says that you should typecast before working with the database, but by typecasting everything as strings as I said before means you can't use integers, floats or arrays.

At the end of the day you should be sanitizing all data before inserting it into any DB not just Mongo

[1] http://php.net/manual/en/mongo.security.php

philsturgeon:
Sustained! Out it comes.

PR #706, opened 2011-12-04
added HttpServiceUnavailableException (error code 503)

tomschlick:
I can add the docs for this also but should probably wait until this https://github.com/fuel/docs/pull/261 pull is accepted and then add to it.


jschreuder:
I think this is getting a bit too specific, the 404 and 500 are pretty generic but this is looking like app code in the core and needs a bit of a case made for itself before I'll even consider it. At this point why not also 400, 401, 403 and 405 errors? (which I wouldn't consider a good idea btw, too much clutter)

tomschlick:
Now that I think about it you may be right on this. I'll just add this exception to my Wrench package and anyone who wants it can use/fork that.

PR #708, opened 2011-12-05
Enhancement: Enable definition of regex for routing named parameters and position independant options in route array translation

antitoxic:
For instance:

```
'_root_' => array('GET','welcome/home') 
// equals
'_root_' => array('welcome/home', 'GET')
```

Regex for named params are easily defined:

```
'user/view/:id' => array('users/view', array('id'=>'[0-9]+'))
```


jschreuder:
Thank you for your contribution, but I do have a quality issue: the code is not up to our [coding standards](http://fuelphp.com/docs/general/coding_standards.html) and will never be accepted before it does.

Also, while I like it, I'm not yet sure if I agree with all this. @dhorrigan will have to take a look as well as the Router is mostly his code. You've added 2 features in a single pull-request (not caring about the order, using regex for named params) which is also not the best idea because it doesn't allow us to pick & choose easily if we were to like 1 proposal but not the other.

antitoxic:
Sorry I haven't set PHPStorm formatting properly for fuel project. I can revert the master and introduce only one of :
- position independant route ptions
- regex for route named parameters

And then I can add the other feature in a new branch.

I guess regarding the coding standards I'll just have to reformat the code to match the format style. I guess braces, spaces after commas...

Will this be ok?

dhrrgn:
I will take a look at this tomorrow (on mobile now).

antitoxic:
Any updates on this ?

alexwhitman:
I'd definitely like to see this go in so that reverse routing can work with routes using regex.

antitoxic:
@whitman There are still problems for reverse routing as Fuel basically allows any kind of regex as a pattern.
Things like :

```
hello(/:name)?/:test
```

The problem is how to detect and remove the ( ), >... and any unmatched optional content. The reverse routing won't work until there's a distinct set of allowed expressions  or at least a Fuel convention what can be used inside the route pattern.

Another way is actually to parse the route regex pattern with a second regex which will search and remove such rendundant tokens :)

alexwhitman:
Hmm, I was thinking something like

```
'hello:name/:test' => array('hello/view', array('name'=>'(/[a-z]+)?', 'test'=>'[0-9]+'))
```

could be used, but maybe not.

antitoxic:
That would mean that you'll receive the "/" in your parameter :)

Anyways I think the example above is not completely fair as currently my commit considers all named parameters as required.
We need that second regex that will allow the reverse routing and also put some context for named parameters. Such as "optional", "multiple"...

alexwhitman:
In the example you would get a "/" but you could use `(?:/([a-z]+))?` to work around that (using a non-capturing group).  It makes the regex a bit messier but could work with some tweaks.

antitoxic:
whitman, still the Router must somehow know `:name` is optional so that when it attempts to reverse engineer the URL it doesn't require it.
May be the Router can check if there's a "/" before the named parameter declaration then it's required?

antitoxic:
@jschreuder I've changed the code to match conventions. I tried to contact @dhorrigan but with no success.

antitoxic:
Any updates on this? I've made several changes since then. 
I've enabled the router to match routes from other modules with minimal overhead (loading and checking routes one by one).
The latter helped me make `\Router::get(..)` function properly even with routes from modules.

jschreuder:
This really is @dhorrigan's code, and these are pretty big changes I don't feel comfortable applying without hearing his opinion on it. I'll poke him on Twitter...

antitoxic:
Added more fixes and changes.

I'll be repeating myself from the commit's message but still:
- The commit adds support for reverse routing for routes defined in modules. This is done with minimum overhead. Routes from modules are not parsed until a route that is not in the main `routes.php` is look-uped. Only the routing configuration is loaded on fuel init.
- `Routes::get()` also accepts a third parameter `secure` to force the generated url as secure.
- Routes in routes.php now accept extra option `secure` which force the urls generated by reverse routing to secure.
- Route parameters also accept transformations/filters. Currently only `prefix` filter is implemented.
- Some formatting

jschreuder:
I studied your code a bit deeper just now. I have two problems with this:
- loading routes from all the modules, whether necessary or not will slow fuel down on each and every request and make modules a resource hog that it wasn't before.
- a lot of your modifications to the route object are a bit specific and over complicated for most. While nice they have a better place in an add-on, one of the reasons why we made it easy to extend classes such as these.

All in all your modifications don't feel like core material. But they would make an awesome package and probably be usefull to others. If you make a package out of it I'll fork in into github.com/fuel-packages to make it easily installable through oil.

antitoxic:
@jschreuder Nice to hear some decision on this :)

I'll try to extract the logic into a package as you suggested.

**However** as I wrote before:  routes from the modules **are not loaded** on every request. Only their configuration is. The only time the configuration will become an actual route object will be the occasions where no core route can be found.
So no actual unnecessary overhead.

I can refactor the code to feel like core material?

PR #709, opened 2011-12-05
Added headers option for to_csv() function

aranw:
I wanted to use the to_csv() function but it kept adding the headers and didn't need them headers so added this.

### DIFF
--- a/classes/format.php
+++ b/classes/format.php
@@ -190,9 +190,10 @@ class Format
 	 * To CSV conversion
 	 *
 	 * @param   mixed   $data
+	 * @param   mixed   $headers
 	 * @return  string
 	 */
-	public function to_csv($data = null)
+	public function to_csv($data = null, $headers = null)
 	{
 		if ($data == null)
 		{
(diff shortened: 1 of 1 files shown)


jschreuder:
Looks good to me but this is more @philsturgeon 's area.

aranw:
This might actually need some more testing, in one situation I noticed odd behaviour where nothing was returned. 

Unfortunately at the time I forgot to take a note of the issue. I'll try find it again.

classes/format.php
```diff
@@ -212,7 +213,12 @@ public function to_csv($data = null)
 			$data = array($data);
 		}
 
-		$output = implode(',', $headings) . "\n";
+		// Does the user want headers?
+		if($headers)
```
philsturgeon: If this is a boolean option can we switch it to `= false` and change the name to reflect that? `$include_headers` or something.

Also, this should be `if (` - notice the space.

Thanks for the pull request, it could be a handy feature.

aranw: Yea I'll make some improvements and try work out that issue, mentioned above.

aranw:
Ooppss pressed wrong button :-/

aranw:
Looking at the code in more detail I think the lack of returned data might of been due to my usage.

WanWizard:
Closing this due to lack of further information. 

There is a separate issue to allow the csv() methods to be parameterized to deal with the different CSV formats out there.

PR #710, opened 2011-12-05
Added encoding based on content type for cUrl requests

nedmas:
I needed to send a cUrl PUT request with an XML payload, so I wrote an encode method to check the content type and use the Format class to encode the data.

### DIFF
--- a/classes/request/curl.php
+++ b/classes/request/curl.php
@@ -269,8 +269,19 @@ class Request_Curl extends \Request_Driver
 		{
 			// Format as XML
 			case 'xml':
-					$base_node = key($input);
-					return \Format::forge($input[$base_node])->to_xml(null, null, $base_node);
+					/**
+					 * If the input array has one item in the top level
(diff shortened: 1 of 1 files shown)


philsturgeon:
This is very close to what I have in mind but not quit. By using "supported_formats" you are making `application/json` work but not `text/json` for example.

You should really be using the `static::$auto_detect_formats` array which is designed for this sort of thing, but it means you will need to rejig the code somewhat.

nedmas:
Thanks for the pointer Phil, I completely missed that. I'll get on and change the code and update the request.

nedmas:
@philsturgeon is this closer to what you had in mind?

philsturgeon:
Magic!

philsturgeon:
As an afterthought, while this is helpful for curl, if this can be moved into the driver logic more that would be very handy, as of course HTTP Stream wrapper driver for example would also need similar logic.

PR #714, opened 2011-12-07
Return empty array/false if using Finder::locate() with absolute path (when it's not a valid file)

crynobone:
When using `Finder::search()` with relative path it will check for `is_file()` https://github.com/fuel/core/blob/1.1/develop/classes/finder.php#L332 however the same action doesn't occur when absolute path is provided https://github.com/fuel/core/blob/1.1/develop/classes/finder.php#L279

### DIFF
--- a/classes/finder.php
+++ b/classes/finder.php
@@ -275,13 +275,20 @@ class Finder
 	 */
 	public function locate($dir, $file, $ext = '.php', $multiple = false, $cache = true)
 	{
+		$found = $multiple ? array() : false;
+
 		// absolute path requested?
 		if ($file[0] === '/' or $file[1] === ':')
 		{
+			if ( ! is_file($file))
+			{
+				// at this point, found would be either empty array or false
+				return $found;
+			}
+
 			return $multiple ? array($file) : $file;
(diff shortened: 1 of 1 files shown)


Savageman:
Hello,

This commit broke something I used, but I agree it's more logical now.

I were saving new config files to a specific location using an absolute path Config::save('/absolute/path/', array())
Now it doesn't work anymore, because Config uses the Finder, which returns false, so the save() function uses the path APPPATH/config/{my_absolute_path_here_with_extension}.php}

How should it work? Maybe the config::save() function should not call the finder when using absolute file?

WanWizard:
It is intended behaviour to save in app/config. The idea behind it is that when you load it, it might be loaded from fuel/core/config or from a package, but you save it to app, thus automatically overriding the config provided by default.

If we need to be able to save config files anywhere (@dhorrigan?), that it should be a check on an absolute path in Config::save(), and bypass the finder call in that case...

Savageman:
Should I create an issue to save config file using an absolute path?

WanWizard:
yes please.

PR #719, opened 2011-12-08
input protocol & domain forward / cloudfront / whatever magic

julesjanssen:


classes/input.php
```diff
@@ -192,7 +192,9 @@ public static function real_ip($default = '0.0.0.0')
 	public static function protocol()
 	{
 		if ((static::server('HTTPS') !== null and static::server('HTTPS') != 'off')
-			or (static::server('HTTPS') === null and static::server('SERVER_PORT') == 443))
+			or (static::server('HTTPS') === null and static::server('SERVER_PORT') == 443)
+			or (static::server('HTTP_X_FORWARDED_PROTO') !== null and static::server('HTTP_X_FORWARDED_PROTO') != off)
```
jschreuder: Did you test this with error reporting on?

julesjanssen: Ah, crap, forgot to include that ' commit.
I can't add that commit to this pull request for some reason though.

WanWizard:
i'm not sure this sort of app specific settings should belong in the core. We had the same dicussion recently about reverse proxies, which need similar settings.

I think if you need something like this, deal with it on the app side, for example by setting $_SERVER['HTTPS'] to 'on' based on the HTTP_X_FOWARDED_PROTO value, or whatever setting you need...

julesjanssen:
Agreed. I just wasn't sure how 'app specific' this is and wanted to let you decide. 
Thanks.

crynobone:
Would be helpful if we can setup a knowledge base center for app's specific configuration.

PR #721, opened 2011-12-11
Add Autoloader paths to profiler

kbanman:
Screenshot: http://i.imgur.com/sV3d9.png

It makes it a bit crowded--maybe combine the GET and POST tabs?


WanWizard:
Crowded, indeed.

I would prefer adding it to the Files tab (in front of the files list), and leave GET and POST seperate.

PR #722, opened 2011-12-11
fix namespace of HttpException

kenjis:
When loading httpexception.php, there is no alias \HttpException yet.
So it causes Class 'HttpException' not found Error.

How to reproduce:

app/myexception.php:

``` php
class MyException extends \HttpException
{
    ...
}
```

``` php
class Controller_Welcome extends Controller
{
    {
        throw new MyException('...');
    }
}
```

### DIFF
--- a/classes/httpexception.php
+++ b/classes/httpexception.php
@@ -36,7 +36,7 @@ class HttpNotFoundException extends \Request404Exception
 	}
 }
 
-class HttpServerErrorException extends \HttpException
+class HttpServerErrorException extends HttpException
 {
 	/**
 	 * When this type of exception isn't caught this method is called by


jschreuder:
It is right for those classes to extend HttpException from global as it allow you to add methods/properties to the base exception that will be inherited by all. The problem is indeed that while the HttpException class is being loaded it's extended a second time, a circular load the PHP interpreter doesn't allow.

The fix shouldn't be to disallow extending the HttpException in a way that no longer allows extending it in a way that is inherited. We should either move the HttpException so it can be loaded independently from the others, or make sure it's loaded before anyone extends it.

@dhorrigan thoughts?

WanWizard:
@dhorrigan @jschreuder Any progress on this topic?

WanWizard:
You can't extend this class from global.

If you do, and debug the autoloader, you'll see that the core/classes/httpexception.php file is loaded, but the next statements (including class_alias) are never executed. The 'class HttpException' not found triggers directly, probably due to the fact that this is an abstract class.

If you do an

```
import('httpexception');
```

before you throw an exception that extends \HttpException, it works without problems.

So it's something in the autoloader that doesn't work as it should when it comes to abstract classes...

jschreuder:
You currently can't extend it because it's all in the same file, read my earlier comment where I describe exactly what goes wrong (tl;dr: circular class load). The way PHP class autoloading works doesn't care what type of class is being loaded, whether it's abstract, an interface or a normal class doesn't matter.

WanWizard:
Ah, wait a sec. 

The issue here isn't HttpException, but HttpServerErrorException in that same file, that also extends \HttpException which at that time is still in the process of being loaded.

The penny has dropped...

jschreuder:
The fix is very easy: move the class to another (it's own?) file. This problem will always occur when both a base class and the class that extends it through global are in the same file. We just have to decide whether we put this into its own file (requireing another file load) or make this part of some global file like base.php (I vote for the latter, it's a very basic class and it'll save us a file load).

[edit] though the additional file load only occurs on errors, so maybe it doesn't really matter.

alexwhitman:
I was thinking about error exceptions recently, how about having

`HttpException`
`HttpException_NotFound`
`HttpException_ServerError`

It would mean an extra file load but as @jschreuder says, it'll only happen on error.  Then there's the option of adding more exceptions such as `HttpException_MethodNotAllowed`, `HttpException_NotAcceptable` etc.  Thoughts?

WanWizard:
If you really want to cover all corners, create an exception for every HTTP status 400 and above. 

They can all be stored in the same file, since they all extend HttpException, which is (going to be) in a separate file.

jschreuder:
@whitman we decided on just camelcasing for exceptions for consistency with the build-in exceptions. Even though the use of an underscore is explainable I would prefer to keep exceptions by convention to just camelcase.

@WanWizard I think the way to go is: one file httpexception.php with just the abstract exception and  second httpexceptions.php with the 2 extending HTTP exceptions. But we'll keep the core extensions to just 404 and 500, the others can be app-specific extensions if and when devs need them.

If no one objects (@dhorrigan) I'll update this tomorrow.

alexwhitman:
@jschreuder I was thinking along the lines of `classes/httpexception.php` and `classes/httpexception/notfound.php` but keeping them consistent is fine by me.

WanWizard:
Once the current file is split, you can do that in app/classes, and just extend \HttpException.

PR #723, opened 2011-12-12
Added support for multiple views

billmn:
Permit to specify an array of view and the class set the first view found


billmn:
Obviously, if this pull is accepted, I can make changes to the Parser package to support this

jschreuder:
I don't really like this, is the type of code that has more of a place in an app than the core.

@philsturgeon @dhorrigan @WanWizard @FrenkyNet thoughts?

WanWizard:
The only reason I can think of where this might be useful is if you have a templating system where it is possible that views from app or modules can be overridden by the template. 

Which is exactly what the Theme class does.

frankdejonge:
I don't think there is a need to have this feature as a duplicate. Like @WanWizard says, the Theme class does this, use that.

jschreuder:
That's 3 votes against.

PR #727, opened 2011-12-13
Fix issue with calling Profiler with profiling off

kbanman:
Having profiling turned off in config, then calling the Profiler, e.g. `\Profiler::mark('Start something');` currently throws an error (`class Console not found`) because the profiler has not yet been initialized. This patch calls `init()` as soon as it's autoloaded.

Ideally, if profiling is turned off, each method would simply return upon being called and the class would never need to be initialized, but it would be ugly to have that code at the beginning of each method. If you think that's a better way to go, I'll gladly make those changes.

### DIFF
--- a/classes/profiler.php
+++ b/classes/profiler.php
@@ -14,11 +14,23 @@ class Profiler
 
 	protected static $query = null;
 
+	protected static $initialized = false;
+
+	public static function _init()
+	{
+		static::init();
+	}
+
 	public static function init()
 	{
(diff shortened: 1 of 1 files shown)


WanWizard:
@dhorrigan, Is this a smart idea?

Ideally, you would want to have debugging code present in your application, that does nothing when debugging is not activated. This way, the profiler and all it's classes will be always loaded, if only because the core contains Mark() calls.

PR #728, opened 2011-12-14
Added querystring config param. Querystrings are appended to links.

rob-mccann:
Has been asked for on the forums. I use it to paginate search results.

querystring can either be a string or an array or key/value pairs.

``` php
\Pagination::set_config(array(
    'pagination_url' => '/search/',
    'total_items' => $total,
    'per_page' => $per_page,
    'uri_segment' => 2,
    'querystring' => array('q' => \Input::param('q',''))
));
```

### DIFF
--- a/classes/pagination.php
+++ b/classes/pagination.php
@@ -73,6 +73,11 @@ class Pagination
 	 * @var	mixed	The pagination URL
 	 */
 	protected static $pagination_url;
+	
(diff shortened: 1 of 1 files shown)


jschreuder:
As per #601 Pagination will be completely rewritten for the next version, we are no longer accepting pull-requests on it.

PR #729, opened 2011-12-14
Adds attribute support to pagination link creation. Also allows an overr...

eggsurplus:
...ide to show the link even if it matches the current page condition. Example pagination config:

'pagination' => array(
        'pagination_url' => 'http://fuelphp.com/docs/',
        'uri_segment' => 3,
        'per_page' => 5,
        'template' => array(
            'wrapper_start' => '<div class="ui-block-b pagination" style="width: 50%; text-align: right">',
            'wrapper_end' => ' </div>',
            'page_start' => '',
            'page_end' => '',
            'previous_start' => '',
            'previous_end' => '',
            'previous_mark' => '',
            'next_start' => '',
            'next_end' => '',
            'next_mark' => '',
            'active_start' => '',
            'active_end' => '',
            'show_current_link' => true,
            'previous_link_attrs' => array(
                'data-role' => 'button',
                'data-icon' => 'arrow-l',
                'data-inline' => 'true',
                'data-theme' => 'b'
            ),
            'previous_current_link_attrs' => array(
                'data-role' => 'button',
                'data-icon' => 'arrow-l',
                'data-inline' => 'true',
                'data-theme' => 'd',
                'class' => 'pageactive ui-disabled',
                'disabled' => 'disabled',
            ),
            'next_link_attrs' => array(
                'data-role' => 'button',
                'data-icon' => 'arrow-r',
                'data-iconpos' => 'right',
                'data-inline' => 'true',
                'data-theme' => 'b'
            ),
            'next_current_link_attrs' => array(
                'data-role' => 'button',
                'data-icon' => 'arrow-r',
                'data-iconpos' => 'right',
                'data-inline' => 'true',
                'data-theme' => 'd',
                'class' => 'pageactive ui-disabled',
                'disabled' => 'disabled',
            ),
            'page_link_attrs' => array(
                'data-role' => 'button',
                'data-inline' => 'true',
                'data-theme' => 'b',
                'class' => 'pagenum',
            ),
            'page_current_link_attrs' => array(
                'data-role' => 'button',
                'data-inline' => 'true',
                'data-theme' => 'd',
                'class' => 'pagenum pageactive ui-disabled',
                'disabled' => 'disabled',
            ),
        ),
    ),

### DIFF
--- a/classes/pagination.php
+++ b/classes/pagination.php
@@ -186,6 +186,7 @@ class Pagination
 		}
 
 		$pagination = '';
+		$show_current_link = (isset(static::$template['show_current_link']) && static::$template['show_current_link'] == true? true: false); //eggsurplus: force the link creation even if it current page
 
 		// Let's get the starting page number, this is determined using num_links
 		$start = ((static::$current_page - static::$num_links) > 0) ? static::$current_page - (static::$num_links - 1) : 1;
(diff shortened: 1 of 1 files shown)


WanWizard:
This is exactly why we don't accept pull requests on Pagination at the moment, and the entire class is under review.

This is very application specific, and has no place in the core. There is already a ticket open to refactor the class.

eggsurplus:
I disagree. But that is because I'm not sure what your direction is with it. This falls in line with the current implementation of passing formatting to the pagination class. That class currently forces a very stringent UI to any user. This brings greater flexibility by allowing users to specify the attributes that should be applied to a link.

Just trying to help this project out. I have no problem not wasting my time in the future if help is not desired at this time.

WanWizard:
We're not happy with the current implementation either. ;-)

The plan is to abstract it even more, and use callbacks so all HTML generating code will be in your application, not in the core class. And then you can make it as fancy as you want...

eggsurplus:
I completely agree with that! Alright, no problem.

PR #735, opened 2011-12-18
Fixing docblock in mongo db

dmyers:


WanWizard:
too bad you've missed the "passwed" in that sentence... ;-) Can you update that too?

PR #736, opened 2011-12-19
Fixing docblock in mongo db

dmyers:


ipalaus:
Can you also change `@usage $mongodb->get('foo');` to `@usage   $mongodb->get_one('foo');` in the same dockblock? My fault. thanks!

PR #738, opened 2011-12-20
Allow request method overriding

alexwhitman:
- Send X-HTTP-Method-Override header to override request method

Allows for PUT/DELETE to be used as the request method.


philsturgeon:
I've merged this, but if you can find somewhere appropriate to document this I would appreciate it.

alexwhitman:
Thanks, documented in https://github.com/fuel/docs/pull/279

PR #749, opened 2011-12-23
Add the controller_name property to the active Request object.

fgelinas:
When the controller class name is resolved by the router and saved to the Request
object, this will also save the base controller name along with it.

For example, if the loaded controller class for the request is 'Controller_Welcome',
then the controller_name would be 'welcome'.

It can then be retrieved with the following command :
  Request::active()->controller_name

### DIFF
--- a/classes/request.php
+++ b/classes/request.php
@@ -339,6 +339,7 @@ class Request
 
 		$this->module = $this->route->module;
 		$this->controller = $this->route->controller;
+		$this->controller_name = $this->route->controller_name;
 		$this->action = $this->route->action;
 		$this->method_params = $this->route->method_params;
 		$this->named_params = $this->route->named_params;
--- a/classes/route.php
+++ b/classes/route.php
@@ -46,9 +46,14 @@ class Route
 	public $directory = null;
 
 	/**
-	 * @var  string  controller name
+	 * @var  string  controller class name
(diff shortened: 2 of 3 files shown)


WanWizard:
This has been discussed before, and denied, as it is not the controller name, it is a part of the name. And by no means a unique identifier of the controller.

Can you give a use-case for this feature?

fgelinas:
I am writing a template controller that will render a default view if no view where loaded by the action. 
This new property will correspond to the name of the view folder for this controller. 

```
    public function after($response)
    {
        [...]
        if ($this->content == '') {
            $file = Request::active()->controller_name.'\\'.Request::active()->action;
            $this->content = \View::forge($file);
        }
        [...]
    }
```

WanWizard:
So it's application specific code, and should not be a part of the core.

It can easily be replaced by substr(Request::active()->controller, 11). And don't use '\', use the DS constant as a path separator, it makes your code platform independent.

Also consider the namespace. If you have modules your application can have multiple controllers with the same name. Which in this case would cause the wrong view to be loaded.

fgelinas:
That will work for me, you are right to keep the core as clean as possible. 
Thanks for the advice about the directory separator and the namespace.

Good job on this framework, I like it !

PR #754, opened 2011-12-28
Modified profiler template

purwandi:
I'm sory for pull request before.
This code now work. See my screen capture http://d.pr/3gPl.


WanWizard:
Can't be merged anymore. Profiler needs to be looked at towards 2.0.

purwandi:
Ok i will update my branch

WanWizard:
You don't have to, we're not going to merge this.

I think the profiler needs to be looked at towards 2.0, because I don't like the restrictive way the current one works.

PR #755, opened 2011-12-29
Validation::add_fields to pass an array containing multiple validation rules at once

tgriesser:
This allows us to add multiple rules to a validation object at once with a single array.

Example:

``` php
$_validation['login'] = array(
    array('username', 'Username', 'required|trim|valid_string[alpha,lowercase,numeric]'),
    array('email', 'Email', 'required|trim|valid_email'),
    array('password', 'Password', 'required|trim')
);

$val = Validation::forge('login')->add_fields($_validation['login']);
if ($val->run())
{
   ...
}


```

### DIFF
--- a/classes/validation.php
+++ b/classes/validation.php
@@ -207,6 +207,28 @@ class Validation
 	}
 
 	/**
+	 * Allows multiple fields to be set with a single array
+	 * @param   string      Field array
+	 * @return  Validation  this, to allow chaining	
+	 */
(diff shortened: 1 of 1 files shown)


jschreuder:
This is the type of thing that is more app logic than a core necessity. It adds code to the core but no functionality. There's a lot of these little things that might be usefull to some, but not necessary or even interesting for all.

PR #761, opened 2012-01-04
Detect rest accept format by quality

alexwhitman:
The current checking of HTTP_ACCEPT only looks to see if a format is in the header but doesn't do anything with the quality scoring.  This makes the search dependent on the order that the formats are defined in `$this->_supported_formats`

This patch will take an Accept header such as:

```
Accept: application/json,application/xml,text/html;q=0.9;text/plain;q=0.8;*/*;q=0.5
```

and assign scores to each format:
- application/json: 1
- application/xml: 1
- text/html: 0.9
- text/plain: 0.8
- _/_: 0.5

The highest scoring format that is supported will be returned.  If multiple supported formats share the same score then the first one is chosen.  Order of same value formats in the Accept header is kept.

Thoughts/comments/merge?

### DIFF
--- a/classes/controller/rest.php
+++ b/classes/controller/rest.php
@@ -191,6 +191,7 @@ abstract class Controller_Rest extends \Controller
 			$next_is_quality = false;
 			foreach ($fragments as $fragment)
 			{
+				$quality = 1;
 				// Skip the fragment if it is a quality score
 				if ($next_is_quality)
 				{
@@ -201,7 +202,6 @@ abstract class Controller_Rest extends \Controller
 				// If next fragment exists and is a quality score, set the quality score
 				elseif ($fragments->hasNext())
 				{
-					$quality = 1;
 					$next = $fragments->getInnerIterator()->current();
 					if (strpos($next, 'q=') === 0)
 					{
--- a/classes/controller/rest.php
+++ b/classes/controller/rest.php
@@ -206,7 +206,6 @@ abstract class Controller_Rest extends \Controller
 					if (strpos($next, 'q=') === 0)
 					{
 						list($key, $quality) = explode('=', $next);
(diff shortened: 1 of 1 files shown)


philsturgeon:
Why was this closed?

alexwhitman:
Closed it by accident, I had multiple github tabs open.  Reopened and pushed a small fix.

classes/controller/rest.php
```diff
@@ -184,32 +184,55 @@ protected function _detect_format()
...
+			{
+				// The Accept header can contain wildcards in the format
+				$find = array('*', '/');
+				$replace = array('.*', '\/');
+				$pattern = '/^' . str_replace($find, $replace, $pattern) . '$/';
+				foreach ($this->_supported_formats as $format => $mime)
+				{
+					if (preg_match($pattern, $mime))
 					{
-						// If it is truly HTML, it wont want any XML
```
philsturgeon, line 202: Why was this code removed?

alexwhitman, line 202: It's not required any more.  This patch finds the most preferable format of formats supported, rather than finding the first format supported.  Firefox 9 sends this as Accept: `text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8`

`text/html` and `application/xhtml+xml` will both have a quality of 1.  `text/html` will, and should, always be picked first if `html` is supported so it's not necessary to do any further checks.  `application/xml` has a quality of 0.9 but even if it had a quality of 1 it still wouldn't be chosen.  The only way it would be chosen with a quality of 1 is if `text/html` wasn't supported or `application/xml` was defined before `text/html` in the header.

philsturgeon, line 202: Thanks for the clarification. :)

PR #763, opened 2012-01-05
Validation::add_field() should accept an array for list of rules

crynobone:
Instead of preparing each rules as a piped string, it would be nice if we can provide an array.

I had a problem using match_pattern with `|` regex because `Validation::add_field` would assume it a separator between two rule and cause validation can't validate the rules properly.
## Example

```
$val->add_field('gender', 'Gender', 'required|match_pattern[/^(m|f)$/]');
```
## Alternative

I know you can use `$val->add_field('gender', 'Gender')->add_rule('match_pattern', '/^(m|f)$/');` but the pull request should work the same way without much changes.

### DIFF
--- a/classes/validation.php
+++ b/classes/validation.php
@@ -167,14 +167,15 @@ class Validation
 	 *
 	 * @param   string      Field name
 	 * @param   string      Field label
-	 * @param   string      Rules as a piped string
+	 * @param   mixed       Rules as a piped string or an array
 	 * @return  Fieldset_Field  $this to allow chaining
 	 */
 	public function add_field($name, $label, $rules)
 	{
 		$field = $this->add($name, $label);
 
-		$rules = explode('|', $rules);
+		is_string($rules) and $rules = explode('|', $rules);
+
 		foreach ($rules as $rule)
 		{
 			if (($pos = strpos($rule, '[')) !== false)


jschreuder:
Why aren't you just using add() if you want to pass the rules as an array? The only reason add_field() was ever added was to allow passing rules as a string...

WanWizard:
If both do the same (apart from the type of the rules argument), what is the point of having two methods that do the same?

Would it not have been easier to have add do the reverse of this, and accept both string and array?

jschreuder:
No, because this is far inferior and I wanted a clear seperation between the methods no act like they're equal. If it'd been purely up to me this string business wouldn't have gotten in in the first place. The method chaining is the primary way to do it, array is a secundary option and the string stuff is really there to help CI people get started. Allowing something in 3 ways is foolish as it will always lead to inconsistent behavior (you can't do it exactly the same in every way). That's why the string stuff was seperated out into its own method.

Now I think about it more, this should be reverted for that very reason. Though maybe the method could use renaming to underscore it, but in the end I care little for the string stuff.

WanWizard:
Ah, ok. I agree. 

Reading your comments I think we should deprecate the string stuff. CI folks have to learn a lot of new tricks anyway, one more won't hurt. Now we're sort of allowing them to stick to their bad habits... ;)

crynobone:
@WanWizard I would prefer not to use string as well, the problem with it is that it can't differentiate `|` for rules or rules option inside `[]`, these inconsistent/gotcha is not documented anywhere and will be a turn-off for newcomer either from CI or not. 

@jschreuder thanks for pointing out about `add()` method, I must have missed it. 

BTW `add()` and `add_rules()` are among things not documented yet in http://docs.fuelphp.com/classes/validation/methods.html

jschreuder:
@WanWizard I'll deprecate it for v2 - should be kept for the 1.x branch if we end up releasing a v1.2.

@crynobone that and a lot of other things, as always all help is welcome.

UPDATE: done in 51e9677267556cfe8aae83486126c95a86f78bb4

crynobone:
will try to do it when I'm free but no promises.

On Saturday, January 7, 2012, Jelmer Schreuder <
reply@reply.github.com>
wrote:

> @WanWizard I'll deprecate it for v2 - should be kept for the 1.x branch
> if we end up releasing a v1.2.
> 
> @crynobone that and a lot of other things, as always all help is welcome.
> 
> ---
> 
> Reply to this email directly or view it on GitHub:
> https://github.com/fuel/core/pull/763#issuecomment-3388561

## 

Sent from Gmail Mobile

PR #765, opened 2012-01-06
Throws an exception when asset file is not found

crynobone:
For some reason, the new Asset class no longer throws an exception when asset file is not available. I'm not sure whether this is an expected behavior with the new class but it does help in my use case to be able to do a fallback for asset.


WanWizard:
Oops, sorry. leftover from some testing...

PR #773, opened 2012-01-14
Add the template options to next_link and previous_link method

faabiosr:
Including next_start and next_end to next_link function.
Including previous_start and previous_end to prev_link function.

### DIFF
--- a/classes/pagination.php
+++ b/classes/pagination.php
@@ -227,12 +227,12 @@ class Pagination
 
 		if (static::$current_page == static::$total_pages)
 		{
-			return $value.static::$template['next_mark'];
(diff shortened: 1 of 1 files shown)


notrab:
I'm currently having this issue and would love the ability to set template options for the previous/next links without having to use an ExtendedPagination Class.

faabiosr:
Me too, but the previous/next link option is not implemented. The options are defined in code, but don't work.

jschreuder:
The pagination class as it currently exists will be replaced (#601), thus we do not accept any pull-requests on it anymore.

faabiosr:
OK, no problem.

PR #774, opened 2012-01-14
Allow router to skip controller lookup if no defined routes match

alexwhitman:
Allows for only defined routes to be processed, rather than searching for a controller that might match.


jschreuder:
One can add a final route like this `':any' => 'error/route404'` that routes everything to your 404 if nothing matched. It does basicly the same without adding complixity.

PR #779, opened 2012-01-18
Adding exceptions to the date class to make catching and handling unexpe...

bencorlett:
...cted values easier.

This makes it possible to catch an error when date's aren't valid and take action appropriately, instead of getting a warning show up and do nothing about it.

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -113,7 +113,8 @@ function setup_autoloader()
 
 		'Fuel\\Core\\Finder'         => COREPATH.'classes/finder.php',
 
-		'Fuel\\Core\\Date'    => COREPATH.'classes/date.php',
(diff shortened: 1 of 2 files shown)


frankdejonge:
I think an UnexpectedValueException would be more in order. @jschreuder ?

jschreuder:
First of all: Thanks @bencorlett, this has needed doing for some time and I kept putting it off.

But I agree with @FrenkyNet, there's no real reason for a specialized exception here. Just the unextended `\UnexpectedValueException` would be fine.

bencorlett:
Okay will fix that in the morning :)

bencorlett:
Sorry mate forgot all about it. I'll do it now.

bencorlett:
There ya go @jschreuder

PR #786, opened 2012-01-23
added session task to create db table

dberry37388:
This is a simple oil task that creates, deletes or truncates the sessions table when using db as the driver.

### DIFF
--- a/tasks/session.php
+++ b/tasks/session.php
@@ -101,4 +101,4 @@ HELP;
     }
 }
 
-/* End of file tasks/robots.php */
+/* End of file tasks/session.php */
--- a/tasks/session.php
+++ b/tasks/session.php
@@ -166,4 +166,4 @@ HELP;
     }
 }
 
-/* End of file tasks/session.php */
+/* End of file tasks/session.php */ 
--- a/tasks/session.php
+++ b/tasks/session.php
@@ -34,7 +34,8 @@ namespace Fuel\Tasks;
  * php oil r session:clear   = clear the sessions table
  */
 
-class Session {
+class Session
+{
 
     // default function if no command is selected. Provided user with menu
     public static function run()
@@ -88,7 +89,7 @@ class Session {
                 'created'      => array('constraint' => 10, 'type' => 'int', 'unsigned' => true),
                 'updated'      => array('constraint' => 10, 'type' => 'int', 'unsigned' => true),
                 'payload'      => array('type' => 'longtext'),
-            ), array('session_id'), false, 'InnoDB', 'utf8');
+            ), array('session_id'), false, 'InnoDB', \Config::get('db.default.charset'));
 
             // make previous_id a unique_key. speeds up query and prevents duplicate id's
             \DBUtil::create_index(\Config::get('session.db.table'), 'previous_id', 'previous_id', 'unique');
--- a/tasks/session.php
+++ b/tasks/session.php
@@ -1,22 +1,45 @@
 <?
 /**
- * Sessions DB Table Task
+ * Fuel is a fast, lightweight, community driven PHP5 framework.
  *
- * Run this task to set add/remove/clear the db sessions table
- * for your app. This could be expanded in app/tasks for application specific stuff.
+ * @package    Fuel
+ * @version    1.1
+ * @author     Fuel Development Team
+ * @license    MIT License
+ * @copyright  2010 - 2011 Fuel Development Team
+ * @link       http://fuelphp.com
  *
- * @version		1.0
- * @author		Daniel Berry
- * @license     MIT License
  */
 
 namespace Fuel\Tasks;
 
+/**
+ * Sessions DB Table Task
+ *
+ * @PullRequest https://github.com/fuel/core/pull/786
+ *
+ * Run this task to set add/remove/clear the db sessions table
+ * Table name will be generated from your config file.
+ * for your app. This could be expanded in app/tasks for application specific stuff.
+ *
+ * @package     Fuel
+ * @version		1.1
+ * @author		Daniel Berry
+ * @license     MIT License
+ *
+ * Usage:
+ * php oil r session         = will prompt with menu
+ * php oil r session:create  = create the db table.
+ * php oil r session:remove  = remove the sessions table
+ * php oil r session:clear   = clear the sessions table
+ */
+
 class Session {
 
+    // default function if no command is selected. Provided user with menu
     public function run()
     {
-        // Will only accept the options in the array
+        // Prompt the user with menu options
         $option = \Cli::prompt('What would you like to do?', array('create','remove', 'clear', 'help'));
 
         switch($option)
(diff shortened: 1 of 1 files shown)


philsturgeon:
Why?

dberry37388:
Well, Oil is meant to automate tasks, this automates the task of setting up the session database table.  Is it borderline laziness? yes, but then again that is what Oil is about. but I was in the IRC channel and after several times hearing people ask about it decided to throw it together.  90% of the people using FuelPHP are probably using sessions at some point, and 99% of those people would rather run a task than have to go to the docs to copy/paste the table structure, especially if they are already sitting right there in front of the cmd line.

-----Original Message-----

philsturgeon:
Ok the up/down threw me a little there and made the whole thing seem a little off.

Tell you what, make it "add/remove" and maybe throw in "clear" to empty the sessions table, and we've got a deal! :)

dberry37388:
yeah sorry, i don't know why I did up/down... I made those changes and added a default run() method that will give the user a menu of choices to create/remove/clear/help....

philsturgeon:
This looks fine now. If you can document it and comment with a link to that pull request in this one then we can merge it.

dberry37388:
Documentation pull request https://github.com/fuel/docs/pull/310

aranw:
Just wondering does this need a check for if a database is configured? or perhaps a specific connection? 

Some people may use different databases for sessions. 

Other than that this looks good and I can imagine myself using :) 

+1 for this.

dberry37388:
It currently uses the sessions config file to check the driver to make sure it is "db". If it is not "db", than it tells you its not.  If your actual database is not configured, it will fail with standard error message.

tasks/session.php
```diff
@@ -0,0 +1,169 @@
...
+    // default function if no command is selected. Provided user with menu
+    public function run()
+    {
+        // Prompt the user with menu options
+        $option = \Cli::prompt('What would you like to do?', array('create','remove', 'clear', 'help'));
+
+        switch($option)
+        {
+            case "create":
+                return self::create();
```
frankdejonge: This should be static::

tasks/session.php
```diff
@@ -0,0 +1,169 @@
...
+            // make previous_id a unique_key. speeds up query and prevents duplicate id's
+            \DBUtil::create_index(\Config::get('session.db.table'), 'previous_id', 'previous_id', 'unique');
+
+            // return success message.
+            return \Cli::color("Success! Your session table has been created!", 'green');
+        }
+        else
+        {
+            // driver is not set to db, so inform the user.
+            return \Cli::color("Oops, your driver is currently set to ".\Config::get('session.driver').'. Please set your driver type to db before continuing.', 'red');
```
frankdejonge: Somebody could still want to do this even if db is not the default driver. A notice would be fine, but somebody should still be able to perform the task.

tasks/session.php
```diff
@@ -0,0 +1,169 @@
...
+                php oil refine session
+
+            Description:
+                The session task will create the nessecary db tables.
+
+            Examples:
+                php oil r session:create
+                php oil r session:remove
+                php oil r migrate:clear
+                php oil r migrate:help
```
frankdejonge: this should be session:clear and session:help

tasks/session.php
```diff
@@ -0,0 +1,169 @@
+<?
```
frankdejonge: The opening tag should include the 'php' part, part of our coding standards.

frankdejonge:
First of all, nice job. I think a lot of people will like this. I added some line notes here and there, minor things.

dberry37388:
does this latest commit look good or is there anything else I need to clean up?

tasks/session.php
```diff
@@ -0,0 +1,178 @@
+<?php
+/**
+ * Fuel is a fast, lightweight, community driven PHP5 framework.
+ *
+ * @package    Fuel
+ * @version    1.1
+ * @author     Fuel Development Team
+ * @license    MIT License
+ * @copyright  2010 - 2011 Fuel Development Team
```
sergiigrebeniuk: Very minor - it's 2012 now :)

tasks/session.php
```diff
@@ -0,0 +1,178 @@
...
+ * @license     MIT License
+ *
+ * Usage:
+ * php oil r session         = will prompt with menu
+ * php oil r session:create  = create the db table.
+ * php oil r session:remove  = remove the sessions table
+ * php oil r session:clear   = clear the sessions table
+ */
+
+class Session {
```
philsturgeon: We've started dropping these brackets to the line below, as we do it with functions and whatnot.

tasks/session.php
```diff
@@ -0,0 +1,178 @@
...
+        \Config::load('session', true);
+
+        if (\Config::get('session.driver') != 'db')
+        {
+            // prompt the user to confirm they want to remove the table.
+            $continue = \Cli::prompt(\Cli::color('Your current driver type is not set db. Would you like to continue and add the sessions table anyway?', 'yellow'), array('y','n'));
+
+            if ($continue === 'n')
+            {
+                return \Cli::color('Database sessions table was not created.', 'red');
```
philsturgeon, line 79: You can just throw exceptions and it will do this for you.

tasks/session.php
```diff
@@ -0,0 +1,178 @@
...
+        // create the session table using the table name from the config file
+            \DBUtil::create_table(\Config::get('session.db.table'), array(
+                'session_id'   => array('constraint' => 40, 'type' => 'varchar'),
+                'previous_id'  => array('constraint' => 40, 'type' => 'varchar'),
+                'user_agent'   => array('type' => 'text', 'null' => false),
+                'ip_hash'      => array('constraint' => 32, 'type' => 'char'),
+                'created'      => array('constraint' => 10, 'type' => 'int', 'unsigned' => true),
+                'updated'      => array('constraint' => 10, 'type' => 'int', 'unsigned' => true),
+                'payload'      => array('type' => 'longtext'),
+            ), array('session_id'), false, 'InnoDB', 'utf8');
```
philsturgeon: Is assuming InnoDB a good thing?

Also, utf8 is a config option, we dont need to specify this.

dberry37388:
I had tried to leave this as it was in the session config docs, so that the generated SQL would be the same as what was provided in the example just for consistency.

tasks/session.php
dberry37388, line 79: When I throw an exception it gives all the extra in for like ....in APPPATH/tasks/session.php on line 78..... and really do not need that in this instance. And like for example the one above is not an exception, it's just a message to the user.

philsturgeon, line 79: Exception shouldnt do that, but I guess it doesnt much matter either way.

dberry37388, line 79: the full error message it gives when outputted is:

``` php
Error - Database sessions table was not created. in APPPATH/tasks/session.php on line 78
```

I would normally use an exception for throwing an actual error or something i could try/catch, but that's not really the case here.  In this case, really just want to show the use a message.

philsturgeon:
The session config docs are an example, so setting this explicitly is not required.

tasks/session.php
dberry37388: if I am just selecting data and making a few updates here and there I generally go with MyISAM.... But when it comes to heavy selects and update requests constantly battering the db, InnoDB is the way to go. You have to take into account the table vs. row locking , crash response, etc...

PR #794, opened 2012-01-26
Offset Sum is doing twice

dbpolito:
This sum is doing again on line 274. So we don't need this sum on this method.


WanWizard:
Eh...

The first correction is for the calculated timestamp from the input of create_from_string(), the second is the GMT correction of the current time in the constuctor.

Not really related I think...

jschreuder:
Actually, this fix is right for the wrong reasons. While @WanWizard is completely correct in stating that line 274 and this one have nothing whatsoever to do with eachother, a string input date shouldn't be corrected for time. The time correction for a bad server configuration is only needed with `time()`, not with a static date that's being input.

PR #797, opened 2012-01-29
Fixes

antitoxic:
Fixed a hard-coded `Fuel\Core\*` class usage for db connection and implmeneted `list_columns()` method on the PDO connection. The implementation was done based on: http://stackoverflow.com/questions/5428262/php-pdo-get-the-columns-name-of-a-table/7091984#7091984


jschreuder:
Good work, thanks!

PR #798, opened 2012-01-29
Feature: Support for per-field error messages for validation

antitoxic:
### Messages in general

Messages can be set to the a specific field and specific rule for the field, even a specific combination of rule and rule parameters.

For instance you can set message to just `min_length` or specifically to `min_length[3]`

Set message(s) at the time of declaration of the field.

``` php
$val->add_field('username', 'Your username', 'required|min_length[5]', array(
     'required' => 'We need the username, come on',
     'min_length[5]' => 'More!'
));
```

Or via a method call afterwards:

``` php
$val->add_field('username',  'Your username', 'required|min_length[5]')
    ->set_message('min_length', 'More!');
```

Even if the validation rule is set in another manner:

``` php
$field = $val->add_field('username', 'Your username', 'required')->add_rule('min_length', 5);
$field ->set_message('min_length[5]' ,'More!');
```

If the arguments of the validation rule cannot be represented as a string that's readable to humans (like array, objects,..) than you can:
- write a validator class
- or add a named closure
### Named closure as validator is passed :

``` php
$field->add_rule(array('complex_rule'=>function($val, $arg1, $arg2..){ }));
```

So that you can set a error message for `complex_rule`

### DIFF
--- a/classes/fieldset.php
+++ b/classes/fieldset.php
@@ -257,9 +257,10 @@ class Fieldset
 	 * @param   string
 	 * @param   array
 	 * @param   array
+	 * @param   array
 	 * @return  Fieldset_Field
 	 */
-	public function add($name, $label = '', array $attributes = array(), array $rules = array())
+	public function add($name, $label = '', array $attributes = array(), array $rules = array(), array $messages = array())
 	{
 		if ($name instanceof Fieldset_Field)
 		{
@@ -305,7 +306,7 @@ class Fieldset
 			return $field;
 		}
 
-		$this->fields[$name] = new \Fieldset_Field($name, $label, $attributes, $rules, $this);
+		$this->fields[$name] = new \Fieldset_Field($name, $label, $attributes, $rules, $this, $messages);
 
 		return $this->fields[$name];
 	}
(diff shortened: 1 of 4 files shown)


jschreuder:
I don't think this is the way to go, I've been thinking about allowing more complex error messages as well but I don't think the solution for param-specific messages is to use string parsing. That's something we are planning to get rid off (the Fieldset::add_field() method will be deprecated). I think the better solution would be to add support for a closure instead of a string that gets passed the `Validation_Error` with all the necessary information to base its output on the params.

I do like the thought to allow field-and-rule-specific error messages, but I'm not sure yet about the way you added them. I'll have to think about that part.

The last part of your post I don't get as that's documented behavior: http://docs.fuelphp.com/classes/validation/errors.html#naming_rules

For now I'll close this, as it's not yet ready as a solution.

[edit: updated the link to the docs to point there instead of to this page itself]

antitoxic:
Ok, it will stay in a package until the logic that you're describing is implemented.

Don't mind the last part. I was just making clear (even for myself) how it all works.

PR #799, opened 2012-01-29
Fix: Flipped condition is_null check

antitoxic:
Fixes minor error introduced from previous pull request. 
You're very quick :)

Thank you.

### DIFF
--- a/classes/database/pdo/connection.php
+++ b/classes/database/pdo/connection.php
@@ -194,7 +194,7 @@ class Database_PDO_Connection extends \Database_Connection
 		$result = $q->fetchAll();
 		$count = 0;
 		$columns = array();
-		is_null($like) and str_replace('%', '.*', $like);
+		!is_null($like) and str_replace('%', '.*', $like);
 		foreach ($result as $row)
 		{
 			if (!is_null($like) and preg_match($like, $row['Field'])) continue;
--- a/classes/database/pdo/connection.php
+++ b/classes/database/pdo/connection.php
@@ -197,7 +197,7 @@ class Database_PDO_Connection extends \Database_Connection
 		! is_null($like) and $like = str_replace('%', '.*', $like);
 		foreach ($result as $row)
 		{
-			if (! is_null($like) and preg_match($like, $row['Field'])) continue;
+			if ( ! is_null($like) and preg_match($like, $row['Field'])) continue;
 			list($type, $length) = $this->_parse_type($row['Type']);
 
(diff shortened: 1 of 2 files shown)


classes/database/pdo/connection.php
```diff
@@ -194,7 +194,7 @@ public function list_columns($table, $like = null)
 		$result = $q->fetchAll();
 		$count = 0;
 		$columns = array();
-		is_null($like) and str_replace('%', '.*', $like);
+		!is_null($like) and str_replace('%', '.*', $like);
```
frankdejonge: You just need a space after the "!" or use "or" instead of "and".

antitoxic: Done. Added a space. :)
Wouldn't `or` cause an error because str_replace will be executed even if  `$like` equals null?

frankdejonge: It wouldn't and the line should be

```
! is_null($like) and $like = str_replace('%', '.*', $like);
```

Now the change is not set to the var.

antitoxic: Ok, sorry, that was silly :)
Done? (looking around)

classes/database/pdo/connection.php
```diff
@@ -189,23 +189,23 @@ public function list_tables($like = null)
...
+		$result  = $q->fetchAll();
+		$count   = 0;
 		$columns = array();
-		is_null($like) and str_replace('%', '.*', $like);
+		! is_null($like) and $like = str_replace('%', '.*', $like);
 		foreach ($result as $row)
 		{
-			if (!is_null($like) and preg_match($like, $row['Field'])) continue;
+			if (! is_null($like) and preg_match($like, $row['Field'])) continue;
 			list($type, $length) = $this->_parse_type($row['Type']);
```
frankdejonge, line 201: just one more space before the ! and then you're done ;) thanks!

antitoxic, line 201: OK, done after a branchly sized headache :) Do you know of an option  in PHPStorm to set spaces around `!` operator?

jschreuder:
One last thing, this is not in the style guide but a PHP best practice: don't use the `@` operator. It only hides the error but it will still have triggered one behind the scenes. As such it's "dirty" code, you should always use `isset($var) ? $var : null;` in these cases.

antitoxic:
@jschreuder Done :)

jschreuder:
Thanks!

Issue #726, opened 2011-12-13
Missing rollback to SAVEPOINT for DB transactions


billmn:
I think that can be useful get a way to set some SAVEPOINT for DB transactions.
Then, it can be simple rollback to this savepoint with something like  :

<pre>DB::rollback_transaction($db, $savepoint)</pre>

or

<pre>DB::rollback_to($savepoint)</pre>

jschreuder:
I remember looking into this earlier and running into many inconsistencies between DB engines in the way it was implemented, which made me decide against implementing it back then. It would be good to add such details to the feature request to see if it's even possible in a way that doesn't make it too engine specific.

billmn:
I don't have experience with other DB engines other than MySQL ... I don't know how I can be useful for you

WanWizard:
@FrenkyNet Can we add this on the list for QB2?

frankdejonge:
We sure can. I'll put it on my list!

WanWizard:
Ok, marked as "Accepted" for the 2.0 milestone.

frankdejonge:
This doesn't fit into the driverless design the 1.x line has. Since 2.0 will not have it's own query builder I'm closing this. It's already accounted for in Cabinet.

PR #800, opened 2012-01-29
Feature: Nested db transactions

antitoxic:
Allowing the multiple `\DB::start_transaction()` usage, so that saving of objects can be delegated to autonomous methods.
#### Reference:
- http://blog.ekini.net/2010/03/05/zend-framework-how-to-use-nested-transactions-with-zend_db-and-mysql/
- http://framework.zend.com/wiki/display/ZFPROP/Nested+Transaction+Support+for+Zend_Db_Adapter_Abstract+-+Bryce+Lohr

### DIFF
--- a/classes/database/mysql/connection.php
+++ b/classes/database/mysql/connection.php
@@ -42,6 +42,11 @@ class Database_MySQL_Connection extends \Database_Connection
 	protected $_in_transaction = false;
 
 	/**
+	 * @var  bool  Allows nested transactions
+	 */
+	protected $_transaction_level = 0;
+
+	/**
 	 * @var  string  Which kind of DB is used
 	 */
 	public $_db_type = 'mysql';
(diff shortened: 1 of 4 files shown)


WanWizard:
Not all databases that are supported by the PDO driver support SAVEPOINT LEVEL.

I feel we need to postpone this modification (or perhaps only merge it for the MySQL drivers) until the new database layer is complete, which will have per-platform drivers.

antitoxic:
@WanWizard I've introduced:

``` php
public static $workarounds = array(
   'nested_transactions' => array(
      'savepoint' => array('mysql', 'pgsql')
    )
 );
```

to the `DB` class.

Which I'm checking in the `PDO` driver so the `SAVEPOINT` is used only for the described db types.

jschreuder:
This is the same as #726

PR #801, opened 2012-01-29
Feature/override query insert values

antitoxic:
Query builder (`Database_Query_Builder_Insert`) currently only adds new sets of values when the methods `set()` and 'values()' are called.

In the case of single set of values it will be a conviniet feature to be able to update this set of values.

Now possible by passing `true` as second parameter.

### DIFF
--- a/classes/database/query/builder/insert.php
+++ b/classes/database/query/builder/insert.php
@@ -88,7 +88,7 @@ class Database_Query_Builder_Insert extends \Database_Query_Builder
 		{
 			throw new \FuelException('INSERT INTO ... SELECT statements cannot be combined with INSERT INTO ... VALUES');
 		}
-		
+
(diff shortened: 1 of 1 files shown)


WanWizard:
@FrenkyNet Any comments on this one?

frankdejonge:
If I'm correct this will affect every insert query that has a boolean value as the last param. Also the function api changes when the overwrite is in place. Normally you'd do `->values(1, 2, 3, 4)` but in the new way it would turn into `->values(array(1,2,3), true)` which is confusing. In other words, I see what the problem is you're trying to solve, but I don't think this is the way to go. This would brake BC in a small amount of cases, but in those case it'll lead to some unforseen weird behavior. One of the small API changes in 2.0 QB is converting to this type of input allowing you to overwrite.

PR #802, opened 2012-01-30
fix Form class dont_prep does not work correctly

kenjis:
- remove "dont_prep" atrribute shown in output HTML
- fix dont_prep of Form::select does not work
- fix Form::select does not prep option tag contents
- add tests


kenjis:
Is there something wrong with this pull request?

jschreuder:
Nope nothing ;) Just have been incredibly busy and hadn't had the time to look into it.

PR #803, opened 2012-01-31
str_sep for Upload class

zoe-edwards:
Allows the developer to change the default underscore to something else in the config of the upload file.

[Here’s the Docs](https://github.com/fuel/docs/pull/316) for it.


WanWizard:
Could you give it a more meaningful name then 'str_sep'? For example 'normalize_separator', which makes it more clear what it does?

zoe-edwards:
Yeah sure, was trying to keep it succinct!

PR #804, opened 2012-01-31
Curl now responds with something useful if curl succeeds but gets >=400 ...

rob-mccann:
...HTTP status

If curl performs correctly but gets, for example, a 401 from the server, there was no way of getting what went wrong just that 'something' went wrong.

If curl performs correctly, it now sets the error message and number when throwing the exception to the request body and http status respectively.

I hope this is the right implementation! It shouldn't break other people's set ups as it just adds more information.

Rob McCann

### DIFF
--- a/classes/request/driver.php
+++ b/classes/request/driver.php
@@ -3,7 +3,7 @@
 namespace Fuel\Core;
 
 class RequestException extends \HttpNotFoundException {}
-class RequestStatusException extends RequestException {}
+class RequestStatusException extends \RequestException {}
 
 abstract class Request_Driver
 {
--- a/classes/request/driver.php
+++ b/classes/request/driver.php
@@ -3,6 +3,7 @@
 namespace Fuel\Core;
 
 class RequestException extends \HttpNotFoundException {}
+class RequestStatusException extends RequestException {}
 
 abstract class Request_Driver
 {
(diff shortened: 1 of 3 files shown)


jschreuder:
Code nitpick: no space between `elseif`

Otherwise this might need some tweaking as to how it's handled returned as @WanWizard points out. The only types of `RequestException` this should produce are those caused by a 400+ or 500+ status code. I'm not sure what other reason there might be for cURL to return `false` for the body, but if there is that should indeed get its own Exception message/code/class (whichever would best fit the problem).

rob-mccann:
Agreed. Any ideas for a name for a different exception? Should it extend request exception?

Apologies if I'm misunderstanding, @jschreuder:  With this patch, it throws RequestException when cURL gets http status >=400 as before but adds some extra info.

jschreuder:
@unforeseen I should have been clearer: that part I like. But it used to be a general error for anything that went wrong, now it gives a fuller error some of the time and the old general error at other times. When splitting things like this up it should handle both equally well and not one part generally and another part specialized. Thus I was wondering what other reasons for a `false` return body there might be and if those could get a better error just like the 400/500 ones.

classes/request/driver.php
```diff
@@ -3,6 +3,7 @@
 namespace Fuel\Core;
 
 class RequestException extends \HttpNotFoundException {}
+class RequestStatusException extends RequestException {}
```
jschreuder: Always extend from global to allow extension of the base class. Without the prefix before `RequestException` it'll be taken from Fuel\Core and any extensions made in the App to that class wouldn't be available in `RequestStatusException`.

rob-mccann: Ah yes, I've now updated this.

PR #809, opened 2012-02-05
Improve Form::prep_value

kenjis:
This provides:
- specify htmlspecialchars() 2nd parameter to escape single quotes
- specify htmlspecialchars() 3rd parameter which makes more secure output. Invalid characters will be dropped

Changes:
- current fuel converts single quote to `&#39;`, but this converts it to `&#039;`
- Invalid charactors will be dropped

### DIFF
--- a/classes/form/instance.php
+++ b/classes/form/instance.php
@@ -579,8 +579,7 @@ class Form_Instance
 	 */
 	public function prep_value($value)
 	{
-		$value = htmlspecialchars($value);
-		$value = str_replace(array("'", '"'), array("&#39;", "&quot;"), $value);
+		$value = \Security::htmlentities($value);
 
 		return $value;
 	}
--- a/tests/form.php
+++ b/tests/form.php
@@ -99,6 +99,18 @@ class Test_Form extends TestCase
 	}
 
 	/**
+	* Tests Form::prep_value()
+	*
+	* @test
+	*/
+	public function test_prep_value()
+	{
+		$output = Form::prep_value('<"\'H&M\'">');
+		$expected = '&lt;&quot;&#039;H&amp;M&#039;&quot;&gt;';
+		$this->assertEquals($expected, $output);
+	}
+
+	/**
 	* Tests Form::select()
 	*
 	* test for dont_prep
@@ -120,6 +132,22 @@ class Test_Form extends TestCase
 	<option value="key_H&amp;M" style="text-indent: 0px;">val_H&amp;M</option>
 	<option value="key_&quot;&#39;&quot;" style="text-indent: 0px;">val_&quot;&#39;&quot;</option>
 </select>';
+		$this->assertEquals($expected, $output);
+	}
+
+	/**
+	* Tests Form::prep_value()
+	*
+	* test of invalid string
+	*
+	* @test
+	*/
+	public function test_prep_value_invalid_utf8()
+	{
+		// 6 byte UTF-8 string, which is invalid now
+		$utf8_string = "\xFC\x84\x80\x80\x80\x80";
+		$output = Form::prep_value($utf8_string);
+		$expected = '';
 		$this->assertEquals($expected, $output);
 	}
 }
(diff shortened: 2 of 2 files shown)


classes/form/instance.php
```diff
@@ -573,8 +573,7 @@ public function label($label, $id = null, array $attributes = array())
 	 */
 	public function prep_value($value)
 	{
-		$value = htmlspecialchars($value);
-		$value = str_replace(array("'", '"'), array("&#39;", "&quot;"), $value);
+		$value = htmlspecialchars($value, ENT_QUOTES, \Fuel::$encoding);
```
jschreuder: This should just use `\Security::htmlentities($value)` as that one does exactly what you did here without code duplication.

jschreuder:
Other than my comment good job, will merge when that's fixed.

kenjis:
Security::htmlentities() does not escape single quotations, because the 2nd parameter of htmlentities() in it is ENT_COMPAT, not ENT_QUOTES. So if we use Security::htmlentities(), the current specification of prep_value() will be changed. Some tests I wrote fails.

In my opition, there is no clear reason not to escape single quotations. The current security practice requires to escape single quotes. So it is better to change Security::htmlentities() to use  ENT_QUOTES.

brtriver:
I agree with kanjis.
We should not use ENT_COMPAT , but use ENT_QUOTES by default like Twig.
I want to know why ENT_COMPAT is default in Security::htmlentities().

jschreuder:
@kenjis using `Security::htmlentities()` doesn't change the current the current security practice. `htmlentities()` uses ENT_COMPAT by default thus your changes change how it currently works. But it's not necessary to escape single quotes as all properties are enclosed in double quotes. Thus I don't see the problem here. You kind of turn it around and state "there is no clear reason not to escape single quotations" which is wrong: you should have a reason to encode, not encode just for the hell of encoding.

@brtriver I think we use ENT_COMPAT because it's secure enough for our usage and the PHP default, but the history of the Security class file doesn't go back enough to check the original commit comment.

brtriver:
thank you for replay.

Anyone do not know whether users use single quote or double quote in HTML, so I think ENT_COMPACT is not enough.
and if use ENT_COMPACT in `htmlentities()`, it is needed to add about this in the documentation.

now:

> By default, Views use output encoding to sanitize anything you pass to them using Security::htmlentities().

... anything but single quote ...

jschreuder:
Now that's another discussion, and not one to have in this pull-request.

> > By default, Views use output encoding to sanitize anything you pass to them using Security::htmlentities().
> > ... anything but single quote ...

Actually that's not what's meant by "anything", what's meant by that is that it sanitizes every type of variable. It says nothing about how it sanitizes. But feel free to add a clarification to the docs on this and pull-request that as well.

kenjis:
@jschreuder current prep_value() escapes single quotes with `str_replace()`, but if we use `Security::htmlentities()`, it does not escape single quotes. So I said specification changes.

The reason to use ENT_QUOTES in `Security::htmlentities()` is just in case if HTML coders write single quotes in HTML tags.

jschreuder:
> @jschreuder current prep_value() escapes single quotes with str_replace(), but if we use Security::htmlentities(), it does not escape single quotes. So I said specification changes.

Ah sorry missed that, hmm that does complicate this. I'd prefer to solve this right.

> The reason to use ENT_QUOTES in Security::htmlentities() is just in case if HTML coders write single quotes in HTML tags.

As I said this is another discussion outside the scope of this pull-request. We can't change this outright without offering some BC. But I may know the answer to this: I can make it a setting in config and a possible 2nd setting for Security::htmlentities() so you can force it to use ENT_QUOTES here. That'll solve both right?

jschreuder:
Here you go: fe911cbf6eb0612b54b5fd1c0003b512845a93d4 - change it to use `Security::htmlentities($value, ENT_QUOTES);` and we're done.

Also defaults to ENT_QUOTES now but possible to overwrite in config.

kenjis:
@jschreuder Thank you for your commit.

I've updated this pull request. 

By the way, what do you mean by "BC"?

jschreuder:
BC = Backwards Compatibility

Thanks for the work & discussion.

kenjis:
I see. Thanks.

PR #814, opened 2012-02-07
Added the Validation config file and now it's loaded on Validation::_init()

ipalaus:
Added the validation config items that are used on Validation and Validation_Error and they are not defined anywhere.

### DIFF
--- a/classes/validation.php
+++ b/classes/validation.php
@@ -39,6 +39,16 @@ class Validation
 	protected static $active_field;
 
 	/**
+	 * Loads in the validation config file
+	 *
+	 * @return  void
+	 */
+	public static function _init()
+	{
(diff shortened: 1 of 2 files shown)


ipalaus:
Hello,

This can be set as invalid... I didn't seen [this intro on Validation documentation](http://docs.fuelphp.com/classes/validation/validation.html#/config):

> Optionally, the validation class can be configured through the global application configuration file, app/config/config.php. Define a section called 'validation', in which the following settings can be defined:

But I see it as a useful thing too... let the bosses decide!

Thanks

jschreuder:
We chose to leave this out to save a file load as we expect this won't be used by many. I stand by that.

Issue #789, opened 2012-01-26
Auto id with prefix


sahanh:
Fieldset_Field 

if ($form->get_config('auto_id', false) === true and $this->get_attribute('id') == '')

this code overrides auto id addition of Form_Instance.

(Form_Instance class code $attributes['id'] = $this->get_config('auto_id_prefix', 'form_').$attributes['name'];) 
This code doesn't add the auto_id_prefix and it ignores auto_id generation (which adds the prefix) because Fielsset_Field add an id (without prefix)

WanWizard:
You mean that when Fieldset_Field detects there is no ID set, and it autogenerates an ID, if doesn't add the prefix?

sahanh:
Fieldset_Field adds an id if an id is not set, but it doesn't add the prefix. However Form_Instance has a code already, which does the job. That code works if I comment out the code in Fieldset_Field.

jschreuder:
Don't have time to look into this now and may forget later, but I think I fixed this in develop...

mikebranderhorst:
In 1.1 dev:

config form.php
`auto_id = true`
`auto_id_prefix = form_`
(custom templates used)

echo \Fieldset::forge()->add_model('Model_Test')->build();
Notice the missing form_ for input...

```
    <div class="control-group">
        <label for="form_slug">slug</label>
        <div class="controls">
            <input type="text" id="slug" name="slug" value="" />
        </div>
    </div>
```

class field.php
comment line 392

```
        //  $this->set_attribute('id', $auto_id);
```

Notice the missing name for label...

```
    <div class="control-group">
        <label for="form_">slug</label>
        <div class="controls">
            <input type="text" name="slug" value="" id="form_slug" />
        </div>
    </div>
```

It seems that label needs the form_ + name set at field.php line 392
but input is set at form instance.php line 221

So when in field.php the id=name isset then in instance.php $attribute[''id'] is not empty anymore thus the prefix form_ is not prefixed.

I could not come up with a solution, sorry...

sahanh:
I think I found the issue, and submitted a commit. https://github.com/sahanz/core/commit/5617041e0ed49799b24c3f7be7e0f467140ca3f4

sahanh:
I think this time I got it right, https://github.com/fuel/core/pull/819

jschreuder:
Closed, will be handled by issue #839.

jschreuder:
This is fixed now by aa197e81bc0fe6ebaaf732747ce6160d3500229f

PR #819, opened 2012-02-07
Fix: Fieldset Auto ID Prefix

sahanh:
In regarding to issue https://github.com/fuel/core/issues/789#issuecomment-3838501

### DIFF
--- a/classes/fieldset/field.php
+++ b/classes/fieldset/field.php
@@ -389,7 +389,7 @@ class Fieldset_Field
 		if ($form->get_config('auto_id', false) === true and $this->get_attribute('id') == '')
 		{
 			$auto_id = str_replace(array('[', ']'), array('-', ''), $this->name);
-			$this->set_attribute('id', $auto_id);
+			$this->set_attribute('id', $this->fieldset()->get_config('auto_id_prefix', 'form_').$auto_id);
 		}
 
 		switch($this->type)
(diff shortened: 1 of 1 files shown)


classes/fieldset/field.php
```diff
@@ -389,7 +389,7 @@ public function build()
 		if ($form->get_config('auto_id', false) === true and $this->get_attribute('id') == '')
 		{
 			$auto_id = str_replace(array('[', ']'), array('-', ''), $this->name);
-			$this->set_attribute('id', $auto_id);
+			$this->set_attribute('id', $this->fieldset()->get_config('auto_id_prefix', 'form_').$auto_id);
```
jschreuder, line 392: The prefix should be added to the $auto_id variable definition, that's the logical place to add it and keeps the code better readable.

classes/fieldset/field.php
```diff
@@ -475,7 +475,7 @@ protected function template($build_field)
 		$form = $this->fieldset()->form();
 
 		$required_mark = $this->get_attribute('required', null) ? $form->get_config('required_mark', null) : null;
-		$label = $this->label ? $form->label($this->label, $this->get_attribute('id', null)) : '';
+		$label = $this->label ? $form->label($this->label, null, array('for' => $this->get_attribute('id', null))) : '';
```
jschreuder, line 478: What's the difference here?

sahanh, line 478: Assuming the prefix set form the config file is "form_".
First part adds the prefix to the id( set_attribute() ), but if it get passed to the label method as 2nd parameter it will become <label for="form_form_input"> because inside label method the prefix is appended again, to override that directly setting the "for" attribute.

jschreuder, line 478: Hmm, this is the type of thing that works but is actually wrong because the problem lies deeper. The label method is inconsistent in how it handles this and it shouldn't care whether the for attribute is passed as the second arg or as part of the attributes. It probably never should have added the prefix to begin with, but now it does it needs a fix that makes it consistent and keeps backwards compatibility for the most part. I have to think about this for a bit...

sahanh, line 478: I tried several other options but I agree with you. For the time being I'm using an extended Fieldset_Field class with custom build() method, however someone told me that it's only work with fieldset form generation. Looking forward to a fix. Thanks for looking in to it.

jschreuder:
Closed, will be handled by issue #839.

jschreuder:
@sahanz I ended up using your solution after all to keep BC, this will be rewritten in 2.0

PR #825, opened 2012-02-10
FQFN config are loaded twice

Savageman:
Signed-off-by: Julian Esperat esperat@novius.fr


Savageman:
The pull request has been updated with another feature (Ability to delete symlink) rather than create a new one. Sorry about that, I'm not sure how to deal with it... :s

PR #826, opened 2012-02-10
Added method to alter table.

peter-fritchley:
Method added so that tables can be altered, this mimics the alter_fields method but allows new foreign keys to be defined

### DIFF
--- a/classes/dbutil.php
+++ b/classes/dbutil.php
@@ -114,7 +114,62 @@ class DBUtil
 
 		return \DB::query($sql, \DB::UPDATE)->execute();
 	}
-
+	
+	/**
+	 * Alters a table.
+	 *
+	 * @throws	 \Database_Exception
+	 * @param    string    $table          the table name
+	 * @param    array     $fields         the fields array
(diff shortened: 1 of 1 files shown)


WanWizard:
This functionality is already present in the form of the drop_fields(), alter_fields() and modify_fields() methods.

peter-fritchley:
That's true, but there is no way of adding foreign keys to the database after the point of which the table has been created - functionality that is needed if you are using migrations to generate a database that has two tables that have foreign keys to one another.

Would my time be better spent adding a simple function to add primary keys? similar to how the create_index works?

WanWizard:
Primary keys should be added when you create the table. It's a bit pointless to do that afterwards, then your migrations are clearly wrong.

As to foreign keys, I understand your point, but would prefer to add the functionality to the existing methods, so they work similar to create_table().

peter-fritchley:
Yeah, i meant foreign keys - not primary keys. I agree that it should work in the same way as create table, but couldn't find where it would logically fit into any of the existing functions.

Issue #835, opened 2012-02-16
Rest Controller: No longer able to have parameters in URL because \Input::extension() is broken.


lukearmstrong:
I have a large existing site that uses the Rest Controller extensively to return json to be used by Javascript.

After updating the core I happened to notice I was getting a error when doing a GET request to this URL `/shop/productrest/change_colour.json/345/`

ErrorException [ Warning ]: Missing argument 1 for Shop\Controller_ProductRest::get_change_colour()

``` php
<?php

namespace Shop;

class Controller_ProductRest extends \Controller_Rest
{
    public function get_change_colour($product_colour_id)
    {
        // stuff that used to work
    }
}
```

This is because a [change](https://github.com/fuel/core/commit/ab5bcf209161addfd503eb96ff8f91e5215903bf#classes/controller/rest.php) was made to Controller_Rest to use \Input::extension() to determine response format. But this is broken if you have parameters in your URL, so it doesn't return the extension, it returns `json/345/` as the extension, and sets `/shop/productrest/change_colour` as the URL, losing the parameters!

I've come up with a temporary fix because I couldn't figure out the regex to do the same thing.

Links to follow...

lukearmstrong:
https://github.com/fuel/core/pull/836

jschreuder:
I think that part of this problem is that this isn't the correct format to pass additional params. You want to implement REST thus you should use REST means, meaning that the ID should be passed as a GET param: `/shop/productrest/change_colour.json?id=345`

lukearmstrong:
Yeah, I'll just rewrite all of my javascript and rest controller methods, that makes so much better sense than fixing a bug.

jschreuder:
You're missing my point: it's not a bug. Read up a bit on REST and you'll learn that a URL for REST shouldn't contain unique identifiers, those should be part of the input GET/POST/PUT/DELETE params.

lukearmstrong:
[What I have done isn't wrong](http://microformats.org/wiki/rest/urls), but yes, I agree your suggestion is better.

Rewriting Rest Controller methods and Javascript would be a very time consuming task.

jschreuder:
Ok, that wasn't in my book. Though there's an important difference between your URLs and the URLs in that document:

```
/people/1.json
/shop/productrest/change_colour.json/345/
```

The file extension should be at the end.

[...]

PR #836, opened 2012-02-16
Temporary Fix for #835

lukearmstrong:
https://github.com/fuel/core/issues/835

### DIFF
--- a/classes/input.php
+++ b/classes/input.php
@@ -133,8 +133,26 @@ class Input
 		// Strip the defined url suffix from the uri if needed
 		if (strpos($uri, '.') !== false)
 		{
-			static::$detected_ext = preg_replace('#(.*)\.#', '', $uri);
-			$uri = substr($uri,0,-(strlen(static::$detected_ext)+1));
+			// Find position of .
+			$dot_pos = strpos($uri, '.');
+
+			// Remove everything before and including the .
+			$ext = substr($uri, ($dot_pos + 1));
+
+			// Check if there are further parameters
+			if (strpos($ext, '/') !== false) {
+				// Find position of /
+				$slash_pos = strpos($ext, '/');
+
+				// Remove everything after and including the /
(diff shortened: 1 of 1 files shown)


jschreuder:
We don't accept temporary fixes into the core: either it fixes it or it doesn't.

lukearmstrong:
It fixes the issue, and after using it for a day on a large site that uses a lot of the functionality of Fuel I haven't encountered any problems.

It works if the URL has parameters after the extension, and it works if the URL does not have parameters after the extension.

The only reason I called it temporary is because it doesn't use Regex, and it may be possible to write the same thing in less lines than I have done if the current Regex was fixed. But as I don't know Regex that well, I opted to use a more long winded method I know would work.

Perhaps I should have had more confidence when submitting this pull request, but it's up to you if you want to disregard a working solution and continue with a broken core.

jschreuder:
I'll say 2 things to this:
1. We all have shit to do, like work and actual lives that are seriously more important to us than any volunteer work. If you don't think your fix is good enough why should I even take the time to check it out? Especially when you explain WHY it's not good enough. (I did check it out btw)
2. Check my answer to the issue you say this solves and you'll see I don't consider this a bug. Thus no broken core. Also even if it did somehow fix something in a bad way, that's just asking for trouble. I never accept bad code, unexplained code or unfinished code - even if it does seem to fix something. Fix something good or not, there're no halfway fixes because the trouble you may cause by bad/unexplained/unfinished code is often a cure that turns out worse than the disease.

lukearmstrong:
1. Well that was my mistake, I should have had more confidence in it. I'm very aware that some people think differently, they look at a problem like that and think that regular expressions should be used to solve it, as the previous person did.
2. There is a comment explaining what each line of my pull request does. It is not badly written, nor is it unfinished, but I am aware that it could have been written differently.

PR #840, opened 2012-02-17
Specify which fields are safe to "expose" in a REST Controller

danharper:
On a recent project I've been building a RESTful API atop Fuel and we found specifying which database fields to expose via the API painful (writing in multiple places etc.). So I wrote in a few helper methods to allow each of our resources to specify exactly which fields to expose, and even the ability to expose computed/dynamic fields.

As a run-down, imagine we have a Users resource with the following fields:
- id
- username
- email
- password
- location
- first_name
- last_name
- last_login_ip
- created_at
- updated_at

On our API we want to expose this data, but let's say we don't want to expose the `password`, `last_login_ip`, `created_at` or `updated_at` fields. In our Controller we can specify exactly which fields to expose with the `$_exposable` variable:

``` php
<?php
protected $_exposable = array(
    'id',
    'username',
    'email',
    'location',
    'first_name',
    'last_name',
);
```

We can optionally also expose computed/dynamic fields by specifying an `_exposable` _function_ in our Controller which returns an array with additional fields to merge in. For example, let's also expose a `name` field which combines the first and last name fields:

``` php
<?php
protected function _exposable($model)
{
    return array(
        'name' => $model->first_name.' '.$model->last_name,
    );
}
```

Exposing our specified fields is now as simple as doing something like:

``` php
<?php
public function get_index()
{
    $models = Model_User::find_all();
    $expose = Model_User::expose($models);
    return $this->response($expose, 200);
}
```

Looking for any criticisms/feedback on this; I haven't looked into whether I need to write tests, or what sort of documentation to write just yet. I've found this sort of functionality to be incredibly useful when building APIs.

### DIFF
--- a/classes/controller/rest.php
+++ b/classes/controller/rest.php
@@ -44,38 +44,18 @@ abstract class Controller_Rest extends \Controller
 		'csv' => 'application/csv'
 	);
 
+	/**
+	 * @var  array  contains a list of model fields which can be exposed on an API
+	 */
 	protected $_exposable = array();
 
+	/**
+	 * Exposable computed/dynamic model fields
+	 * 
+	 * @param  Model $model The current model to computed exposable fields on
+	 * @return array computed/dynamic fields to expose
+	 */
 	protected function _exposable($model) {}
-
-	public function expose($models)
-	{
-		$exposed = array();
-		if (is_array($models))
-		{
-			foreach ($models as $model)
-			{
-				$exposed[] = $this->_expose_one($model);
-			}
-		}
-		else {
-			$exposed[] = $this->_expose_one($models);
-		}
-		return $exposed;
-	}
-
-	protected function _expose_one($model)
-	{
-		$exposed = array();
-		foreach ($this->_exposable as $key)
-		{
-			$exposed[$key] = $model->$key;
-		}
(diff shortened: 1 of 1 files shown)


jschreuder:
This is more app logic than core logic. Probably a good topic for a tutorial on how to solve the problem, but doesn't really have a place in the core. Especially not in the REST controller as that's not necessarily tied to one single model and thus not the right place for this logic (it should probably go into a base model).

danharper:
I did originally have this in the models in our app, just wasn't too sure where it would be best to put it in the core. I do feel though that something similar to this would be very beneficial to the project, and very useful for building REST APIs.

jschreuder:
It is very usefull and I have something similar setup in my base model, but I felt then that it was App logic and didn't add it back into Fuel for that very reason. This is one way of solving it, I did another and maybe someone else will have their own approach. And there's nothing wrong with that, not everything has a one-size-fits-all solution and the framework can't solve each and every tiny problem. A framework, especially one like ours that wants to stay light, should cover the big and most common problems - but there will be parts that you'll still have to add yourself and there's nothing wrong with that either.

danharper:
Ok, no problem :)

PR #842, opened 2012-02-19
Removed $_properties from model/crud.php

arbme:
Removed $_properties as its not used with model/crud.php as far as I can see.

### DIFF
--- a/classes/model/crud.php
+++ b/classes/model/crud.php
@@ -440,11 +440,6 @@ class Model_Crud extends \Model implements \Iterator, \ArrayAccess {
 
 		$vars = $this->prep_values($vars);
 		
-		if (isset(static::$_properties))
-		{
-			$vars = \Arr::filter_keys($vars, static::$_properties);
-		}
-		
 		if(isset(static::$_updated_at))
 		{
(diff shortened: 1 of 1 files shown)


jschreuder:
You are removing the very point where it is used:

``` php
<?php
if (isset(static::$_properties))
{
    $vars = \Arr::filter_keys($vars, static::$_properties);
}
```

This filters the variables that are saved by the properties that are set, all others are ignored.

arbme:
I understand the concept but in the docs for model crud under configuration, it has no mention of $_properties so dont see why it is needed. Does ORM use this class?

jschreuder:
It removes properties that are set on the model but you don't want in the database. This allows you to set additional properties that aren't DB data. And read what I said as I pretty much explained this in my previous post.

This has nothing to do with the ORM and that something isn't in docs doesn't mean anything, especially when it's usefull functionality. All it might mean is that someone added this without documenting it.

arbme:
Thanks you for the explaination

frankdejonge:
Just for the record, the docs where updated right after this push.

PR #844, opened 2012-02-22
Validation::_empty is a static method.

ounziw:
I propose to use self::_empty instead of $this->_empty

### DIFF
--- a/classes/validation.php
+++ b/classes/validation.php
@@ -627,7 +627,7 @@ class Validation
 	 */
 	public function _validation_required($val)
 	{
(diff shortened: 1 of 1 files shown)


jschreuder:
Not necessary, you can call static methods dynamicly and that's not bad practice.

EDIT: and never ever use `self::` in core classes, they should be able to use methods from extensions thus always use `static::`. But as this happens in dynamic context `$this->` works just as well.

ounziw:
Dear,  jschreuder.

I understand, Thanks.

PR #851, opened 2012-02-28
New method in Fieldset to return values as associative array

aaaaron:
The populate method is powerful for pulling user input into a Fieldset and validating it, but I want to pull those values out to hand them to a model.


jschreuder:
The fieldset class is not meant to maintain or keep values and this would give the wrong signal. The intention is you get the data from the validation class using `$val->validated()` once the validation was run.

PR #853, opened 2012-02-28
Save raw SQL for profiler's use with EXPLAIN

jimcottrell:
Currently PHPQuickProfiler runs an EXPLAIN query for any saved SELECT queries, but it builds the EXPLAIN using the SQL that has already been escaped for output.  This is a simple patch to save the raw SQL and use that instead.

### DIFF
--- a/classes/profiler.php
+++ b/classes/profiler.php
@@ -50,6 +50,7 @@ class Profiler
 		if (static::$profiler)
 		{
 			static::$query = array(
+				'raw_sql' => $sql,
 				'sql' => \Security::htmlentities($sql),
 				'time' => static::$profiler->getMicroTime(),
 			);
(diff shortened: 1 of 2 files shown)


WanWizard:
I believe the profiler shouldn't touch the profiling data, so it should not be escaped there. That should be the task of the display code in the profiler.

This fix will also store every query twice, which will increase the memory consumption of the application.

jimcottrell:
I definitely agree with you that ideally the display code should be responsible for escaping the data.  I just went for the simplest possible change, figuring that the memory issue isn't that big of a deal because profiling is a development/testing activity, and not something for production anyway.

jimcottrell:
I'd be happy to submit a slightly more involved pull request if you'd like.

WanWizard:
Please do so, I don't have the time atm.

jschreuder:
As this one won't go in I'm closing this pull-request, you can open a new one when you modified it as agreed upon.

PR #857, opened 2012-03-02
Establish new view in classes/viewmodel.php

true2trance:
Opportunity to establish new view on one line, like $this->set_view('new_view');


jschreuder:
Not really a good idea because is signifies that this is a good idea, while the whole concept actually intends for one ViewModel to map directly to one View. While it is possible to break this, it's not the intended usage and should as such not be encouraged.

Beside that point, you can already replace the internal view with just one line: `$this->_view = \View::forge($new_view);`

true2trance:
I can not quite correctly understood the ideology of ViewModel!
Thanks for reply!

PR #858, opened 2012-03-02
Renders the field based on the tag instead of the type (when specified)

Savageman:
This is the continuation to allow `<button type="submit">`


huglester:
Just a note, it should be:
switch ( ! empty(
(according to the coding standarts)

Savageman:
Right, thanks.

PR #862, opened 2012-03-05
changed handling of constraint to cater for easier definition of enums.

PWesterdale:
Current syntax for defining an enum in an add fields statement is:

'origin' => array('type' => 'enum','constraint' => "'foo','bar'"),

I don't think this fits correctly in with the way fuel works in pretty much every other case where arguments like those would be specified in an array.
With my change you can specify as follows:

'origin' => array('type' => 'enum','constraint' => array('foo','bar')),


WanWizard:
@FrenkyNet, any comments?

frankdejonge:
I presume you've tested this? The order has changes, but that shouldn't matter as far as I know. Update this docs for this and I'll merge it. Thanks!

PWesterdale:
I've tested this and its working with no issues - I checked to make sure the order wasn't causing any issues and it doesn't appear too.
Amended the docs here:
https://github.com/fuel/docs/pull/330

PR #864, opened 2012-03-07
Add and drop foreign keys

peter-fritchley:
Added ability to add and drop foreign keys after a table has been created.

### DIFF
--- a/classes/dbutil.php
+++ b/classes/dbutil.php
@@ -373,10 +373,12 @@ class DBUtil
 	 * @param	array 	$foreign_key	a single foreign key
 	 * @return 	int		number of affected rows
 	 */
-	public static function add_foreign_key($table, $foreign_key = array()) 
+	public static function add_foreign_key($table, $foreign_key) 
 	{
-		if ( ! $foreign_key)
-			return;
+		if ( ! is_array($foreign_key))
+		{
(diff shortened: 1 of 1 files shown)


philsturgeon:
Possibly not a great idea, we're trying to phase out this odd "MySQL only" class not add new features to it.

peter-fritchley:
So what is the alternative? re-writing this class to make use of the db abstraction?

philsturgeon:
The class is being rewritten already. Just ignore it till it goes away.

Foreign keys are only supported by some engines anyhow.

Emailing from my iPhone 4S like a BOSS!

jschreuder:
We can allow it in here, we just can't promise sustained compatibility in 2.0 (other than Legacy package).

I do have a few notes though:
- It shouldn't just return on empty input, that's not valid and thus should throw an exception (InvalidArgumentException to be precise)
- Using parenthesis is required per our [coding standards](http://docs.fuelphp.com/general/coding_standards.html), thus the same `if` statement should have them.
- It must have an accompanying pull-request to the docs with documentation of these methods.

If you're quick it may be included in 1.2.

peter-fritchley:
Associated documentation pull request https://github.com/fuel/docs/pull/333

PR #874, opened 2012-03-16
Ability to delete symlinks

Savageman:
Sorry for the commit overflow. I'm not getting any better :'(
Only 1 file is modified in the end...


jschreuder:
Don't worry about the merges, that's normal and can always happen.

PR #878, opened 2012-03-20
Incorrectly named var $sorted_migrations doesn't exist it should be $mig...

ghost:
...rations


WanWizard:
Has been fixed in 21122558d63ca262f7666de2bd01c70a1d3d6cac

PR #879, opened 2012-03-20
Incorrect var $class it should be $name.

ghost:


WanWizard:
No, it shouldn't. 

Has been fixed in be7d052b67cae78c80a99648afa16f3f30d98eb4

PR #880, opened 2012-03-20
Small fixes.

antitoxic:
### URI

Fixes a problem when an extension is detected in the URI and there is no slashes in the URI besides the first one. In this case `pathinfo()` will return "\" or "/" as "dirname" .
#### Example

If we load a url like `www.example.com/my-first-page.html` we expect the URI to be `/my-first-page.html` but:
- on windows it will be `\/my-first-page.html`
- on unix based platforms it will be `//my-first-page.html`
### The other change

Just adding a getter for the `Model_Crud` `$_table_name`.

### DIFF
--- a/classes/input.php
+++ b/classes/input.php
@@ -135,7 +135,7 @@ class Input
 		if ( ! empty($uri_info['extension']))
 		{
 			static::$detected_ext = $uri_info['extension'];
-			$uri = $uri_info['dirname'].'/'.$uri_info['filename'];
+			$uri =  ltrim($uri_info['dirname'], '\\/').'/'.$uri_info['filename'];
 		}
 
 		// Do some final clean up of the uri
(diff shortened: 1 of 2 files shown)


WanWizard:
Can you send separate pull requests for separate issues?

This will allow us to accept them individually, now we are forced to accept them all (or none of them).

antitoxic:
You make it a bit harder for me than needed. Is the getter a unwanted feature?

Can you point me to an explanation how can I extract a commit to a new branch painlessly?

WanWizard:
Then you'll have to wait until someone agrees on both. I want to accept the URI fix, I don't do Model_Crud. ;)

frankdejonge:
To add to @WanWizard's message. That's the way we manage our pull request. One feature/change per pull request. So it's not "to make it harder for you than needed", it's part of the rules we set up to maintain quality of code (and project) by keeping things organized.

antitoxic:
:) @FrenkyNet  yes I know it's not on purpose. I'll try to do something about extracting the commit to a feature branch.

I hope it's this week at some point.

PR #887, opened 2012-03-26
Make partials globally accessible so they can be used in child views.

ghost:


WanWizard:
as discussed in IRC, core classes should not make something global.

In your specific situation with nested views, either use a normal template for the page layout and theme instances for the sections in this layout, or nest theme instances (assign a new theme instance to a partial instead of a view).

PR #911, opened 2012-04-06
Added change_fields method to DBUtil class. Closes #910

### LINKED ISSUE #910: DBUtil:modify_fields produces incorrect SQL when renaming columns
When renaming fields, you should use `ALTER TABLE...CHANGE` instead of `ALTER TABLE...MODIFY` which is what is currently hard-coded.

This issue originates from a change this commit: cd151a68b961f11074e461f268dfd06441140f22. However, simply reverting this commit would re-introduce the original issue that prompted the change. A possible fix would probably check to see if you're renaming the field and adjust the `alter` type appropriately.
#### Test Code

```
\DBUtil::modify_fields('members', array(
    'creation_date' => array('name' => 'created_at', 'type' => 'INT', 'default' => 0),
    'modified_date' => array('name' => 'modified_at', 'type' => 'INT', 'default' => 0),
));
```
#### Produces this SQL

```
ALTER TABLE `members` 
MODIFY `creation_date` `created_at`  INT DEFAULT '0' NOT NULL,
MODIFY `modified_date` `modified_at`  INT DEFAULT '0' NOT NULL
```

Note: This SQL would be fine if we weren't renaming the columns.
#### But Should Produce this SQL

```
ALTER TABLE `members` 
CHANGE `creation_date` `created_at`  INT DEFAULT '0' NOT NULL,
CHANGE `modified_date` `modified_at`  INT DEFAULT '0' NOT NULL
```

bunchjesse:
Signed-off-by: Jesse Bunch jesse@getbunch.com

### DIFF
--- a/classes/dbutil.php
+++ b/classes/dbutil.php
@@ -162,6 +162,20 @@ class DBUtil
 	}
 
 	/**
+	 * Changes fields in a table.  Will throw a Database_Exception if it cannot.
+	 * Use this over modify_fields when you need to change the field name.
+	 *
+	 * @throws	Fuel\Database_Exception
+	 * @param	string	$table			the table name
+	 * @param	array	$fields			the modified fields
+	 * @return	int		the number of affected
+	 */
+	public static function change_fields($table, $fields, $db = null)
+	{
+		return static::alter_fields('CHANGE', $table, $fields, $db);
+	}
+
+	/**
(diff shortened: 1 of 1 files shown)


bunchjesse:
I'm an idiot and got the wrong issue number in my commit. I'll re-submit.

PR #914, opened 2012-04-10
Asset: Prevent duplicate files for being added in a group.

alejandroiglesias:
The title pretty much says everything. I believe this is very helpful.

### DIFF
--- a/classes/asset/instance.php
+++ b/classes/asset/instance.php
@@ -457,7 +457,9 @@ class Asset_Instance
 		{
 			// Prevent duplicate files in a group.
 			if (Arr::get($this->_groups, "$group.$key.file") == $asset)
+			{
 				continue;
+			}
 
 			$this->_groups[$group][] = array(
 				'type'	=>	$type,
--- a/classes/asset/instance.php
+++ b/classes/asset/instance.php
@@ -455,6 +455,10 @@ class Asset_Instance
 
 		foreach ($assets as $key => $asset)
 		{
+			// Prevent duplicate files in a group.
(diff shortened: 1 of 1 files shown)


frankdejonge:
This is why you should always test your code before sending a pull request. Arr::path doesn't exist.

alejandroiglesias:
I'm sorry, you're right. In fact I was developing in Kohana and used Arr::path(), but the wonderful FuelPHP already has dot-notated paths for arrays in Arr::get() ! I'm still in the transition from Kohana to FuelPHP and learning the simmilarities and differences.
I just commited the fix. Best regards.

tomschlick:
Also I believe as per the fuel coding standards the "continue;" statement should be enclosed within braces.

http://fueldevdocs.exite.eu/general/coding_standards.html#/control_structures

It's not entirely explicit on not using them for one line statements but just from previous experience I believe so.

jschreuder:
What @tomschlick said: all control structures should always use braces, no exceptions.

alejandroiglesias:
Sorry about that, thanks to all for instructions. Already fixed that also.

classes/asset/instance.php
```diff
@@ -455,6 +455,12 @@ protected function _parse_assets($type, $assets, $attr, $group)
 
 		foreach ($assets as $key => $asset)
 		{
+			// Prevent duplicate files in a group.
+			if (Arr::get($this->_groups, "$group.$key.file") == $asset)
```
frankdejonge: Arr should be prefixed with a backslash to get it from global

alejandroiglesias:
@FrenkyNet thanks. Fixed.

PR #915, opened 2012-04-13
Fix for Migrate::_find_migrations when $start or $end are not of the form '00x_...'

richard-myers:
_find_migrations() can receive $start and $end as either ints, a simple number string (007) OR as a migration file name (007_create...).

This should patch the detection of the $start and $end to work when receiving the int / numeric only strings which do not contain the _'s which previously resulted in the $start/$end going to empty strings when no '_' was found

### DIFF
--- a/classes/migrate.php
+++ b/classes/migrate.php
@@ -373,12 +373,14 @@ class Migrate
 		// normalize start and end values
 		if ( ! is_null($start))
 		{
-			$start = ltrim(substr($start, 0, strpos($start, '_')), '0');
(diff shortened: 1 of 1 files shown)


WanWizard:
Thanks for reporting this.

This patch is a bit complicated.

First when the start or end contains a prefix, you strip everything after the prefix. Then, you use a substr() again, which is not needed as the value will never contain an underscore at that point (as you have stripped everything).

I've committed an alternative.

richard-myers:
Sorry. The start one should of read like the end version. I messed it up a little when trying to do the pull request on here. Hopefully I'll get it right on the next pull request. Will look out for your fix shortly.

PR #922, opened 2012-04-17
Added cache config argument to the Cache::forge calls

tarnfeld:
I noticed the `Cache::forge` calls weren't passing in any of the set `static::$config['cache']` params when they were being created, so changing the `cache->driver` option in the `agent.php` config file had no effect.

### DIFF
--- a/classes/agent.php
+++ b/classes/agent.php
@@ -362,7 +362,7 @@ class Agent
 	 */
 	protected static function get_from_browscap()
 	{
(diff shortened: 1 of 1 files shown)


westin:
This looks great! I hope they integrate this asap. Thanks man!

kenjis:
This pull request does not pass tests.

There was 1 error:

1) Fuel\Core\Test_Agent::test_browser
Fuel\Core\FuelException: No cache driver given or no default cache driver set.

There were 14 failures:

WanWizard:
Fixed now.

PR #924, opened 2012-04-19
Allow input params nested in array.

peter-fritchley:
This allows field names to be specified with array syntax e.g. foo[bar], and get these values back out using the input class.

### DIFF
--- a/classes/input.php
+++ b/classes/input.php
@@ -357,6 +357,8 @@ class Input
 		{
 			static::hydrate();
 		}
+		
+		$index = str_replace(array('[', ']'), array('.', ''), $index);
 
 		return \Arr::get(static::$input, $index, $default);
 	}


frankdejonge:
We use dot notation for all nested arrays. The way to go would just to use `foo.bar` as the param instead of `foo[bar]`. So if none of the other devs object, I'm closing this pull request.

peter-fritchley:
You can't have a dot in POST, so if the field is created like so $fieldset->add('foo.bar', 'FooBar', array()); then your solution above wont work, because by the time you try to access the POST array 'foo.bar' will have become 'foo_bar'.

frankdejonge:
Yes, but this won't fix that either. You request it by foo[bar], it's converts to foo.bar, and then dot-style retrieves that from the array, which is the same as requesting foo.bar in the first place.

peter-fritchley:
Yes, that's correct. But validation uses the key of the field name. which will mean that any required fields which have an array style name will always fail validation checks. So the problem should be addressed in either all of the core code that uses \Input::param(), or in the param function itself in order to convert the array style field names into dot style syntax.

PR #925, opened 2012-04-20
Add another value to the substitutions that template() understands. 

aaaaron:
This allows the container for fields to be identified uniquely, which can be required for doing more complex layouts with CSS.  I'd prefer to have found a descriptor for each field that didn't include form_ in the name, but the general idea works.  There is no parent selector in CSS.


WanWizard:
@jschreuder could you check this one?

WanWizard:
This can no longer be merged.

PR #927, opened 2012-04-21
Fixing BMP pull #640

huglester:
Hello,

did partialy fix pull 640.
Probably the perfect way would be to add support to BMP somehow, but we with kriasna aren't strong enough to do that :) 

If anyway wants - go on! :)

I did run these tests threw my image testing controller - everything seems fine

### DIFF
--- a/classes/image/gd.php
+++ b/classes/image/gd.php
@@ -18,7 +18,7 @@ class Image_Gd extends \Image_Driver
 {
 
 	protected $image_data = null;
-	protected $accepted_extensions = array('png', 'gif', 'jpg', 'jpeg');
+	protected $accepted_extensions = array('png', 'gif', 'jpg', 'jpeg', 'bmp');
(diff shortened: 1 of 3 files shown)


huglester:
If that's get merget, I'll update the docs too

frankdejonge:
We've talked about this, but supporting BMP formats is way out of date, and shouldn't be encouraged any longer. Closing pull-requests.

huglester:
@FrenkyNet as we talked with @WanWizard , we ended up with leaving support for imagick/imagemagick. Even I know people who often upload BMP not knowing they aren't 'INTERNET" images.

But why disable BMP support for imaging/imagemagick if they work like they should? I even made an exception, to pas images to imagick driver if the extention is BMP. And I have around 10% of such images.

I think it is not worth removing bmp support, as it works as expeceted for imagick/imagemagick drivers

Thanks

PR #930, opened 2012-04-22
Twitter Bootstrap CSS 2.0.2 Compliance

hlibco:


WanWizard:
Pull requests like this will not be accepted.

You can override the template structure in the pagination config file if you need to.

philsturgeon:
Anything that is set as Twitter Bootstrap 1.x should be updated for 2.x, but the core shoul not be changed from generic to any specific framework (if that makes sense).

hlibco:
Hi, Phil.

Thanks for the reply. I added 2 new indexes to the end of the config, it's really necessary to make te pagination more flexible, and without this, is impossible to make it compliance to twitter bootstrap.

Take a closer look at this, please. I want to help fuelPHP to become as much flexible as it can be. Well, congrats to the good job on it.

Bests,
--Felipe

WanWizard:
I don't have an issue with adding the inactive template variables, but the other template changes will break current applications that use these values. Changes should be backwards compatible.

hlibco:
Hi, Harro.

You are right about backwards compatibility, and my changes were fully compatible.

I sent another pull request, easier to see the changes, just 3 lines of changes (2 added and 1 edited).

Bests,
--Felipe

WanWizard:
I don't see an open pull request?

hlibco:
I´m sorry, I´ve closed the pull request (patch-3), now it´s open again, could you take a look please?

hlibco:
I'm sorry, I've closed it. Now it's reopened. Could you take a look please?

Bests,
--Felipe

WanWizard:
Just for the record: those pull requests were accepted.

PR #931, opened 2012-04-23
Added 2 new configs to make it more flexible for HTML markup with

hlibco:
"regular_start" and "regular_end" that embrace the regular page links (all page links, except the active one). Today, just the active page link can be customized, I added an option where people can customize the other page links.

### DIFF
--- a/classes/pagination.php
+++ b/classes/pagination.php
@@ -56,6 +56,8 @@ class Pagination
 		'next_mark'      	  => ' &raquo;',
 		'active_start'   	  => '<span class="active"> ',
 		'active_end'     	  => ' </span>',
+		'regular_start'           => '',
+		'regular_end'             => '',
 	);
 
 	/**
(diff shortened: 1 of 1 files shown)


classes/pagination.php
```diff
@@ -56,6 +56,8 @@ class Pagination
 		'next_mark'      	  => ' &raquo;',
 		'active_start'   	  => '<span class="active"> ',
 		'active_end'     	  => ' </span>',
+		'regular_start'           => '',
```
hlibco, line 59: You can add it into a list, for example, using "li" tag

classes/pagination.php
```diff
@@ -56,6 +56,8 @@ class Pagination
...
+		'regular_end'             => '',
```
hlibco, line 60: You can close the "li" tag if you would like to embrace links into a list

classes/pagination.php
```diff
@@ -207,7 +209,7 @@ public static function page_links()
 			else
 			{
 				$url = ($i == 1) ? '' : '/'.$i;
-				$pagination .= \Html::anchor(rtrim(static::$pagination_url, '/').$url, $i);
+				$pagination .= static::$template['regular_start'].\Html::anchor(rtrim(static::$pagination_url, '/').$url, $i).static::$template['regular_end'];
```
hlibco, line 212: Is very common see paginations displayed as lists, so we need this flexibility, to embrace page links into the "li" tag or another markup people want.

PR #938, opened 2012-05-01
Added Inflector::pluralize_if() method

m4tthumphrey:
Also added 'sheep' and 'antelope' to uncountable words!

### DIFF
--- a/classes/inflector.php
+++ b/classes/inflector.php
@@ -25,7 +25,8 @@ class Inflector
 
 	protected static $uncountable_words = array(
 		'equipment', 'information', 'rice', 'money',
-		'species', 'series', 'fish', 'meta'
+		'species', 'series', 'fish', 'meta', 'sheep',
+		'antelope',
 	);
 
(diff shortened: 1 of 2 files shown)


m4tthumphrey:
I had actually started to write that first, but decided against it. Feel free to rewrite it!

frankdejonge:
`$num > 1 and $word = \ Inflector::pluralize($word);` No real need for this function as far as I'm concerned. @WanWizard @jschreuder @philsturgeon ?

WanWizard:
Agree, in this form it doesn't add anything.

I assume the idea here is to mimic gettext() behaviour, where (2, 'elephant') returns '2 elephants'. I can see a need for such a feature, but if that is required, use gettext.

Problem with this is that pluralize() is very limited, and limited to English only. Try pluralizing a Russian word... ;)

frankdejonge:
Since it doesn't add anything significant and promotes misuse of the class (it's a tool for internal use, pluralizing tables and classes) I'm closing this.

PR #939, opened 2012-05-01
Set the $_POST or $put_delete based on php://input on "applicaiton/json" request

tgriesser:
When a request is sent as "application/json" (typical in backbone.js applications) rather as the standard  application/x-www-form-urlencoded or multipart/form-data, the data in the header payload should be json_decoded and used as the request input.

### DIFF
--- a/classes/input.php
+++ b/classes/input.php
@@ -43,7 +43,25 @@ class Input
 	 * @var  $put_delete  All of the put or delete vars
 	 */
 	protected static $put_delete = null;
-
+	
+	/**
+	 * Sets the $_POST or $put_delete based on php://input on "applicaiton/json" requests
+	 */
+	public static function _init()
+	{
+		if (static::server('CONTENT_TYPE') === "application/json")
+		{
+			if (static::method() === 'POST')
+			{
+				$_POST = json_decode(file_get_contents("php://input"), true);
+			}
+			else
+			{
+				static::$put_delete = json_decode(file_get_contents("php://input"), true);
+			}
+		}
+	}
+	
 	/**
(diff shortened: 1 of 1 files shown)


philsturgeon:
Why are you trying to confuse POST variables and HTTP body? POST vars are a combination of HTTP body with a specific MIME Type as you have already suggested, everything else is just HTTP doing its thing. Why should JSON be given preference? Why not XML? Or CSV? Or JPG?

Leave HTTP well alone, it's fine as it is.

tgriesser:
It seemed intuitive for the Controller_Rest, that if a put or post request comes in with a content/type of "application/json", the JSON variables could be accessed via Input::post('key') or Input::put('key'), rather than needing to emulate the structure in a urlencoded form on the client side.

And JSON was given preference because it's structure mimics php's array. I don't know how you'd easily go about putting the other structures into associative arrays.

See : 
  http://fuelphp.com/forums/topics/view/3737
  http://saaboke.com/?p=207

Not trying to step on HTTP's toes, just looking for a practical solution for something that (for me at least) is a common use case.

philsturgeon:
The two are nothing to do with each other. A body is perfectly allowed but should not be forced into the $_POST structure because that is entirely irrelevant.

tgriesser:
What about this: https://github.com/tgriesser/core/commit/53c7db3df06689a9b148076bd0078add497cae37

frankdejonge:
That url is hidden from me? Can't look at it.

frankdejonge:
Fixed differently: https://github.com/fuel/core/blob/1.1/develop/classes/input.php#L57

PR #944, opened 2012-05-05
Reverse the order asset paths are added to a theme

nedmas:
For some reason when using the theme class the fallback theme was taking precedence over the active theme. Reversing the order in which asset paths were added seems to solve the issue.

Signed-off-by: Tom Densham tomdensham@googlemail.com


WanWizard:
Good find!

PR #946, opened 2012-05-10
trigger an event to signify that fuel has finished it's initialization p...

peter-fritchley:
...rocess.

### DIFF
--- a/classes/fuel.php
+++ b/classes/fuel.php
@@ -205,6 +205,7 @@ class Fuel
 		}
 
 		static::$initialized = true;
+		\Event::trigger('initialized');
 
 		if (static::$profiling)
 		{


frankdejonge:
I can see why this is usefull. Only thing I'd opt is to go for `startup` rather than `initialized`. @WanWizard ?

peter-fritchley:
You're probably right, i called the event after the property in order to maintain consistency. Although `startup` is ambiguous as it could signify that the application is starting up, or that the application has started up, whereas it's clear when using `initialized`.

frankdejonge:
then again, `initialized` is the state the app is in, not a process name.

peter-fritchley:
It makes sense to me that the event signifies that the app is now in an `initialized` state, or am i missing the point?

WanWizard:
@FrenkyNet The issue I have is that this is a bit of an adhoc solution. I can think of many places where you would like to "inject" code and where an event would be handy. Are we going to inplement a "hooks" kind of system by implementing event calls everywhere?

The only reason for this that I can think of is some kind of "autoexec" functionality for every request, which can currently be implemented in an always load class. Which you have to make anyway otherwise you can't register the event as proposed here. So I don't see the added benefit of this single and specific event.

peter-fritchley:
When you have complex dependencies e.g modules depending on packages and visa versa, the autoexec through the auto load doesn't work great, as what you're depending on might not have been loaded yet. So without adding custom events that are triggered when the dependency is loaded (which is what we're currently doing, but gets messy), this seems like a cleaner solution.

WanWizard:
I don't really get that statement, as the only difference between the always load section of the initialisation and the call of the event you want to add is loading the routes. In both cases, the same classes are loaded (providing you load your class last in the always load list).

I am not against events perse, but I think we need to think hard about adding event hooks to the core. Not a lot of people will want to use them, so they will waist CPU cycles, just to please the odd one that needs this. This is against one of FuelPHP's primary design principles.

@jschreuder, any commits on this?

jschreuder:
@WanWizard I think that more events within Fuel aren't a bad thing. In the 2.0 repo there's a lot more events than there's currently available within 1.1, part of that is because I made logging event-based. And they're all app-based, which means there's no events outside the app and thus no event for 'Fuel initialization' (which would be Environment object initialization in 2.0)

The event class is loaded anyway and a call to check whether there are events registered shouldn't add overhead with any significance. So I think that it's acceptable, not sure about the naming though. The current events within 2.0 are:

```
app_created
request_created
request_started
controller_created
controller_started
response_created
controller_finished
request_finished
```

lukearmstrong:
> Not a lot of people will want to use them, so they will waist CPU cycles, just to please the odd one that needs this.
> @WanWizard

dafuq?

WanWizard:
Ok, I'll see if I can map those into locations in the 1.3/develop code.

WanWizard:
Implemented the following events:

```
app_created
request_created
request_started
controller_started
controller_finished
response_created
request_finished
shutdown (was already defined)
```

I've used 'app_created' to indicate FuelPHP has finished it's initialisation, and kept the name for compatibility reasons with 2.0. See https://github.com/fuel/core/commit/23f05d4f74267010828bbea8b3bb5317700fdd99

PR #951, opened 2012-05-14
1.2/develop

steadweb:
Refine 400 status codes by throwing 'correct' exception(s). 

At the moment, all 400 status codes are thrown as \RequestStatusException. If environment is set to development, all 400 statues bar 404 throw \Exception. This gives the developer more information when debugging, and not just a "Not Found" page!

### DIFF
--- a/classes/request/curl.php
+++ b/classes/request/curl.php
@@ -181,7 +181,7 @@ class Request_Curl extends \Request_Driver
 			$this->set_defaults();
 						
 			// Only show specific 400 error message with \Exception if not in production mode
-			if(Fuel::$env != 'production')
+			if(Fuel::$env != Fuel::PRODUCTION)
 			{
 				switch($this->response->status)
 				{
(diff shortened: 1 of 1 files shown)


WanWizard:
@philsturgeon any comments?

classes/request/curl.php
```diff
@@ -179,7 +179,27 @@ public function execute(array $additional_params = array())
 		elseif ($this->response->status >= 400)
 		{
 			$this->set_defaults();
-			throw new \RequestStatusException($body, $this->response->status);
+						
+			// Only show specific 400 error message with \Exception if not in production mode
+			if(Fuel::$env != 'production')
```
Daniel15: This should use the Fuel::PRODUCTION constant.

steadweb: Thanks Dan. I'll amend now.

frankdejonge:
I'm throwing my two cents in. While I understand where this is coming from, this can lead to unexpected behavior. When in production mode the server could also throw a non 404 status. For example when there is work being done on that server. I think the way to go about this, is to fetch the exception, check the status code, and re-throw (or handle differently) based on that, not based on what kind of exception is thrown.

WanWizard:
I agree that handling the exception is application specific, and should be dealt with there.

Closing this request.

Issue #952, opened 2012-05-14
Format to_csv() bug


billmn:
There is a bug with to_csv() method if the array is multidimensional and doesn't have a "0" key ... for example (an ORM result or an associative array)

<pre>
// ORM result
["3"]=> object(stdClass)#61 (10)
{
  ["name"]   => string(1) "foo"
  ["name_1"] => string(1) "6"
}
["1"]=> object(stdClass)#61 (10)
{
  ["name"]   => string(1) "foo1"
  ["name_1"] => string(1) "6"
}

// Associative array
array('one' => array('test' => 1), 'two' => array('test' => 2));
</pre>

<h3>Problems : </h3>

1- The method doesn't allow objects, only array
2- To check if array is multidimensional, the method use :

<pre>
if (is_array($data) and isset($data[0]))
{
....
</pre>

<h3>Possible solutions : </h3>

1- To allow to use objects we can add (after line 194):

<pre>
if (is_object($data) and ! $data instanceof \Iterator)
{
    $data = $this->to_array($data);
}
</pre>

2- I think that can be a better check for multidimensional arrays using:

<pre>
if (is_array($data) and \Arr::is_multi($data))
{
    $data = array_values($data);

    if (\Arr::is_assoc($data[0]))
    {
        $headings = array_keys($data[0]);
    }
    else
    {
        $headings = array_shift($data);
    }
}
</pre>

Suggestions? (I've already talked about that with @FrenkyNet on IRC)

billmn:
Now we can discuss about the code optimization

PR #954, opened 2012-05-14
Fixed issue #952 for multidimensional arrays and added separator support

billmn:


philsturgeon:
Everything looks good except for the use of `\Arr::is_multi($data)`. 

I will soon be porting this class to be part of a standalone package, meaning reliance on framework classes needs to be avoided. Is this considerably better than checking for `isset($data[0])`?

billmn:
With <code>isset($data[0])</code> there's a problem when you using an array without "0" index, like :

<pre>
array('one' => array('test' => 1), 'two' => array('test' => 2), ...)
</pre>


See https://github.com/fuel/core/issues/952

philsturgeon:
Yeah it's tough to avoid that. Ok, I'll merge this then worry about a way to make it stand alone when I get to that.

PR #956, opened 2012-05-14
cURL method param not set

phillbrown:
At the moment if you pass a method parameter into a cURL request, it will be overwritten by the parent constructor in Request_Driver. As set_method isn't overwritten in the cURL driver, the patch just passes the parameter to Request_Driver to handle.

### DIFF
--- a/classes/request/curl.php
+++ b/classes/request/curl.php
@@ -19,8 +19,6 @@ class Request_Curl extends \Request_Driver
 	 */
 	public function __construct($resource, array $options, $method = null)
 	{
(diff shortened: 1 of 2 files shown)


frankdejonge:
Can you request a pull for this on 1.2/develop (from a 1.2/develop branch), this should be included in the upcoming 1.2 release. Thanks.

phillbrown:
Sure, should be one there now.

frankdejonge:
I merged this from the 1.2/develop branch. Thanks for the pull-request!

PR #958, opened 2012-05-15
Allows to pass params to process input data in Format::forge()

billmn:
For example the delimiter of CSV string

### DIFF
--- a/classes/format.php
+++ b/classes/format.php
@@ -36,26 +36,27 @@ class Format
 	 *
 	 *     echo Format::forge(array('foo' => 'bar'))->to_xml();
 	 *
-	 * @param   mixed  general date to be converted
+	 * @param   mixed   general date to be converted
 	 * @param   string  data format the file was provided in
+	 * @param   array   parameters for processing the data by format
 	 * @return  Format
 	 */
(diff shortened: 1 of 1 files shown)


billmn:
I've closed this pull cause I think that _form_csv can be more cleaner using str_getcsv ... I'm going to rewrite this method

philsturgeon:
It's not about being clean, its about working properly. Using str_getcsv if you have multiple lines in a field it will break.

The built in PHP functions are terrible, why do you think I wrote all this code?

billmn:
I didn't know that you are wrote this method

During the implementation I saw this problem, so I've mantained your code ... I'm trying only to allow to pass parameters like delimiter, enclosures and escape char

billmn:
There's a new pull ... https://github.com/fuel/core/pull/959

philsturgeon:
Thanks, im not against improvements I am just explaining why I went to the effort of avoiding the pre-built methods. :)

billmn:
Sure, and your reasons are right! :)

The CSV method needs the support for delimiters because Excel 2007 allow to specify a delimiter on import, but not on export (only ";") !!!!

#excel fail

PR #959, opened 2012-05-15
Some improvements on Format class and Ftp mirror bug fix

billmn:
- Allows parameters to process input data (on _from_\* methods)
- Support for "delimiter" and "enclosure" on CSV
- Fixed CSV test

### DIFF
--- a/classes/model/crud.php
+++ b/classes/model/crud.php
@@ -112,7 +112,7 @@ class Model_Crud extends \Model implements \Iterator, \ArrayAccess {
 
 		if ($result !== null)
 		{
-			return current($result);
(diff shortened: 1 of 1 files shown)


frankdejonge:
Bill, could you split this in two separate pull-requests? I'm not sure about your modification to the format class, might be a bit over engineered. The other one I'll merge when in a separate pull-request. Thanks!

billmn:
Sure ...the ftp fix is here : https://github.com/fuel/core/pull/966

What can we do with the Format class improvements?
I think that the delimiter is an important option for the CSV methods

PR #985, opened 2012-05-25
Better error handling for ajax requests.

tgriesser:
If the request is ajax, the output of the error should be similar to the Cli output, rather than a rendered view, so it is easier to read in the console.

### DIFF
--- a/classes/error.php
+++ b/classes/error.php
@@ -188,7 +188,7 @@ class Error
 		{
 			if (\Input::is_ajax())
 			{
-				exit($data['severity'].' - '.$data['message'].' in '.\Fuel::clean_path($data['filepath']).' on line '.$data['error_line']);
+				echo $data['severity'].' - '.$data['message'].' in '.\Fuel::clean_path($data['filepath']).' on line '.$data['error_line'];
 			}
 			else
 			{
--- a/classes/error.php
+++ b/classes/error.php
@@ -169,7 +169,14 @@ class Error
 
 			try
 			{
-				exit(\View::forge('errors'.DS.'php_fatal_error', $data, false));
+				if (\Input::is_ajax())
+				{
+					exit($data['severity'].' - '.$data['message'].' in '.\Fuel::clean_path($data['filepath']).' on line '.$data['error_line']);
+				}
+				else
+				{
+					exit(\View::forge('errors'.DS.'php_fatal_error', $data, false));
+				}
 			}
 			catch (\FuelException $view_exception)
 			{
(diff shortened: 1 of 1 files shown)


jschreuder:
This is actually View logic and doesn't really belong there. For the CLI we made some special arrangements, but that is a distinctively different type of access while AJAX is just a normal request with an additional header. This may be very much unwanted functionality for someone who is loading in rendered HTML through AJAX. The solution here might be to add something like this to the View, but in the end it really is APP logic and belongs in your application and not in the core. Thus such a View should also go into the APP.

frankdejonge:
I actually agree with adding this logic. While view-logic should indeed have it's place in the app, this is a framework error. Therefore, not really part of the application. When getting errors on ajax requests, you get a ball of useless html to work through in order to get to the information you need. The error class renders the error in a format that aid the user, and in this case it doesn't. Even the regular PHP error messages would be easier to work through. The changes @tgriesser proposes do indeed add some sort of view logic, but in my opinion for the better, and not app logic intrusive.

WanWizard:
An alternative is to return a generic message, and refer to the log file which should contain the same error information.

jschreuder:
The problem is that this is magic behavior and has a strong potential for spagetti status:

**Magic** because it is built in behavior that can only be changed by extending the class and replacing the method. It cannot be controlled one way or another. Instead of just a generic way of working with this (an error view) there is code that makes decissions for the developer at the core level on how it will be displayed. Thus if such code would ever go into the core it should be in the View(Model) layer and not at this level.

**Spagetti** because now there is CLI handling at one point and AJAX handling at another. At some point we may get HMVC or specific REST errors for one very plausable reason or another. All of these might sound individually as good ideas but will in the end lead to many decission points within the core that are for all intends and purposes spegetti code. You might have one special exception like the CLI which is in the 1.x branch handled very different from a normal request, but if you add more exceptions to the rule you start breaking the rule.

There's nothing wrong with the idea in itself, but this is not the right place or approach to this.

jschreuder:
@WanWizard 

> An alternative is to return a generic message, and refer to the log file which should contain the same error information.

That is not really an option here as the fatal_error view gives much more information which really has no place in a log file, they'd become unreadable if we started including backtraces and such.

tgriesser:
So would a better alternative be to just create a ViewModel in the core which handles all of the error logic, rather than loading specific views from the error class?

Savageman:
Also there is the same problematic for REST requests. The application may have a generic formatting rules for the errors.

I have in mind the Twitter API as example: https://dev.twitter.com/docs/error-codes-responses which returns error message in the appropriate format (json, xml, etc.).

The problematic is even more generic. This goes beyond this pull request (maybe we should create another issue about this?), but the Profiler also could also benefits from this.
Right now, it only has an HTML output and does not behaves nicely with REST. The application may want to include profiling / debugging informations within a JSON or XML response... (same as formatting the error according to the extension).

If it's being rewritten, maybe it's good to have this in mind.

WanWizard:
This issue goes way further then only errors and ajax calls.

I've created an issue for this so it can be looked at in it's broader scope: https://github.com/fuel/core/issues/1038

PR #991, opened 2012-05-29
cache php://input to be read from both hydrate and hydrate_raw_input

tgriesser:
An error that was brought up in the irc earlier today, php://input can only be read once, and if hydrate() is called to populate `static::$input` `Input::json()` and `Input::xml()` will have no content

### DIFF
--- a/classes/input.php
+++ b/classes/input.php
@@ -50,6 +50,11 @@ class Input
 	protected static $content = null;
 
 	/**
+	 * @var  $php_input  cache for the php://input stream
+	 */
+	protected static $php_input = null;
+
+	/**
 	 * Get the request body interpreted as JSON.
 	 *
 	 * @return  array  parsed request body content.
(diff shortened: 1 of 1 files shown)


tgriesser:
Also, not sure if this should be in 1.3 or 1.2, since it's a bug that one works and not the other.

WanWizard:
i'm not to happy with this, and with some other changes that were introduced recently to provide more json/xml support.

This does the same in two different places. The static property introduced does something similar to static::$content. static::$content is determined only once. And is it cleaned for some reason (but you request raw input?), while hydrate() isn't cleaned. if you request json() first, an xml() call later will return rubbish.

I think this needs to be refactored. @jschreuder?

frankdejonge:
@WanWizard I've done the cleaning like we do with every type of input. We do it with $_GET/$_POST/etc which is also "raw input". Also in one execution the input can either be xml or json, not both, so calling the other method has no use in the first place. I don't know why the other input (PUT/DELETE) is not cleaned though.

WanWizard:
Ok, I'll try to have a look at the code later.

tgriesser:
Cleaned up and moved to pull request fuel/core#992

PR #993, opened 2012-05-30
call to set_option only keeping the newly set option and removing all previous options

richard-myers:
the set_options method of Request_Driver is replacing the current options with the new options array parsed to it but the set_option method is assuming that set_options will add the new option to the options array and not replace existing options.

For example in the Request_Soap class multiple calls are made to set_option but only the very last option to be set is being kept and no any other configuration options.

### DIFF
--- a/classes/request/driver.php
+++ b/classes/request/driver.php
@@ -152,7 +152,7 @@ abstract class Request_Driver
 	 */
 	public function set_options(array $options)
 	{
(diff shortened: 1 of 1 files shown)


WanWizard:
Fixed in https://github.com/fuel/core/commit/82d4fcb8725b299ee200035eeb375a1c39dc2c55

PR #996, opened 2012-06-02
Added Turkish translation

barisozdag:
I've added Turkish language files.

### DIFF
--- a/lang/tr/byte_units.php
+++ b/lang/tr/byte_units.php
@@ -1,0 +1,44 @@
+<?php
+
+/**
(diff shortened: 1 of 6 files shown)


jschreuder:
We decided pretty early on that we'd only include English in the official download. The reason is that it's impossible for us to maintain this and it would end up keeping back releases because we couldn't guarantee all languages are up-to-date.

We do however encourage making contributions like this public. Share them on our forums or tweet them to @fuelphp and we'll advertise. And once the Fuel Depot community site is release all of this should get a better place.

PR #1004, opened 2012-06-08
Add: Validation::get_errors()

mamor:
Returns all errors in array.

Signed-off-by: mamor mamor.dev@gmail.com
## 

I think that would be nice to have a function that returns an array.
And This commit is for https://github.com/mp-php/fuel-oil/commit/1313f841eda7166afd15cc359fdea36fde8c9153

### DIFF
--- a/classes/validation.php
+++ b/classes/validation.php
@@ -565,6 +565,36 @@ class Validation
 	}
 
 	/**
+	 * Get errors
+	 *
+	 * Returns all errors in array
+	 *
+	 * @param   array  uses keys message_prefix
+	 * @return  array
+	 */
+	public function get_errors($options = array())
+	{
+		$default = array(
+			'message_prefix' => \Config::get('validation.message_prefix', '* '),
+		);
+		$options = array_merge($default, $options);
+
+		if (empty($this->errors))
+		{
+			return array();
+		}
+
+		$array = array();
+
+		foreach($this->errors as $e)
+		{
+			$array[] = $options['message_prefix'].$e->get_message();
+		}
+
+		return $array;
+	}
+
+	/**
 	 * Alias for $this->fieldset->add()
 	 *
 	 * @return  Fieldset_Field


frankdejonge:
You can already fetch the errors by doing $val->error(); Adding a prefix to it is application logic and should be handled elsewhere. So there is no need for this function.

kenjis:
But $val->error() returns Validation Error objects. So if I want to pass them to views, I have to convert them to array.

And $val->show_errors() contains view logic, so it is difficult to use. I hope a method like this PR rather than show_errors().

frankdejonge:
@kenjis you can whitelist the Validation_Error object, when echo'd it will be the validation message. Won't that sort your problem?

jschreuder:
Or better yet: whitelist the Validation object and work with that in the View. It'll give you access to the errors ($val->error()), the validated input ($val->validated()) and the original input ($val->input()).

kenjis:
Better is sometimes worse.

To whitelist objects increases risk of XSS. If you echo $val->validated() or $val->input() directly in views, there may be XSS vulnerability.

I hope the option we can use easily FuelPHP without using objects in PHP views.

jschreuder:
Not always, depends on what you're submitting and how it's displayed. Also on what your validation rules are, if you perform XSS checks on the input your validated() output is safe. IMO the best solution is to prepare these things in ViewModels.

Though even then I just pass them like this:

```
$this->set('errors', $val->error(), false); // false switched off output encoding for this var
```

Because I know exactly where the errors are comming from and there is no XSS risk in passing them unchecked.

kenjis:
There are many kinds of programmers. Someone is excellent like you, someone is not. And human being sometimes makes a mistake.

Your code is no problem to you, because you know where the errors are coming from and there is no XSS risk. But if it is written by others, a code reviewer must check the source of the object and its validation rules to confirm no XSS. If we do not use objects in views, FuelPHP auto output encoding takes care of escaping all variables.

To pass objects is convenient for coding, but may not convenient for code review.

Issue #1005, opened 2012-06-08
Form::textarea generates incorrect textarea content


ghost:
Form::textarea generates incorrect textarea content in case value is "0"

The problem is in this check:
core/classes/form/instance.php ~433
$value = empty($attributes['value']) ? '' : $attributes['value'];

because empty("0") === true

personally I don't see why such check is necessary..

ghost:
Sorry, may be I don't understand, but why to check:
strlen($attributes['value']) == 0 and then return ''

isn't:
$attributes['value'] = (string) $value
and 
strlen($attributes['value']) == 0

already mean that $attributes['value'] === '' ?

Stelian:
The initial check was for empty(); empty(0) returns true so another check is in order to preserve the empty() check. The length of the 0 character however is 1, so adding those two conditions, that only returns true only when the variable is trully empty.

jschreuder:
I'm closing this one, further discussion can be done in the pullrequest #1007

PR #1007, opened 2012-06-08
Added strlen check to reinforce empty()

Stelian:
Fix for issue #1005

### DIFF
--- a/classes/form/instance.php
+++ b/classes/form/instance.php
@@ -473,7 +473,7 @@ class Form_Instance
 			$attributes['value'] = (string) $value;
 		}
 
-		$value = empty($attributes['value']) ? '' : $attributes['value'];
+		$value = (0 == strlen($attributes['value']) && empty($attributes['value'])) ? '' : $attributes['value'];
 		unset($attributes['value']);
 
 		if ($this->get_config('prep_value', true) && empty($attributes['dont_prep']))


classes/form/instance.php
```diff
@@ -473,7 +473,7 @@ public function textarea($field, $value = null, array $attributes = array())
 			$attributes['value'] = (string) $value;
 		}
 
-		$value = empty($attributes['value']) ? '' : $attributes['value'];
+		$value = (0 == strlen($attributes['value']) && empty($attributes['value'])) ? '' : $attributes['value'];
```
jschreuder, line 476: Your point is valid, there are however 3 problems with your solution that need to be fixed before we will merge:
- you are running empty() after the strlen() check, meaning that this code will trigger a PHP notice when the key 'value' doesn't exist in $attributes.
- you are using `&&`, but our coding standards state that `and` is preferred over it.
- the check in itself says `if 0 equals the length of $value` but in natural language and logic you'd put this the other way around, which is thus prefered: `if the lenght of $value equals 0` or in PHP: `strlen($attributes['value']) == 0`.

jschreuder, line 476: 1. And now check 5 lines before that, you can't be sure thus it needs empty() first.
2. You're right, that's an old left-over from before we standardized it. Still any new contributions need to follow the [coding standards](http://docs.fuelphp.com/general/coding_standards.html)
3. It was a binding suggestion, non negotiable. We do screen pull-requests to keep things consistent and ensure quality. I always allow people to fix it themselves so they'll get credit when we pull from them, but if it's too much trouble to work with feedback I'll just code up the fix myself in the suggested manner.

Stelian:
Please refactor the way you see fit and close this issue.

PR #1009, opened 2012-06-10
Fix validation test

billmn:


billmn:
So I've changed the code to use add_rule() as you asked me

PR #1011, opened 2012-06-11
Fixes "ErrorException [ Notice ]: Undefined index: extension" when using File_Area and file has no extension

Shne:
in COREPATH/classes/file/area.php in the get_handler function, $info['extension'] is used two places (line 89 & 95) with no check whether the index is defined. The index 'extension' is not defined if the file has no extension, as documented at http://php.net/manual/en/function.pathinfo.php

minor notice: I also noticed the line $info = pathinfo($path); is repeated (line 85 & 88).

### DIFF
--- a/classes/file/area.php
+++ b/classes/file/area.php
@@ -86,17 +86,18 @@ class File_Area
 			$info = pathinfo($path);
 
 			// check file extension
-			$info = pathinfo($path);
-			if ( ! empty($this->extensions) && ! in_array($info['extension'], $this->extensions))
-			{
-				throw new \FileAccessException('File operation not allowed: disallowed file extension.');
-			}
+			if(isset($info['extension'])) {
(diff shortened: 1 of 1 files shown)


WanWizard:
This is not the way to fix this, as you will now never be able to block file names that don't have an extension.

A better solution would be to check after the pathinfo() call if the "extension" key exists in the array, and if not, create it:

```
isset($info['extension']) or $info['extension'] = "";
```

WanWizard:
@Shne are you planning an update? 

If so, also checkout the coding style ( && should not be used ), and could you pull it against 1.2/develop, as it is a bugfix on the existing code?

WanWizard:
Dealt with in https://github.com/fuel/core/commit/43b50e91fabed4404ab10b8687369d0f140a105a

Issue #997, opened 2012-06-02
Fieldset Field Label "for" does not match Input "id"


kenjis:
How to reproduce:

``` php
        $form = Fieldset::forge('form_test');

        $ops = array('male', 'female');
        $form->add('gender', '', array(
            'options' => $ops, 'type' => 'radio', 'value' => 1
        ));
```

Generated HTML:

```
<input type="radio" value="0" id="form_gender_0" name="gender" />
<label for="form_form_gender_0">male</label><br />
<input type="radio" value="1" id="form_gender_1" name="gender" checked="checked" />
<label for="form_form_gender_1">female</label><br />
```

"id" is `form_gender_0`, but "for" is `form_form_gender_0`.

billmn:
Can you post your form.php file content to check your configuration?
Which version of fuelphp are you using?

kenjis:
@billmn 

I use fuelphp 1.2. I did not change form.php.

``` php
return array(
    'prep_value'            => true,
    'auto_id'               => true,
    'auto_id_prefix'        => 'form_',
    'form_method'           => 'post',
    'form_template'         => "\n\t\t{open}\n\t\t<table>\n{fields}\n\t\t</table>\n\t\t{close}\n",
    'fieldset_template'     => "\n\t\t<tr><td colspan=\"2\">{open}<table>\n{fields}</table></td></tr>\n\t\t{close}\n",
    'field_template'        => "\t\t<tr>\n\t\t\t<td class=\"{error_class}\">{label}{required}</td>\n\t\t\t<td class=\"{error_class}\">{field} <span>{description}</span> {error_msg}</td>\n\t\t</tr>\n",
    'multi_field_template'  => "\t\t<tr>\n\t\t\t<td class=\"{error_class}\">{group_label}{required}</td>\n\t\t\t<td class=\"{error_class}\">{fields}\n\t\t\t\t{field} {label}<br />\n{fields}<span>{description}</span>\t\t\t{error_msg}\n\t\t\t</td>\n\t\t</tr>\n",
    'error_template'        => '<span>{error_msg}</span>',
    'required_mark'         => '*',
    'inline_errors'         => false,
    'error_class'           => 'validation_error',
);
```

billmn:
Ok ... I've solved the issue and created a test

When my previous pull request (https://github.com/fuel/core/pull/1009) was closed I can create a new pull request that solve this problem

WanWizard:
The PR that fixes this was merged.

PR #1017, opened 2012-06-13
Fixed issue #997 : Label "for" does not match Input "id" in fieldsets

billmn:


WanWizard:
@jschreuder could you have a look at this when you have the time?

WanWizard:
@jschreuder @FrenkyNet any one that can check this? Fieldsets isn't my thing... ;-)

jschreuder:
I'll try and find the time. This is a recurring problem because any fix to it breaks it somewhere else. It's a conflict between the Form & Fieldset logic.

billmn:
@jschreuder do you think that can be solved?

I think that create tests also for the Form class can be useful to identify conflicts

jschreuder:
@billmn that'd be the best approach. I've attempted to solve this twice already and I'm afraid the problem is that part of the current functionality is based on buggy behavior.

WanWizard:
I see there's a similar test in the Fieldset test class, so it looks like this was fixed elsewhere.

PR #1018, opened 2012-06-13
1.2/develop - Finder prefix error

kastermester:
Finder prefixes cross module loading with a backslash. No namespaces in Autoloader is prefixed with a backslash. Removed the prefixing from Finder


WanWizard:
Can you give an example of this?

This "fix" will break the finder, as will will not find a single file in another module anymore?

kastermester:
It seems I was too hasty on this, sorry.

PR #1020, opened 2012-06-14
A wrapper for Debug::dump() called dd()

zoe-edwards:
So I was inspired by Laravel’s `dd()` function which is just a wrapper for `die(var_dump($var))`, but Fuel has a lovely function called Debug::dump() so rather than **D**ie **D**ump, I felt Fuel’s should be **D**ebug **D**ump.

Here’s the docs: https://github.com/fuel/docs/pull/372

### DIFF
--- a/base.php
+++ b/base.php
@@ -172,6 +172,21 @@ if ( ! function_exists('__'))
 }
 
 /**
+ * A wrapper function for Debug::dump()
+ * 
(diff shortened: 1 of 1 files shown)


huglester:
Nice.
+1

WanWizard:
I don't think this has a place in the core.

Shortcuts are always handy to have, but different for everyone. I personally don't have a need to dump method arguments on a regular basis. I suggest you add this to your code in app/bootstrap.php.

PR #1031, opened 2012-06-20
Remove return on parent::before() (void method)

felixgilles:
The Controller_Rest don't have the "return".
To be consistent, we have two possibilities:
- Add the "return" in the Controller_Rest
- Remove the "return" in the Controller_Hybrid

I chose the second


WanWizard:
The Controller_Hybrid is a hybrid between Controller_Template and Controller_Rest. And Controller_Template::before() does return a value and so does Controller_Hybrid to stay compatible with that.

PR #1047, opened 2012-06-30
some unittest for pagination.php

ounziw:
I made some unittest for pagination.php


WanWizard:
There's something wrong with the indentation. The method code should ident inside the braces.

ounziw:
I'm sorry for awkward indentation.
I just corrected now.

PR #1051, opened 2012-07-05
Added array-naming to select while presents multiple attribute.

SCIF:
Class Form didn't allow to post array of values because naming of field didn't check isset of multiple attribute.


WanWizard:
As discussed in IRC, this will break existing apps that address this by passing "name[]" as name. At least it should check if the array suffix isn't already present.

frankdejonge:
@SCIF any updates on this?

PR #1053, opened 2012-07-05
Allow image extension to be forced.

m4tthumphrey:
At the moment you can only pass in a filename with a valid file extension. This prevents you from passing a temporary file or an otherwise "invalid" file. This commit will allow you to pass an additional parameter to force (or override) the file extension so it is processed correctly. It can be useful when when manipulating images directly from a file upload via the <code>Upload</code> class.

### DIFF
--- a/classes/image.php
+++ b/classes/image.php
@@ -96,12 +96,13 @@ class Image
 	/**
 	 * Loads the image and checks if its compatable.
 	 *
-	 * @param   string  $filename  The file to load
+	 * @param   string  $filename							The file to load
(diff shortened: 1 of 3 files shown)


m4tthumphrey:
```
Upload::process(array(
   'path' => DOCROOT.DS.'files'
));

if (Upload::is_valid())
{
    $data = Upload::get_files(0);

    Image::load($data['file'], false, $data['extension'])
        ->crop_resize(200, 200)
        ->save('image');    
}
```

PR #1054, opened 2012-07-06
Update 1.3/develop

itcom:
If you specify the charset, SQL produces incorrect.

### DIFF
--- a/classes/dbutil.php
+++ b/classes/dbutil.php
@@ -311,7 +311,6 @@ class DBUtil
 			$sql .= \DB::quote_identifier($field);
 			$sql .= (array_key_exists('NAME', $attr) and $attr['NAME'] !== $field) ? ' '.\DB::quote_identifier($attr['NAME'], $db ? $db : static::$connection).' ' : '';
 			$sql .= array_key_exists('TYPE', $attr) ? ' '.$attr['TYPE'] : '';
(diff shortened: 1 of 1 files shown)


frankdejonge:
Hey, can you explain a little more about what's going on. What is generated, why is it wrong, what did you fix, and what is the result?

itcom:
In this way,

\DBUtil::create_table('dt_member', array(
    'id' => array('constraint' => 10, 'type' => 'bigint', 'auto_increment' => true, 'unsigned' => true),
    'uid' => array('constraint' => 255, 'type' => 'varchar', 'charset' => 'utf8_bin', 'comment' => 'unique id'),
    ), array('id'), false, 'InnoDB', 'utf8'
);

SQL like this will be generated.

CREATE TABLE `dt_member` (
`id` bigint(11) UNSIGNED NOT NULL AUTO_INCREMENT,
`uid` varchar CHARACTER SET utf8 COLLATE utf8_bin(255) NOT NULL COMMENT 'unique id',
PRIMARY KEY `id` (`id`),
) ENGINE = InnoDB  DEFAULT CHARACTER SET utf8;

itcom:
By performing this fix, SQL will generate the correct.

CREATE TABLE `dt_member` (
`id` bigint(11) UNSIGNED NOT NULL AUTO_INCREMENT,
`uid` varchar(255) CHARACTER SET utf8 COLLATE utf8_bin NOT NULL COMMENT 'unique id',
PRIMARY KEY `id` (`id`),
) ENGINE = InnoDB  DEFAULT CHARACTER SET utf8;

PR #1055, opened 2012-07-07
Controller response property

gizzmo:
It was deprecated, but not removed in 1.2 because a proper alternative wasn't found. This is my
suggested alternative.

According to the docs, the controller should make sure a Response object should be returned. Its
not the Requests job to do that.

Signed-off-by: Gizzmo justgiz@gmail.com

### DIFF
--- a/classes/request.php
+++ b/classes/request.php
@@ -74,6 +74,9 @@ class Request
 			static::$active->children[] = $request;
 		}
 
+		// fire any request created events
+		\Event::instance()->has_events('request_created') and \Event::instance()->trigger('request_created', '', 'none');
+
 		return $request;
 	}
 
@@ -328,6 +331,9 @@ class Request
 	 */
 	public function execute($method_params = null)
 	{
+		// fire any request started events
+		\Event::instance()->has_events('request_started') and \Event::instance()->trigger('request_started', '', 'none');
+
 		if (\Fuel::$profiling)
 		{
 			\Profiler::mark(__METHOD__.' Start');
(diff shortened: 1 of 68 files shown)


classes/request.php
```diff
@@ -421,17 +421,13 @@ public function execute($method_params = null)
...
-		}
-		elseif ($response instanceof \Response)
+		if ($response instanceof \Response)
 		{
 			$this->response = $response;
 		}
 		else
 		{
-			$this->response = \Response::forge($response, 200);
+			throw new \FuelException(get_class($this->controller_instance).'::'.$method.'() or the controller after() method must return a Response object.');
```
frankdejonge, line 445: This shouldn't throw an exception. Lazy responses make usage much more enjoyable :) Returning a view (or viewmodel) is valid and should be converted to a response.

gizzmo, line 445: As you can see, i just moved that exception down from the is_null check. Also controller after method makes sure a Response is returned. I was basically following what was said in the docs: http://docs.fuelphp.com/general/controllers/base.html#/returning_results

frankdejonge, line 445: Missed that, never mind, still waking up ;)  Looks good to me. We'll have to decide wether this is for 1.2.2 or for 1.3. Thanks!

jschreuder:
Judging by the comments it should have been removed for 1.2, but as it wasn't it is too big a change for a 1.2.x release. As such I think we should merge this into the 1.3/dev branch.

gizzmo:
So should close this and redo the pull on the 1.3 branch?

WanWizard:
Yes please. 

And also consider the impact on Controller_Template, which still does use $this->response internally. If it is no longer passed by the request, it need to be constructed internally.

And don't forget to update the documentation accordingly.

gizzmo:
Woops, should have closed it first :P

PR #1057, opened 2012-07-09
Autoload "Init" class after module loading

sp72:
Line 99: Try to execute \$namespace\Init::_init() after loading a module. This trick allows to initialize the module.
NB: The method _init() is called by the autoloader if class exists.

### DIFF
--- a/classes/module.php
+++ b/classes/module.php
@@ -96,6 +96,10 @@ class Module
 			$ns  => $path.'classes'.DS,
 		), true);
 
+		// try to execute $ns\Init::_init()
+		// this trick allows to initialize the module
+		\Autoloader::load($ns.'\\Init');
+
 		static::$modules[$module] = $path;
 
 		return true;


WanWizard:
If modules should get something like this, it would be better to take the bootstrap approach, like app, core and all packages. That would be more consistent.

I remember a discussion in the past wether or not to implement it. @jschreuder, do you know the reason why we decided not to?

sp72:
You are right about the bootstrap approach.
I remain convinced this is a good feature, even for a module. In addition, it's consistent with the "always_load" feature.

jschreuder:
If we would implement something like this we should take the bootstrap approach to mirror packages, we shouldn't add a random new approach.

The reasoning behind not allowing for something like this was that we felt that any module is fully depedent upon the application and should be setup by that (this is waaaay back when we were months away from the 1.0 release).

Whether that approach was right is debateable seeing how modules ended up in the 1.x branch. My problem with fixing this now is that modules are pretty much removed as a concept for 2.0, they're only left as a possibility for code structuring within a package. As such, by implementing this in the 1.x branch we'd create a compatibility problem that is pretty much unfixable.

sp72:
**jschreuder** commented

> If we would implement something like this we should take the bootstrap approach to mirror packages, we shouldn't add a random new approach.
> The reasoning behind not allowing for something like this was that we felt that any module is fully depedent upon the application and should be setup by that (this is waaaay back when we were months away from the 1.0 release).
> Whether that approach was right is debateable seeing how modules ended up in the 1.x branch. My problem with fixing this now is that modules are pretty much removed as a concept for 2.0, they're only left as a possibility for code structuring within a package. As such, by implementing this in the 1.x branch we'd create a compatibility problem that is pretty much unfixable.

Thanks for your answer and I understand your point of view.
I moved my code into a package and made some changes, it works fine with the bootstrap.

PR #1058, opened 2012-07-09
Fully remove $response property on controllers.

gizzmo:
It was deprecated, but not removed in 1.2 because a proper alternative wasn't found. This is my
suggested alternative.

According to the docs, the controller should make sure a Response object should be returned. Its
not the Requests job to do that.

Signed-off-by: Gizzmo justgiz@gmail.com

### DIFF
--- a/classes/controller.php
+++ b/classes/controller.php
@@ -21,21 +21,13 @@ abstract class Controller
 	public $request;
 
 	/**
-	 * @var  Response  The current Response object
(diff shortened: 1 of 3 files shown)


jschreuder:
Btw, this is a bit of Forward Compatibility as it is how 2.0 solves this problem :)

gizzmo:
What are the odds, I hadn't noticed.

PR #1059, opened 2012-07-10
Update 1.3/develop

itcom:
@see http://www.php.net/manual/en/function.array-merge.php

Values in the input array with numeric keys will be renumbered with incrementing keys starting from zero in the result array.


WanWizard:
Related to https://github.com/fuel/core/pull/1060#issuecomment-6876098

itcom:
Do not forget to merge?

PR #1060, opened 2012-07-10
Update 1.3/develop

itcom:
@see http://www.php.net/manual/en/function.array-merge.php

Values in the input array with numeric keys will be renumbered with incrementing keys starting from zero in the result array.

### DIFF
--- a/config/cache.php
+++ b/config/cache.php
@@ -57,7 +57,7 @@ return array(
 	'memcached'  => array(
 		'cache_id'  => 'fuel',  // unique id to distinquish fuel cache items from others stored on the same server(s)
 		'servers'   => array(   // array of servers and portnumbers that run the memcached service
-			array('host' => '127.0.0.1', 'port' => 11211, 'weight' => 100)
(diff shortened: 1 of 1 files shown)


WanWizard:
We are aware of that array-merge does. 

What I would like to know is what issue this fix addresses? The memcached storage driver doesn't do anything with the keys in that array, so wether or not they get renumbered is not really relevant.

itcom:
As an example,
If you have set as follows in app\config\cache.php,

return array(
    'memcached'     =>  array(
        'servers'   =>  array(
            array('host' => '192.168.0.1', 'port' => 11211, 'weight' => 100)
        ),
    ),
);

At run time, the contents of the `servers`,

'servers'   =>  array(
    0 => array('host' => '127.0.0.1', 'port' => 11211, 'weight' => 100),
    1 => array('host' => '192.168.0.1', 'port' => 11211, 'weight' => 100)
),

If you have become, memcached is not running locally,
Did not work properly.

WanWizard:
Ah, I see it get's loaded with the overwrite parameter, which causes array_merge() to be used instead of Arr::merge() which doesn't have this issue (but has others).

PR #1081, opened 2012-07-27
- Added ability to overwrite the options in the field completely

ajvwhite:
When using the ORM and the fieldset generation based on the Model, I
have found sometimes it is necessary to overwrite a select field's
options after being initialised by an init function completely
(different load conditions for the select perhaps based on a user action).

This small tweak will allow that, but the default behaviour stays the
same.

### DIFF
--- a/classes/fieldset/field.php
+++ b/classes/fieldset/field.php
@@ -336,7 +336,7 @@ class Fieldset_Field
 	 *
 	 * @param   string|array  one option value, or multiple value=>label pairs in an array
 	 * @param   string
-	 * @param   bool            Whether or not to replace the current optiions
+	 * @param   bool            Whether or not to replace the current options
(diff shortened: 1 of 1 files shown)


classes/fieldset/field.php
```diff
@@ -361,7 +362,7 @@ public function set_options($value, $label = null)
 			}
 		};
 
-		empty($this->options) ? $this->options = $value : $merge($this->options, $value, $merge);
+		($replace_options || empty($this->options)) ? $this->options = $value : $merge($this->options, $value, $merge);
```
huglester: should be $replace_options or empty($this->options)
according to coding standards

ajvwhite: Wasn't sure if that preference was an absolute requirement, but thank you for pointing it out, has been amended to match to the stipulated standards

PR #1083, opened 2012-07-29
Bugfix: If return code of display.php is other than LF, script error occ...

mamor:
...urs in browser.

Signed-off-by: mamor mamor.dev@gmail.com
## 

In windows, it may be set to CRLF by setup of Git.
It is hard to discover, I think this problem as it is better to correct.

http://fuelphp.com/forums/topics/view/9132

### DIFF
--- a/vendor/phpquickprofiler/display.php
+++ b/vendor/phpquickprofiler/display.php
@@ -17,7 +17,7 @@
 
 function displayPqp($output) {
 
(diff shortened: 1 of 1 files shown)


WanWizard:
The FuelPHP coding standards state that ALL files should use unix-style LF termination. 

This is therefore not a bugfix but a hack to get away with a standards deviation. The proper fix is to make sure your Git setup is correct.

PR #1084, opened 2012-07-30
Made file class exceptions more verbose.

huglester:
Made file class exceptions more verbose. Added dir/file paths to exceptions.


tomschlick:
+1, I've had to add those in a few times myself as a test when creating multiple files.

PR #1089, opened 2012-07-31
Prevent an exception when trying to delete a cache that does not exists.

andreoav:
Fix an exception when deleting a cache file that does not exists.


WanWizard:
This fix is against 1.2, it should be against 1.3/develop. I'll merge it.

PR #1091, opened 2012-08-01
Added is_json() and is_serialized() methods in Str class

billmn:


huglester:
Was missing these methods :)

+1

tomschlick:
+1 on these, I'm sure it will help get this in faster if you link the docs pull request as well

billmn:
@tomschlick Yeah ... and this is the docs :
https://github.com/fuel/docs/pull/392

classes/str.php
```diff
@@ -355,6 +355,41 @@ public static function tr($string, $array = array())
...
+	/**
+	 * Check if a string is json encoded
+	 * 
+	 * @param  string $string string to check
+	 * @return bool
+	 */
+	public static function is_json($string)
+	{
+		json_decode($string);
+		return json_last_error() == JSON_ERROR_NONE;
```
frankdejonge: == should be ===

frankdejonge:
The rest looks good to me. I thought maybe this would something for the format classes, but it actually makes sense having them in the Str class since that's what we're handling.

classes/str.php
billmn: Added ... thanks ;P

huglester:
gogo @FrenkyNet , merge this in!

Issue #1096, opened 2012-08-05
Can't use Log::write() as documented


kenjis:
Documentation says:

```
// and finally, we create a log entry with a custom $level
Log::write('Link', 'More info on http://fuelphp.com/')
```

But it does not work.

The first param is int and must be set in `$loglabels`.

The documentation is right or wrong? I hope the function as decumented.

kenjis:
I sent Pull Request: https://github.com/fuel/core/pull/1097

WanWizard:
Pull request merged.

PR #1097, opened 2012-08-05
Fix Log::write() which can't write custom level log

kenjis:
See: https://github.com/fuel/core/issues/1096


tomschlick:
Thanks for fixing this! Ran into this the other day but I didn't have time over the weekend to put in a proper pull request... now I dont have to! :)

PR #1098, opened 2012-08-06
Added condition on find_file when path is absolute

sdrdis:
I think we should check if the file exists even if its path is absolute


WanWizard:
This can be a lot simpler as the Finder will already check on absolute filenames. I'll code up an alternative.

PR #1100, opened 2012-08-06
fix to only strip the url suffix from the uri if it's defined (url_suffix in config.php) and equal

ewya:
fix to only strip the url suffix from the uri if it's defined (url_suffix in config.php) and equal

### DIFF
--- a/classes/input.php
+++ b/classes/input.php
@@ -178,7 +178,7 @@ class Input
 
 		// Strip the defined url suffix from the uri if needed
 		$uri_info = pathinfo($uri);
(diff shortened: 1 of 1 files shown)


frankdejonge:
I'm pretty sure this will break the REST controller's inner workings, right @philsturgeon?

philsturgeon:
Correct, the URI suffix here is not about the normal "add this suffix to random page links" but is meant to be variable. It could ne .json, .xml, etc.

ewya:
Ok, understood. In consequence if the last segment is a parameter it must not contain a dot (the point in my case).

PR #1104, opened 2012-08-14
Added a TextileParser

josegonzalez:
This should balance out the existing Markdown Fuel class, allowing users the flexibility to choose between the two most popular and powerful markup languages.


dhrrgn:
Nah, we're good.  Thanks tho.

PR #1106, opened 2012-08-18
Added loading of env specific module configs from main config dir

iturgeon:
Adds loading of environment specific configurations for modules in to the main app config directory.

This will allow all of the environment/project specific configuration to be in the same location instead of partially buried inside each module's own configuration folder.

Added to 1.3/develop instead of 1.2/develop as requested (functional change not a bug fix)

There is also a pull request for updates to the documentation related to this addition inbound

### DIFF
--- a/classes/config/file.php
+++ b/classes/config/file.php
@@ -117,6 +117,7 @@ abstract class Config_File implements Config_Interface
 	 */
 	protected function find_file()
 	{
+		// search for configs in packages,modules,and the main app
 		$paths = \Finder::search('config', $this->file, $this->ext, true);
 
 		// absolute path requested?
@@ -127,6 +128,12 @@ abstract class Config_File implements Config_Interface
 		}
 
 		$paths = array_merge(\Finder::search('config/'.\Fuel::$env, $this->file, $this->ext, true), $paths);
+
+		// for modules, also look for environment configs in app/config/env/modules/mymodule/
+		if ($pos = strripos($this->file, '::'))
+		{
+			$paths = array_merge(\Finder::search('config/'.\Fuel::$env.'/modules/'.substr($this->file, 0,$pos), substr($this->file, $pos + 2), $this->ext, true), $paths);
+		}
 
 		if (count($paths) > 0)
 		{


iturgeon:
Related doc pull request: https://github.com/fuel/docs/pull/394

WanWizard:
This doesn't work if you simply do

```
\Load::config('blah');
```

in a module controller, but only if you explicitly load it from a module

```
\Load::config('module::blah');
```

This is not really consistent.

WanWizard:
Ahum... Talking out of my ...

Looking at this again, the description is confusing. There is already support for environments when loading config files from a module. You won't want to achieve that here, you want module configs in app/config.

And I'm not sure that is a good idea. Module stuff should be contained in the module, so why not use

```
/modules/mymodule/config/env/myconfig
```

?

WanWizard:
@FrenkyNet @jschreuder Any comments on this?

frankdejonge:
Yeah I'm not fond of having this doubled, Also naming conflicts can occur, leading to unexpected behaiour. Module configs belong inside the module, not in the app.

WanWizard:
Closed as incorrect framework usage.

iturgeon:
I can see your point and perfectly fine with killing this, but I'll explain where this came from before it dies forever:

This comes from in the differences between Packages and Modules. 

# Packages

Default configs in package. Overrides in main app.  **Environment overrides in main app**.

```
/fuel/packages/stuff/config/stuff.php
/fuel/app/config/stuff.php
/fuel/app/config/development/stuff.php
```

# Modules

Default configs in module, **Environment overrides in Module**.

```
/fuel/app/modules/stuff/config/stuff.php
/fuel/app/modules/stuff/config/development/stuff.php
```

So package env overrides occur in the main app config directory, and Module env overrides occur in their own directories. It feels a bit of a mess, especially when you are using both.

So my desire was to locate all environment configs to the main app directory.

WanWizard:
The flow should be:
- module env config overloads module config
- module config overloads app env config
- app env config overloads app config

When a package is loaded, it's seen as a core extension so package configs are thrown into the mix:
- module env config overloads module config
- module config overloads package env config
- package env config overloads package config
- package config overloads app env config
- app env config overloads app config

If you say it doesn't work that we then that needs to be fixed.

PR #1116, opened 2012-08-23
Disables dump or inspect in production environment

younes0:
The diff is buggy and I don't understand why... Check this one instead: https://gist.github.com/3434809


frankdejonge:
The diff is blurry because you've got different line-endings. Can you fix that?

younes0:
Okay I'll check that

PR #1117, opened 2012-08-23
Disables dump and inspect in production mode

younes0:
We might forget some Debug::dump() and do not want to see the outputs in production mode.

### DIFF
--- a/classes/debug.php
+++ b/classes/debug.php
@@ -49,7 +49,7 @@ class Debug
 				var_dump($arg);
 			}
 		}
-		else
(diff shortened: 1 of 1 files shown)


tomschlick:
Just my opinion but I dont believe that dump statements are something that should be left in production code. Plus when you actually do need to dump something on production (cant reproduce in development) you wouldn't be able to with this change.

WanWizard:
I agree with @tomschlick, it's the developers responsibility to make sure the code released to production is clean and tested. (besides, a lot of devs use var_dump(), which this will not protect against).

PR #1118, opened 2012-08-23
added Input::param_valid($key, $array)

younes0:
Taken from Flourish:
http://flourishlib.com/docs/fRequest#RestrictingValidValues

The method param_valid() simplifies the process a great deal by taking
exactly two parameters, $key, the key to request the value for and
$valid_values, an array of permissible values. If the value is not one
of the valid value, the first valid value will be picked. Here is an
example:

<?php
// $action will get the value 'list' if no value was present
$action = Input::param_valid('action', array('list', 'search'));
?>

### DIFF
--- a/classes/input.php
+++ b/classes/input.php
@@ -424,6 +424,25 @@ class Input
 	}
 
 	/**
+	 * Fetch an item from either the GET, POST, PUT or DELETE array restricting to a specific set of values
+	 * 
+	 * @param  string $key           The index key
+	 * @param  array  $valid_values  The array of values that are permissible, if one is not selected, picks first
+	 * @return mixed  The value
+	 */
+	public static function param_valid($key, $valid_values)
+	{
+		settype($valid_values, 'array');
+		$valid_values = array_merge(array_unique($valid_values));
+		$value = static::param($key);
+		if (!in_array($value, $valid_values)) 
+		{
+			return $valid_values[0];
+		}
+		return $value;
+	}
+
+	/**
 	 * Hydrates the input array
 	 *
 	 * @return  void


tomschlick:
Whats the use case for it just picking the first $valid_value if the key does not exist? I would think if the key does not exist it should just return false because it could not find a valid value.

frankdejonge:
I can see a usecase for this. Sort of like an enum restriction, but then in code. I just don't think this is the right implementation and naming of such a feature, nor am I sure this really has a place in the core. And if it was implemented I'd like to see a restriction on which input array is fetched from. Any thoughts @WanWizard @jschreuder @philsturgeon @dhorrigan ?

younes0:
It might be not have its place in the core but it should be added to Fuel since it's a very common usecase.

```
 Class Controller_Inbox{
      {
           $status  = Input::param_valid('status', array('all', 'unread', 'unresponded', 'hidden'));
           $orderby = Input::param_valid('orderby', array('datetime', 'name'));
           $rpp = Input::param_valid('rpp', array(10, 50, 100)); // results per page
      }
 }
```

if no get parameters : default values are set: status=all, orderby=datetime, rpp=10
if rpp is equal to 10000 which is not a whitelisted value, then the default value is set (10)

WanWizard:
In my opinion it's very application specific. 

Input should not concern itself with parameter validation, and given the fact that you'd have to define everything in the controller anyway it's easily replaced by

```
$status = Input::param('status', false);
$status = in_array($status, array('all', 'unread', 'unresponded', 'hidden')) ? $status : 'all';
```

Or by a helper method that does this. I can understand the use case, but that is by far not restricted to Input parameters. I can come up with lots of scenario's where you would want to do something like this, not related to Input.

So in this state -1.

tomschlick:
@FrenkyNet oh i definitely can see a use case for it, i would just rather the developer provide a $default like we do in Arr::get instead of just picking the first item.

younes0:
Okay I understand, thanks for your explanations, I'll close the pull then.

tomschlick:
basically

``` php

public static function param_valid($key, $valid_values, $default = null)
{
    settype($valid_values, 'array');
    $valid_values = array_merge(array_unique($valid_values));
    $value = static::param($key);
    if (!in_array($value, $valid_values)) 
    {
        return $default;
    }
    return $value;
}

```

tomschlick:
Also I dont think this whole line is needed as it wouldn't matter if a whitelisted param was in the allowed multiple times as this method is just checking for the existence of it.

``` php
$valid_values = array_merge(array_unique($valid_values));
```

philsturgeon:
-1

Emailing from my iPhone 4S like a BOSS!

PR #1120, opened 2012-08-24
Don't cache Config defaults

### LINKED ISSUE #1119: \Config::get() doesn't return the correct default.
Now that config has been changed to cache config items in a static property, the default parameter only ever works the first time you get a particular item.

The problem can be reproduced easily with the following code:

``` php
$foo = \Config::get('does.not.exist', 'foo');
$bar = \Config::get('does.not.exist', 'bar');

var_dump($foo, $bar);
```

The problem is caused because the result of the first call to \Arr::get() is cached, no matter whether the item existed in the array or the default was used.

andrewwheal:
Fixes #1119

The default value does not want to be saved in `static::$itemcache`, the best way I can think to do this is detect when `null` (or any other value that is fairly uncommon) is returned by the `Arr::get()` and return the default without caching it.

I also thought about using the actual default in the `Arr::get()` and checking the value returned against `$default`, but imo this is more likely to stop an actual value from being cached than checking null.

### DIFF
--- a/classes/config.php
+++ b/classes/config.php
@@ -170,9 +170,9 @@ class Config
 		}
 		elseif ( ! isset(static::$itemcache[$item]))
 		{
-			$val = \Fuel::value(\Arr::get(static::$items, $item, null));
+			$val = \Fuel::value(\Arr::get(static::$items, $item, hex2bin('DEADBEEFCAFE')));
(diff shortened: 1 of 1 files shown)


tomschlick:
+1

PR #1121, opened 2012-08-24
Made non-custom URLs routing case-sensitive

neemzy:
I have modified the `parse_segments` function in the `Fuel\Core\Router` class so it checks each segment of the URL to route. If one of these segments isn't all lowercase, the function will return `false` which will result in a 404.

This additional constraint on non-custom URLs will be triggered depending on the current value of the `routing.case_sensitive` config parameter, which is involved in the similar behavior for custom (routed) URLs.

The aim of this modification is to avoid duplicate content (having multiple URLs targetting the exact same page) for SEO.

### DIFF
--- a/classes/router.php
+++ b/classes/router.php
@@ -257,9 +257,13 @@ class Router
 	protected static function parse_segments($segments, $namespace = '', $module = false)
 	{
 		$temp_segments = $segments;
+        $case_sensitive = \Config::get('routing.case_sensitive', true);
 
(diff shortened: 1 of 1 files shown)


WanWizard:
@jschreuder @FrenkyNet To me this sounds application specific, not default behavior I would like to see in the core.

frankdejonge:
This is app specific. And also, the config setting that is refered to is for controller/action routing and shouldn't interfere with possible action arguments. It's up to the dev himself to validate those arguments.

Issue #1125, opened 2012-08-28
[Feature] Log Enhancements


dbpolito:
I think that would be a great addition to the framework some enhancements to log:
1. Changes Log to be Driver-Based
   - Makes possible to have logs stored on file or database simply changing on config (similar to session).
2. Makes easier to understand custom levels.
3. Creates an Event (log_called for example) so we can easily use that for example to send an e-mail if we get error log.

Let me know what do you think about this so i can send pull requests.

tomschlick:
+1 from me. Logging can definitely be improved via drivers. Code it up!

kenjis:
+1
Driver is Good!

bproctor:
+1
Great ideas! Sounds good to me

dbpolito:
I'm closing this issue as we can continue this on the pull request above.

PR #1126, opened 2012-08-29
Log Enhancements

dbpolito:
Pull Request related to #1125

I think that would be a great addition to the framework some enhancements to log:
1. Changes Log to be Driver-Based
   - Makes possible to have logs stored on file or database simply changing on config (similar to session).
2. Makes easier to understand custom levels.
3. Creates an Event (log_called for example) so we can easily use that for example to send an e-mail if we get error log.

**Important Note:** This change will not make any application crash, only will be need a heads up about configuration because it moves from config.php to a log.php file.

This is an initial commit so i can get feedback and see if it's getting good.

Still pending:
1. Makes an Oil Task for create the logs table as we have on session.
2. Change Docs
3. Event for Log Calls

Let me know your thoughts.

### DIFF
--- a/base.php
+++ b/base.php
@@ -36,35 +36,6 @@ if ( ! function_exists('logger'))
 {
 	function logger($level, $msg, $method = null)
 	{
-		// defined default error labels
-		static $labels = array(
-			1  => 'Error',
-			2  => 'Warning',
-			3  => 'Debug',
-			4  => 'Info',
-		);
-
-		// get the levels defined to be logged
-		$loglabels = \Config::get('log_threshold');
-
-		// bail out if we don't need logging at all
-		if ($loglabels == \Fuel::L_NONE)
-		{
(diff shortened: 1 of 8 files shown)


tomschlick:
+1

tomschlick:
My only suggestion for this would be to add the ability to change the location / filename of the file based logs. For instance I prefer to have logs created for each hour instead of each day so i may want to format mine like LOGPATH/2012/08/23/02.php

WanWizard:
@dbpolito any proogress? If so, can you re-submit against 1.4/develop?

@tomschlick's suggestion could be implemented by allowing strftime() tokens in the "log_path" config key, and expand that when it's used to create the correct foldername/filename prefix.

dbpolito:
I think we can continue this one in the new pull request. Thanks.

PR #1128, opened 2012-08-31
Update classes/database/pdo/connection.php

peter-fritchley:
Avoid Error:A non well formed numeric value encountered by checking whether the error code is numeric.


lukearmstrong:
Getting it too, and can confirm the proposed fix works.

lukearmstrong:
This commit, 3 days ago, introduced this bug.

49ba30a

lukearmstrong:
This commit, fixed the bug elsewhere in the file, using the same fix.
5ad870e

PR #1129, opened 2012-08-31
MySQL / MySQLi client compression

it-can:
MySQL / MySQLi client compression added


it-can:
I will add a check so it only is called when mysql is in use, ok?

WanWizard:
Yeah, that will do.

PR #1133, opened 2012-09-04
Model_CRUD : Set empty values to NULL

younes0:
I noticed that the Model_CRUD was not updating the database field when an empty value was set().

```
$my_model_crud->set(array( 
    'name'      => Input::param('name'), 
    'birthdate'     => Input::param('birthdate'),  
    'language'      => Input::param('language'),  // Value is empty ($_POST['language'] exists)
))->save();
```

This commit set empty properties to NULL.

### DIFF
--- a/classes/model/crud.php
+++ b/classes/model/crud.php
@@ -420,6 +420,11 @@ class Model_Crud extends \Model implements \Iterator, \ArrayAccess {
 
 		$vars = $this->to_array();
 
+		// Set empty values to NULL
+		foreach ($vars as &$var) {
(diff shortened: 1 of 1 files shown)


frankdejonge:
This is app-specific. Values van be an empty string, you can use the `prep_values` method to set the fields to `null` when required by your app.

younes0:
Okay, thanks.

```
// set null if empty
protected function prep_values($data)
{
    foreach ($data as &$var) {
        empty($var) and $var = null;
    }
    return $data;
}
```

PR #1135, opened 2012-09-04
Added check if path key exists while iterating

jromero:
If a path with position lower than any other in iteration, it will throw a notice due to non-existing key.

### DIFF
--- a/classes/finder.php
+++ b/classes/finder.php
@@ -149,8 +149,8 @@ class Finder
 	public function remove_path($path)
 	{
 		for ($i = 0, $count = count($this->paths); $i < $count; $i++)
-		{
-			if ($this->paths[$i] === $path)
+		{			
+			if (array_key_exists($i, $this->paths) && $this->paths[$i] === $path)
 			{
 				unset($this->paths[$i]);
 				break;


classes/finder.php
```diff
@@ -149,8 +149,8 @@ public function add_path($paths, $pos = null)
 	public function remove_path($path)
 	{
 		for ($i = 0, $count = count($this->paths); $i < $count; $i++)
-		{
-			if ($this->paths[$i] === $path)
+		{			
+			if (array_key_exists($i, $this->paths) && $this->paths[$i] === $path)
```
frankdejonge, line 153: The only reason why this would happen is when a path is removed before it and the array numeration doesn't match $i no more right?

jromero:
Yes, I have a lazy importer class that uses the following class and the notice is thrown on the second iteration.

```
    foreach ($imports as $base_path=>$files) {
        $finder->add_path($base_path);

        foreach ($files as $file) {
            if (is_dir($base_path . $file)) {
                foreach ($finder->list_files($file) as $dir_file) {
                    self::_do_import($dir_file, $method);
                }
            } else {
                $file = $base_path . $file;
                self::_do_import($file, $method);
            }
        }

        $finder->remove_path($base_path);
    }
```

classes/finder.php
WanWizard, line 153: Should become a foreach then instead of a for. slower, but probably faster then reordering the array keys after removal or a path. Also, using && is not valid as per our coding standards.

PR #1136, opened 2012-09-04
Adding support for Themes to be Module Aware

dbpolito:
Making Theme Module Aware we can keep all views on the modules and only extend the views we need to change in certain theme.

Example:

If use it on a module:

```
$data['content'] = Theme::instance()->view('index'); // with $module_aware = true in config
```

or

```
\Theme::instance()->set_partial('content', 'index', true, true);
```

It will do the following search:

```
themes/<theme>/<module>/index
modules/<module>/views/index
```

Let me know your thoughts.

### DIFF
--- a/classes/theme.php
+++ b/classes/theme.php
@@ -105,6 +105,7 @@ class Theme
 		'view_ext' => '.html',
 		'require_info_file' => false,
 		'info_file_name' => 'theme.info.php',
+		'module_aware' => false,
 	);
 
 	/**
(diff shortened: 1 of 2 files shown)


WanWizard:
This looks like a simpler solution is possible. 

As we're in feature freeze at the moment pending the release of 1.3, there is some time to think about it.

WanWizard:
Looking at this I noticed another issue. This might work if your only using HMVC requests to modules, but goes horribly wrong otherwise, where you'll have a mix of prefixed and non-prefixed view paths (for example all global templates).

dbpolito:
@WanWizard Yes... because of this i added a param for that so we can force a view to be module aware or not if we need... In this way we can mess things up...

dbpolito:
The solution would be a way to not need this automatic fallback on global views https://github.com/fuel/core/commit/851f1f66eafa165d807922263a60926078776fee#L0R713 and have a way to get a global view if we want.

WanWizard:
I don't want to go that route, I don't want dependencies between View and Theme, or changes to View because Theme needs them. This way it's all handled inside the theme class.

Assuming module aware is default, you can still load a global template like this

```
$theme->use_modules(false)->set_partial('layout', global/template')
$theme->use_modules();
```

If you insist on avoiding the second file check. But in a modular application, you shouldn't have more then a few theme files that are global, so I don't think this is an issue.

dbpolito:
Yeah, seeing in that way, sounds good... Thanks! :D

PR #1138, opened 2012-09-05
Update classes/database/pdo/connection.php

gatodomato:
Manual: http://www.php.net/manual/en/pdo.errorinfo.php - PHP 5 >= 5.1.0
Inclusion of method 'error_info()' to support the method 'PDO::errorInfo()' that returns an array of information about the last error that occurred operation performed by the database.


WanWizard:
@FrenkyNet shouldn't this method also be in the other drivers, to keep the API consistent? Now the app will fail if this method is used and one switches from PDO to MySQLi.

frankdejonge:
Indeed, the mysql driver should use mysql_errno and mysql_error, and the mysqli driver should use the _connection->error.

Issue #694, opened 2011-11-28
Fieldset In Fieldset Template Generation problem


CoolGoose:
If you have a fieldset into another fieldset the error template for the field from the build method doesn't get correctly populated since the validation class returns that field to be invalid after the fieldset processing is complete.

dhrrgn:
Can you provide the code to reproduce this please?

CoolGoose:
Form: http://scrp.at/aPX
Config: http://scrp.at/aPY

dhrrgn:
Thanks will take a look soon.  Unless @jschreuder knows exactly what the issue is.

jschreuder:
@CoolGoose You've shown how you generate the form, but without the important part: how do you show the errors, an example of your output and what it should have been. Probably there's something wrong in a method that should collect from all the sub-fieldsets as well and doesn't. Without the info it's hard to replicate, especially if I don't know what exactly is wrong ("correctly populated" is a very subjective non-specific issue)

CoolGoose:
The thing is that it's not showing the error class on the input fields from the subfieldsets.

It validates the inputs properly, but the template used for generating the form is being called before the actual validation happens, so it doesn't show the 'error' class on the div for the subfieldsets.

jschreuder:
> but the template used for generating the form is being called before the actual validation happens

I'm not sure how that's possible... the form field html is only generated when it must, at the very last moment when you echo it. That should be well after the validation took place.

From a quick look through the code it looks like the problem is that each Fieldset instance has its own Validation instance. Validation was performed by the Validation instance belonging to the parent Fieldset, but when the Fieldset_Field instance is being build it takes its error status from the Validation instance that belonged to the child Fieldset to which the Field belongs - which isn't the same Validation instance that determined the error. Thus no error is detected and the error class isn't assigned.

CoolGoose:
That's pretty much it.

Do or can I overwrite each validation instance to only use the main one or ? :)

jschreuder:
I don't think so, each Fieldset instance is a tripod with Form & Validation objects allowing specific settings per fieldset. Part of the problem is that Validation wasn't originally written with nested fieldsets in mind. And I'm not sure if the current solution is even the most optimal one, this needs some further looking into before it can be fully solved.

CoolGoose:
Gotcha ;) .

Good to know it's somewhat of a bug :P . On track for 1.2 / 2.0 ? :D

jschreuder:
[...]

PR #1140, opened 2012-09-06
Re: Issue #694 Fieldset In Fieldset Template Generation problem.

initializard:
Re: Issue #694 Fieldset In Fieldset Template Generation problem. When validation is run, if a field has a fieldset, add the field's error to the fieldset's errors also. This allows for nested fieldsets to display errors.

### DIFF
--- a/classes/validation.php
+++ b/classes/validation.php
@@ -372,6 +372,11 @@ class Validation
 			catch (Validation_Error $v)
 			{
 				$this->errors[$field->name] = $v;
+				
+				if($field->fieldset())
+				{
+					$field->fieldset()->Validation()->add_error($field->name, $v);
+				}
 			}
 		}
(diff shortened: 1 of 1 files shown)


WanWizard:
Why is add_error() protected? If it's not intended to be called from the outside, given the few lines of code you could add this to the run() method and save the method call.

initializard:
add_error() is not run on $this in run(). It's on the field's fieldset's validation and needs to access that validation's errors array which is protected.

Does that make sense?

WanWizard:
What I mean is that on L#378 you already know that name and error are valid. So you can add it to $this->errors there, instead of making a method call, which are expensive in PHP.

edit: wait, different validation instance. forget it, I need some sleep... ;)

PR #1142, opened 2012-09-07
Model_Crud accepts now composite primary keys

younes0:
The Model_Crud accepts a composite primary keys with this commit:

```
 protected static $_primary_key = array('user_id', 'friend_id');
```

So the model crud will correctly saves, updates and deletes tables with multiple/composite primary keys.
$_primary_key and find_by_pk accepts an array (too) now :

```
 \MyModel::find_by_pk(array(
      'user_id'  => 10,
      'friend_id' => 14,
 ));
```

See todos: I didn't figure out how to returns ids since:
- I don't know if it should returns an array  
- I use postgresql which causes model_crud/orm not returning ids anyway.

### DIFF
--- a/classes/model/crud.php
+++ b/classes/model/crud.php
@@ -83,7 +83,14 @@ class Model_Crud extends \Model implements \Iterator, \ArrayAccess {
 	 */
 	public static function find_by_pk($value)
 	{
-		return static::find_one_by(static::primary_key(), $value);
+		if (is_array(static::primary_key())) 
+		{
+			return static::find_one_by($value);
+		} 
+		else
+		{
+			return static::find_one_by(static::primary_key(), $value);
+		}
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


WanWizard:
@FrenkyNet can you have a look?

classes/model/crud.php
```diff
@@ -363,9 +384,28 @@ protected static function post_find($result)
...
+				else
+				{
+					$this->is_new(true);
+					break;
+				}
+			}
+		}
+		else
+		{
+			if (isset($this->{static::primary_key()}))
```
frankdejonge, line 405: This if can be replace the else as an elseif, no need for double nesting.

younes0, line 405: You're right. How can I fix my pull request without making a new one ? (sorry I don't use git often)

WanWizard:
You've got to make a new one anyway ;-)

This pull request is against 1.3/develop, but it needs to be against 1.4/develop, now that 1.3 is released.

younes0:
okay : #1156

PR #1147, opened 2012-09-13
Added is not set condition on lines[language] for Lang::load()

ghost:


it-can:
this is a duplicate of #1146

PR #1152, opened 2012-09-20
Fix auto_rename file checking logic

kenjis:
There is a bug if there are uploadings with the same filename at the same time, uploaded files may be overwritten even if set auto_rename true.

Fix it by creating file with fopen 'x' and ensure not to be overwritten.

### DIFF
--- a/classes/upload.php
+++ b/classes/upload.php
@@ -605,9 +605,20 @@ class Upload
 					do
 					{
 						$save_as[3] = '_'.++$counter;
-						$fp = @fopen($path.implode('', $save_as), 'x');
 					}
-					while ($fp === false);
+					while (file_exists($path.implode('', $save_as)));
+
+					// create file to ensure that there is no file after checking above
+					$fp = @fopen($path.implode('', $save_as), 'x');
+					if ($fp === false)
+					{
+						do
+						{
+							$save_as[3] = '_'.++$counter;
+							$fp = @fopen($path.implode('', $save_as), 'x');
+						}
+						while ($fp === false);
+					}
 				}
 				else
 				{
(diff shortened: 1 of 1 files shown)


WanWizard:
I understand the issue, but I'm not that fond of the proposed solution. Too many error suppressions (which are very expensive).

kenjis:
I understand your comment. Yes, the solution may have too many error suppressions.
I'll fix it.

WanWizard:
Still it comes across as complex. 

I think all is needed is after the "while (file_exists)", create a dummy file with the determined name, so on the next loop the file will exist and the next available name is used. To that effect, this should be enough:

```
touch($path.implode('', $save_as));
```

Possible issue is that a "before" callback can alter the filename, leaving you with an orphaned file.

WanWizard:
I've been going through the code again, and I don't see how uploading the same file twice can cause an overwrite.

Files are processed one at the time in a loop. So if auto_rename is in effect, when the second file is processed the first is picked up by file_exists() so a new name is used. If auto_rename is not used, and overwrite is not set, you get a "duplicate file" error on the second file.

WanWizard:
@kenjis is this still an issue?

kenjis:
file_exists() and file creation are not atomic, and there is no locking or something. So there is a possibilities that a file is overwritten.

For example, two processes run file_exists() at the same time (or before the first process create the file), both will get `false`.

But the probability someone meets the bug is very low. And I have no idea to improve my patch, which is complex.

WanWizard:
Ah, ok. I'll leave this open for the moment.

WanWizard:
i've implemented the solution with a temporary file in the 2.0 Upload package.

Still not completely atomic (there is a little time between the last file_exists() check and creation of the placeholder file, but better than it was.

PR #1153, opened 2012-09-20
Make Hybrid Controller's rest/template switching logic easily customizable.

regexident:
Take advantage of the Hollywood principle in order to easily provide custom switching logic for hybrid controllers.

Example:

If one wanted to enable RESTful responses for ajax calls as well as all requests ending with any of the known format extensions (`…/foo/bar.json`) one could simply override `shall_return_restful()` in the subclass:

```
public function shall_return_restful()
{
    if (array_key_exists(\Input::extension(), $this->_supported_formats))
    {
        return true;
    }
    return parent::shall_return_restful(); // or return false for full override
}
```

### DIFF
--- a/classes/controller/hybrid.php
+++ b/classes/controller/hybrid.php
@@ -34,7 +34,7 @@ abstract class Controller_Hybrid extends \Controller_Rest
 	 */
 	public function before()
 	{
-		if ( ! $this->shall_return_restful())
+		if ( ! $this->is_restful())
 		{
 			if ( ! empty($this->template) and is_string($this->template))
 			{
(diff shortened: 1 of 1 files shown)


frankdejonge:
Nice addition to the hybrid controller, I'd just rename the function to `is_restfull`...[php is american-english, not british-english](http://www.addedbytes.com/blog/if-php-were-british/) ;-)

regexident:
Yeah, use whatever naming you want.

frankdejonge:
@regexident can you update the pull-request? Then it'll get merged.

PR #1155, opened 2012-09-22
Use '.' as directory separator in view file path

ronan-gloo:
Add Arr::get() like syntax support for view paths. Ie:

`View::forge('folder.subfolder.file')`

### DIFF
--- a/classes/uri.php
+++ b/classes/uri.php
@@ -34,6 +34,17 @@ class Uri
 	{
 		if ($request = \Request::active())
 		{
+			switch ($segment)
+			{
+				case 'first':
+				$segment = 0;
+				break;
+				
+				case 'last':
+				$segment = count($request->uri->segments());
+				break;
+			}
+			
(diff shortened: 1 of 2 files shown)


WanWizard:
What's the point of using dots? Doesn't this do exactly the same?

```
View::forge('folder/subfolder/file')
```

WanWizard:
It will also break the Parser package, as using

```
View::forge('folder.subfolder.file.tpl')
```

will be converted to "folder/subfolder/file/tpl", which causes the smarty or twig template not to be found.

frankdejonge:
I think this was to mimic Laravel's view behaviour? And indeed not bc, adds unneeded/unwanted string processing, and Parser difficulties.

ronan-gloo:
I didn't know about parser, so you're right WanWizard...

FrenkyNet:
Not especially, but i through that sharing same path logic for Lang, Config and View can be usefull if based on controller / action names detection.

PR #1156, opened 2012-09-27
Model_Crud accepts now composite primary keys (fixed)

younes0:
#1142

### DIFF
--- a/classes/model/crud.php
+++ b/classes/model/crud.php
@@ -83,7 +83,14 @@ class Model_Crud extends \Model implements \Iterator, \ArrayAccess {
 	 */
 	public static function find_by_pk($value)
 	{
-		return static::find_one_by(static::primary_key(), $value);
+		if (is_array(static::primary_key())) 
+		{
+			return static::find_one_by($value);
+		} 
+		else
+		{
+			return static::find_one_by(static::primary_key(), $value);
+		}
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


classes/model/crud.php
```diff
@@ -483,12 +524,31 @@ public function save($validate = true)
 			if ($result[1] > 0)
 			{
 				// workaround for PDO connections not returning the insert_id
-				if ($result[0] === false and isset($vars[static::primary_key()]))
+				if ($result[0] === false)
 				{
-					$result[0] = $vars[static::primary_key()];
+					if (is_array(static::primary_key()))
+					{	
+						// TODO
```
frankdejonge, line 531: This is what you should find out when you're implementing this feature ;-)

younes0, line 531: Sure, as said in #1142, I didn't figure out how to returns ids since:
-   I don't know if it should returns an array
-   I use postgresql which causes model_crud/orm not returning ids anyway.

I'll install mysql and investigate this, but in some weeks.
In the mean time, I would gladly follow your suggestions to improve this feature.

WanWizard:
Can't merge anything that still has a TODO... 

In the case of MySQL (and most other RDBMS's that have auto increment features) you can only use that on a single column, and that column MUST be the primary key.

So in the case of a compound key, you're not going to have an insert id anyway, I think $result[0] should be an array containing the PK values, which could be retrieved from $var.

WanWizard:
Reviewing this PR, I'm not sure it's the correct solution. It is perhaps more logical to make the primary key property always an array, that may contain one or more column names. 

This is the way ORM does it, it keeps the code cleaner and free of a lot of checking and if statements.

@FrenkyNet your opinion?

younes0:
+1 , to avoid breaking changes, a string would be still allowed in the constructor, then put in the static::primary_key() array, right?

WanWizard:
Correct, the PK should always be stored as an array, so you can have one piece of code working with it, no matter how many columns make up the PK.

younes0:
#1187
did what you suggested and removed those TODOs.

PR #1167, opened 2012-10-12
Disable path cache on config file reload

andrewwheal:
This fixes a problem whereby when you reload a config file (e.g. `\Config::load('db', true, true)`) it was still looking in a cached list of paths, whereas more paths had subsequently been added to Finder since that config file was initially loaded.

### DIFF
--- a/classes/config.php
+++ b/classes/config.php
@@ -94,7 +94,7 @@ class Config
 		{
 			try
 			{
-				$config = $file->load($overwrite);
+				$config = $file->load($overwrite, ! $reload);
 			}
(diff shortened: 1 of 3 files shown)


WanWizard:
You also need to change the interface definition of the load() method.

andrewwheal:
@FrenkyNet pointed that out on IRC but imo this is not needed as custom cache drivers may not implement a cache

WanWizard:
@SicoAnimal can you update this pull request? It can't be merged anymore.

PR #1168, opened 2012-10-15
Add post method option to Pagination.

omoon:
In this change, you will be able to make post-method pagination links.

I add two properties 'http_method' and 'form_id'.

If you want to make post-method pagination links with the form whose id is 'form1', you can make it as below:

```
$config = array(
    'pagination_url' => 'pagination_url',
    'http_method'    => 'post',
    'form_id'        => 'form1',
);
Pagination::set_config($config);
```

You can get markup links like..

```
<a onclick="var form = document.getElementById('form1'); form.action = 'pagination_url'; form.method = 'POST'; form.submit(); return false;" href="javascript:void(0)">1</a>...
```

To avoid repetition, I also make a new static method 'regular_link' which returns a 'regular' link. This method is used in 'page_links', 'next_link', and 'previous_link' methods.

Signed-off-by: omoon omooon@gmail.com

### DIFF
--- a/classes/pagination.php
+++ b/classes/pagination.php
@@ -87,6 +87,16 @@ class Pagination
 	protected static $pagination_url;
 
 	/**
+	 * @var	string The http method
+	 */
+	public static $http_method = 'get';
+
+	/**
+	 * @var	string The form id (used when static::$http_method is 'post') 
+	 */
+	public static $form_id = 'form';
+
+	/**
 	 * Init
 	 *
 	 * Loads in the config and sets the variables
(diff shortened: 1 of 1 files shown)


WanWizard:
Why would you need a POST option, pagination is a GET operation used on a table list, it has no relation with a form?

frankdejonge:
Also this is javascript behaviour for html output in a PHP framework. I'm not liking this at all.

kenjis:
Pagination should be GET method.
In stead of POST, I hope Pagination can use query string like ?page=1.

omoon:
I use PEAR Pager in some of my applications, and it has POST mode.

http://pear.php.net/manual/en/package.html.pager.factory.php

My applications sometimes receive very huge query, so I choose POST mode in order to handle such requests.

PR #1170, opened 2012-10-17
Print string if translation of it did't found.

SCIF:
It allows don't duplicate translating if in views use one of site languages.

### DIFF
--- a/classes/lang.php
+++ b/classes/lang.php
@@ -193,6 +193,7 @@ class Lang
 			$language = reset($languages);
 		}
 
+		$default = ($default) ? $default : $line;
 		return isset(static::$lines[$language]) ? \Str::tr(\Fuel::value(\Arr::get(static::$lines[$language], $line, $default)), $params) : $default;
 	}
 


WanWizard:
I'm not to happy with the permanent character of this change, it's new behaviour and might break all code that expects null to be returned on a 'not-found' situation.

I would prefer to fix this using a callback or a separate method (that you can overload) so it happens in userspace (and you can do whatever you want).

@FrenkyNet, your thoughts?

frankdejonge:
@WanWizard Agree with you on this. This is one of the few conventions we have within FuelPHP. Changing this will most likely break a good number of apps.  Also, this is why we made the extending of the framework so easy ;-)

WanWizard:
FYI, I use this in some projects:

`````` php
class Lang extends \Fuel\Core\Lang
{
    /**
     * Returns a (dot notated) language string
     *
     * @param   string       $line      key for the line
     * @param   array        $params    array of params to str_replace
     * @param   mixed        $default   default value to return
     * @param   string|null  $language  name of the language to get, null for the configurated language
     * @return  mixed                   either the line or default when not found
     */
    public static function get($line, array $params = array(), $default = null, $language = null)
    {
        $default === null and $default = '<span style="color:000;background-color:yellow;font-weight:bold;padding:5px;">language key missing: '.$line.'</span>';

        return parent::get($line, $params, $default, $language);
    }
}```
``````

frankdejonge:
As @WanWizard shows above, you CAN in fact put "dirty" code in extension ;-)

PR #1174, opened 2012-10-24
optgroup nested indent label

offthewall:
optgroup nested indent label


frankdejonge:
This is app-specific and hacky, optgroups have a natural styling by the browser and this will change the layout of existing apps unexpectedly. If you really need this, extend the class and use that.

offthewall:
Okay, thanks.

WanWizard:
Besides, it should not be needed, there's already a text-indent style on it for that purpose.

Issue #1175, opened 2012-10-25
Hybrid controller doesn't work with get_ and post_ method names


bloveless:
Basically I couldn't get the get and post HTTP verb prefixes working. After a little working I changed the /fuel/core/controller/hybrid.php controllers routing function to return the parent::route function and it works now. I'm pretty sure I followed the docs correctly, but I couldn't get it to work correctly.

The following is what I changed the router function to. Note: commented out the throw new \HttpNotFoundException() and added the return parent::router($resource, $arguements);

```
public function router($resource, array $arguments)
    {
        // if this is an ajax call
        if (\Input::is_ajax())
        {
            // have the Controller_Rest router deal with it
            return parent::router($resource, $arguments);
        }

        // check if the action method exists
        if (method_exists($this, 'action_'.$resource))
        {
            // if so, call the action
            return call_user_func_array(array($this, 'action_'.$resource), $arguments);
        }

        return parent::router($resource, $arguments);

        // if not, we got ourselfs a genuine 404!
        // throw new \HttpNotFoundException();
    }
```

frankdejonge:
@bloveless is this for ajax or non-ajax requests?

bloveless:
Non ajax requests. I actually just submitted a pull request for this. Figured that would be easier.

frankdejonge:
@WanWizard wasn't it designed not to do this?

bloveless:
@FrenkyNet If it was designed not to do this then what is the Hybrid controller for? Just purely for ajax requests? Or is it there to merge the Template and Rest Controller functions together?

WanWizard:
It's only there to allow you to add a few REST enabled methods to a template controller.

The main issue is that the return value of a REST method is different from one made to deal with a browser request. If you allow every method to be called by both mechanisms, you need to deal with the different result types. To be able to do this, the Hybrid controller needs more then a few fixes.

bloveless:
Would I be able to extend a rest controller in my application and copy the constructor from the template controller into that controller without too many issues? To essentially create the hybrid controller that I'd like to have?

PR #1176, opened 2012-10-25
Update the hybrid controller to fall back to the rest controller for rou...

bloveless:
This fix allows the hybrid controller to fall back to the rest controller for routing. I found this because the get_ and post_ prefix don't work when using the hybrid controller. This allows them to work.


WanWizard:
See the discussion in #1175

PR #1178, opened 2012-10-29
Log Enhancements

dbpolito:
Related to https://github.com/fuel/core/pull/1126

I think that would be a great addition to the framework some enhancements to log:
1. Changes Log to be Driver-Based
   - Makes possible to have logs stored on file or database simply changing on config (similar to session).
2. Makes easier to understand custom levels.
3. Creates an Event (log_called for example) so we can easily use that for example to send an e-mail if we get error log.

**Important Note:** This change will not make any application crash, only will be need a heads up about configuration because it moves from config.php to a log.php file.

This is an initial commit so i can get feedback and see if it's getting good.

Still pending:
1. Makes an Oil Task for create the logs table as we have on session.
2. Change Docs

Let me know your thoughts.

### DIFF
--- a/base.php
+++ b/base.php
@@ -36,35 +36,6 @@ if ( ! function_exists('logger'))
 {
 	function logger($level, $msg, $method = null)
 	{
-		// defined default error labels
-		static $labels = array(
-			1  => 'Error',
-			2  => 'Warning',
-			3  => 'Debug',
-			4  => 'Info',
-		);
-
-		// get the levels defined to be logged
-		$loglabels = \Config::get('log_threshold');
-
-		// bail out if we don't need logging at all
-		if ($loglabels == \Fuel::L_NONE)
-		{
-			return false;
-		}
-
-		// if it's not an array, assume it's an "up to" level
-		if ( ! is_array($loglabels))
-		{
-			$loglabels = array_keys(array_slice($labels, 0, $loglabels, true));
-		}
-
-		// do we need to log the message with this level?
-		if ( ! in_array($level, $loglabels))
-		{
-			return false;
-		}
-
 		! class_exists('Fuel\\Core\\Log') and import('log');
 		! class_exists('Log') and class_alias('Fuel\\Core\\Log', 'Log');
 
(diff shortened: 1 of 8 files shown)


dbpolito:
Working on Docs and Oil Task, i'm going to pull it very soon.

WanWizard:
If the levels are no longer a constant, then perhaps something needs to happen to the static methods currently being used (I might not have a level called 'warning', and I might have a level called 'something' for which no method exists).

Perhaps they can be replaced by a __callStatic() that checks the existance of the level?

WanWizard:
@dbpolito I'm planning a 1.4 release, do you think you can wrap this up in the next few days?

dbpolito:
@WanWizard Yes i can... I think until monday i get this done... loved the __callStatic idea, I'm implementing it too.

WanWizard:
Cool, thanks. I'll plan the 1.4 release for sometime next week.

WanWizard:
We had a short discussion about this within the team. We might switch to monolog as our logging backend when we have integrated composer support. 

Nothing has been decided yet, but given this I don't think we should include this change in 1.4.

tomschlick:
Monolog would be a great option but are there plans for composer by default in 1.x? or just 2.0?

WanWizard:
There are plans to add composer to 1.5, as one of the steps to "ease" the transition to a new architecture.

tomschlick:
Awesome.

dbpolito:
@WanWizard So it's not going to be merged right? Should we close this one? If you change your mind, i can finish that up.

WanWizard:
It's entirely up to you. Current plan is to introduce Monolog in 1.5 or 1.6.

dbpolito:
Thinking about this if the plan is moving to Monolog i don't think it's a good idea implement this, it's going to change the same thing in a small portion of time.

I can do it if you want but i don't think it's going to worth it.

Let me know your thoughts.

dbpolito:
@WanWizard Thoughts?

WanWizard:
Monolog will be introduced in 1.5, so I'll close this.

PR #1179, opened 2012-10-29
Added ::is_xml() method to Str class

billmn:


billmn:
Done. Added also libxml dependency information in the documentation.

frankdejonge:
Could you use the correct way to skip the test using anotations? Like this:

```
/**
 * Test description
 *
 * @requires extension libxml
 */
```

billmn:
@FrenkyNet Sure ... you're right, my bad

frankdejonge:
@billmn One last thing. You'll need to retrieve the old value of libxml_use_internal_errors before overwriting it, and set it back after the check. So it'll be like this:

``` php
$internal_errors = libxml_use_internal_errors();
libxml_use_internal_errors(true);
$result = simplexml_load_string($string) !== false;
libxml_use_internal_errors($internal_errors);

return $result;
```

We don't want to interfere with global settings the user might have set.

frankdejonge:
@billmn Thanks!

billmn:
@FrenkyNet thank you! There's also the documentation to update https://github.com/fuel/docs/pull/442

PR #1181, opened 2012-10-31
Fix misjudgment of Response object if users write \Fuel\Core\Response in...

kenjis:
... controllers

I don't know this is a bug or not.

If you write like this in your controller

```
public function action_image($id = null, $name = null)
{
    …
    $body = file_get_contents($path);
    return \Fuel\Core\Response::forge($body, 200, array('Content-Type' => 
'image/png'));
}
```

`if ( ! $response instanceof \Response)` will return true, because there is no \Response class. And Status code and Content-Type will be gone, because of next line `$response = \Response::forge($response);`

This PR fixes it.

### DIFF
--- a/classes/controller.php
+++ b/classes/controller.php
@@ -41,7 +41,7 @@ abstract class Controller
 	public function after($response)
 	{
 		// Make sure the $response is a Response object
-		if ( ! $response instanceof \Response)
+		if ( ! $response instanceof Response)
 		{
 			$response = \Response::forge($response);
 		}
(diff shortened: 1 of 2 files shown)


WanWizard:
Not sure we should fix this, because it can promote bad habits. You should NEVER use \Fuel\Core, except when you're extending a core class. @FrenkyNet, your opinion?

frankdejonge:
@kenjis We've had a discussion about this with the team. It's going in, thanks!

PR #1182, opened 2012-10-31
add test case

kousuke:
Signed-off-by: kousuke kousuke.kikuchi@gmail.com

added unit test

### DIFF
--- a/tests/date.php
+++ b/tests/date.php
@@ -49,7 +49,6 @@ class Test_Date extends TestCase
 	public function test_days_in_month_0_exception()
 	{
 		$output = Date::days_in_month(0);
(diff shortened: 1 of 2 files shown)


frankdejonge:
Could you reformat the exception testing to use annotations?

``` php
/**
 * @expectedException InvalidArgumentException
 */
```

kousuke:
this is what you want?

frankdejonge:
@kousuke you can even lieve the ->fail calls out, the annotations now suggest what is expected, so if the exception is not thrown it will fail. No need to explicitly state that is has failed.

kousuke:
OK, I removed the codes.

kousuke:
Great, Thanks :)

PR #1184, opened 2012-11-02
Update classes/session/driver.php

quetzyg:
The latest commit to the classes/session/driver.php introduced a bug that prevented flash data from being erased from the session once it was used. This small change fixes it.


WanWizard:
Ah, chips, didn't finish that one. Thanks for catching it!

PR #1187, opened 2012-11-04
composite primary keys support

younes0:
#1142

### DIFF
--- a/classes/model/crud.php
+++ b/classes/model/crud.php
@@ -325,13 +325,7 @@ class Model_Crud extends \Model implements \Iterator, \ArrayAccess, \Serializabl
 	 */
 	protected static function primary_key()
 	{
-		// Convert primary key from string to array 
-		if (is_string(static::$_primary_key)) 
-		{
-			static::$_primary_key = array(static::$_primary_key);
-		}
-
-		return isset(static::$_primary_key) ? static::$_primary_key : array('id');
+		return isset(static::$_primary_key) ? (array) static::$_primary_key : array('id');
 	}
 
 	/**
--- a/classes/model/crud.php
+++ b/classes/model/crud.php
@@ -21,7 +21,7 @@ class Model_Crud extends \Model implements \Iterator, \ArrayAccess, \Serializabl
 	// protected static $_table_name = '';
 
 	/**
-	 * @var  string  $_primary_key  The primary key for the table
+	 * @var  mixed  $_primary_key  The primary key(s) for the table
 	 */
 	// protected static $_primary_key = 'id';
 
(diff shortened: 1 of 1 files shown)


classes/model/crud.php
```diff
@@ -484,13 +505,15 @@ public function save($validate = true)
...
-				// workaround for PDO connections not returning the insert_id
-				if ($result[0] === false and isset($vars[static::primary_key()]))
-				{
-					$result[0] = $vars[static::primary_key()];
-				}
 				$this->set($vars);
-				empty($result[0]) or $this->{static::primary_key()} = $result[0];
+
+				// Retrieve autoincremented pkey (doesn't apply to composite pkeys or some pdo connections)
+				if ( $result[0] and count(static::primary_key() === 1) )
```
younes0: fixed on #f1672cb

classes/model/crud.php
```diff
@@ -241,12 +247,13 @@ public static function find($config = array(), $key = null)
...
 	public static function count($column = null, $distinct = true, $where = array(), $group_by = null)
 	{
-		$select = $column ?: static::primary_key();
-
 		// Get the database group / connection
 		$connection = isset(static::$_connection) ? static::$_connection : null;
 
 		// Get the columns
+		$select   = $column ?: '*';
+		$distinct = $column ? $distinct : false;
```
frankdejonge, line 255: This will break BC and is not wanted. The user doesn't know why this happens. Please remove it.

classes/model/crud.php
```diff
@@ -312,13 +319,19 @@ public static function __callStatic($name, $args)
...
 	protected static function primary_key()
 	{
-		return isset(static::$_primary_key) ? static::$_primary_key : 'id';
+		// Convert primary key from string to array 
+		if (is_string(static::$_primary_key)) 
+		{
+			static::$_primary_key = array(static::$_primary_key);
+		}
+
+		return isset(static::$_primary_key) ? static::$_primary_key : array('id');
```
frankdejonge: This can be converted into:

```
return isset(static::$_primary_key) ? (array) static::$_primary_key : array('id');
```

The check above can be removed.

classes/model/crud.php
younes0, line 255: What do you mean by BC ?
The thing is if we want to select count on composite primary keys, the easiest way is to use a wildcard.
The distinct expression is also completly useless when selecting primary keys

frankdejonge, line 255: Backwards compatibility. I understand what you mean, but this is how it was designed, we do not change the outcome because we think it's better, we do what the dev tells the function to do.

classes/model/crud.php
younes0: +1

classes/model/crud.php
younes0, line 255: Okay, so:
if pkey is composite:  'SELECT COUNT(*)' , distinct always set to false
otherwise =>  'SELECT COUNT(mypk)'  with ability to change distinct, like before

Is it what you were thinking about?

frankdejonge, line 255: The query that's generated would have to incorporate a GROUP BY clause. So,  when the $column is an array, select COUNT(*) and group by the $columns. Only if the $group_by column isn't set. This will still change the output in some cases, but has the least affect.

younes0, line 255: Actually, if $columns is not null, the behavior remains the same as before. Just making sure that we expose the same problem.

WanWizard:
@FrenkyNet what to do with this one?

frankdejonge:
After discussing this, I've decided to close this due to the fact this class needs a more structural refactoring. It's put on feature freeze until futher notice.

younes0:
And I decided to get rid of composite PK in my new projects since they are replaceable, may be slower than surrogate keys and are not supported by most of ORMs

PR #1190, opened 2012-11-09
Fix Pagination Class to be testable, and restore tests

kenjis:
Keep testing. ;-)


frankdejonge:
This won't fly. You're hardcoding spoofed test results in the class. In my book this is even worst than not testing. We're aware of the problem and know what needs to be done to fix it.

kenjis:
You mean the change in Pagination class is  hardcoding spoofed test results?

frankdejonge:
Normally you'd supply a mocked object object as a dependancy. Which is not possible in the current structure. You're now returning 1, which is dummy data, thus is incorrect.

PR #1193, opened 2012-11-11
Add query string pagination functionality

kenjis:
Supporting pagination with query string.

Eg:

```
    $config = array(
        'url_segment'  => 'page',
        'total_items'    => 100,
        'per_page'       => 10,
    );

    $pagination = Pagination::forge('mypagination', $config);
    echo $pagination->render();
```

If url_segment is int, URI segment pagination same as before, else Query string pagination.

### DIFF
--- a/classes/pagination.php
+++ b/classes/pagination.php
@@ -183,37 +183,41 @@ class Pagination
 	{
 		$type = is_int($this->config['uri_segment']) ? 'uri_segment' : 'query_string';
 
-		switch ($type)
+		// URI segment pagination
+		if (is_int($this->config['uri_segment']))
 		{
-			case 'uri_segment':
-				if (is_null($this->config['pagination_url']))
+			if (is_null($this->config['pagination_url']))
+			{
+				$url  = Uri::base();
+				$segs = Uri::segments();
+
+				if (count($segs) < $this->config['uri_segment'] - 1)
 				{
-					$url = Uri::base();
-					$segs = Uri::segments();
-					unset($segs[$this->config['uri_segment'] - 1]);
-					if (end($segs) !== Request::main()->action)
-					{
-						$segs[] = Request::main()->action;
-					}
-					$url .= implode('/', $segs);
-					$this->config['pagination_url'] = $url;
+					throw new \RuntimeException("Could not detect pagination_url.");
 				}
 
-				$this->config['pagination_url'] = rtrim($this->config['pagination_url'], '/').'/{page}';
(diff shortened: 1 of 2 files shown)


WanWizard:
Removing recalculate() from set() means any get() will return incorrect data, as the values are now only calculated on render().

As for the uri_segment, it would also be an option to say: if it's numeric, it's a segment, if it's a string, it's a GET variable? So you don't have to pass the current page number manually? This would be a one-line fix.

The class needs a fix for the pagination URL too. Currently it's a literal, and the page is "glued" to it. While it could be that the segment is in the middle somewhere, or the GET variable is already in the query string.

kenjis:
I'm sorry I forgot get(). I'll add recalculate() in __get(). Placing recalculate() in set() makes too many meaningless calls.

As for the uri_segment, your idea is nice. Thank you.

And yes pagination URL needs a fix. I'll consider.

kenjis:
I changed pagination_url using `{page}` when query string pagination.

`'pagination_url' => 'http://example.com/pagination/get?p={page}',`

And it could use POST pagination (but not yet confirmed).

"uri_segment" is not changed at all. This config name means Uri segment and if it has another meaning like GET variable or POST varible, it is confusing, isn't it?

Any comments are welcome.

WanWizard:
I still think that getting the current page is something the class should do, it should not be part of the pagination config.

I tend to have all config in the pagination config file, so I would like to define my URL and GET variable name in there, so I don't have to do that in every controller method that uses pagination. I don't want to have to include an Input call in my config file (it feels like a hack, and means you can not alter the value in code if you would want to do that).

kenjis:
I changed, now if `url_segment` is integer, pagination by URI segment the same as before.
If `url_segment` is not integer, it is pagination by query string.

And If `pagination_url` is null, Pagination class automatically detects it.

And I changed the places to call _recalculate() to reduce the calls.

kenjis:
Removed switch.

And I fixed logic to detect pagination_url, just throwing Exception if URL segments is too short.

WanWizard:
Cool. Can you change the docs accordingly, then I can merge this in.

WanWizard:
Reverted this merge. It broke all applications big time, due to your changes to the internal logic.

kenjis:
Oh, I'm sorry.

Where is broken?

WanWizard:
You moved the preparations to the constructor, which caused all settings after forging to be ignored.

PR #1199, opened 2012-11-13
Construct a namespace for the migration file from the current module or ...

Altreus:
...package name


Altreus:
Closing this so I can switch branches and sort out what we're actually doing.

PR #1200, opened 2012-11-13
Added numeric_between validation

billmn:


WanWizard:
Isn't this the same as specifying both a min and a max value?

billmn:
Yes but is more concise ... imagine :

<pre>
// Validation
->add_field('ten', 'Number', 'numeric_between[9,11]');
->add_field('ten', 'Number', 'numeric_min[9]|numeric_max[11]');
// Fieldset
->add_rule('numeric_between', array(9,1))
->add_rule('numeric_min', array(9))->add_rule('numeric_max', array(11))
// ORM
'field' => array(
    'validation' => array(
        'numeric_between' => array(9, 11)
    ),
)
'field' => array(
    'validation' => array(
        'numeric_min' => array(9),
        'numeric_max' => array(11),
    ),
)
</pre>

What about you?

huglester:
I was missing something like this. +1 :)

WanWizard:
Between 9 and 11 is only 10, right?

billmn:
Personally, I thought to include ... so between[9,11] => 9,10,11

billmn:
I haven't motivated my choice.
I would include the extreme numbers because I don't want to force to make mental calculations.

:-1:  [9,20] ... so, 9+1 and 20-1

but only from 9 to 20 ... simple

WanWizard:
Ok. Can you add the docs for this rule? Then it can be merged.

billmn:
I've already sent it : https://github.com/fuel/docs/pull/455

WanWizard:
Sorry, missed that.

billmn:
Any news?

WanWizard:
Yeah, no time. It's on the todo list ;)

Issue #1204, opened 2012-11-14
Validation error if rules passed as array


billmn:
If I pass rules as array (like an example in the docs) :

<pre>
$input = array('username' => 'fuelphp');

$val = Validation::forge();
$val->add_field('username', 'Username', array('required', 'trim', 'valid_string[alpha,lowercase,numeric]'));

Debug::dump($val->run($input));
</pre>

I get an exception :

<pre>
ErrorException [ Warning ]: explode() expects parameter 2 to be string, array given
COREPATH/classes/validation.php @ line 178
</pre>

This because at this line :
https://github.com/fuel/core/blob/1.5/develop/classes/validation.php#L178

It execute an explode on an array.

On IRC WanWizard said (14/11/2012 [21:26]) :

<pre>
<@WanWizard> billmn: please create an issue for it so it stays on the radar. no time to look into it now
</pre>

PR #1205, opened 2012-11-15
fix issue 1204

it-can:
fix issue for issue #1204


it-can:
okay @WanWizard changed it...

PR #1215, opened 2012-11-21
Cast debug_lines as an array to stop some Twig errors from causing 'Invalid argumen...

drcongo:
...t supplied for foreach()' errors as this was stopping the real error showing through.


frankdejonge:
Could you add a space after the `(array)`? Then I'll merge it, thanks!

drcongo:
Done. Thanks.

PR #1217, opened 2012-11-22
Fix 'for' in labels for radio&checkbox (Fieldset)

hayashi311:
before:
`<label for='form_form_id_0'>`

after:
`<label for='form_id_0'>`

### DIFF
--- a/classes/fieldset/field.php
+++ b/classes/fieldset/field.php
@@ -484,8 +484,7 @@ class Fieldset_Field
 						{
 							$attributes['id'] = null;
 						}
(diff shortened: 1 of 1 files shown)


kenjis:
I don't know why, but I can't see the description of this issue. Only I can see:

```
before:

after:
```

And would you please look at https://github.com/fuel/core/issues/997 ? Which is better?
And I hope you add unit tests for this.

WanWizard:
Corrected the description, Github seems to swallow the html...

PR #1218, opened 2012-11-22
Added count to Inflector::pluralize

zoe-edwards:
I thought of this today and I seemed to write a lot of unnecessary code just to get this to work.

Current situation:

```
echo $count.' ';
echo ($count > 1) ? 'people' : 'person';
```

or if you use Inflector::pluralize

```
echo $count.' ';
$entity = 'person';
echo ($count > 1) ? Inflector::pluralize($entity) : $entity;
```

However, this is what I’m proposing:

```
echo $count.' '.Inflector::pluralize('person', $count);
```

And it’ll simply return it if the second parameter is 1.

### DIFF
--- a/classes/inflector.php
+++ b/classes/inflector.php
@@ -132,7 +132,7 @@ class Inflector
 
 		// If a counter is provided, and that equals 1
 		// return as singular.
-		if ($count == 1)
+		if ($count === 1)
 		{
 			return $result;
 		}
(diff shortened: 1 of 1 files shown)


zoe-edwards:
(Of course assuming 0 is safe, because you would have ‘0 cows’ or ‘0 people’, not the singular.)

PR #1228, opened 2012-12-03
Image Class: load(), save(), reload(), load() bug.

### LINKED ISSUE #1225: Image::load() fails with all GIFs and PNGs
I was using 1.4/develop, and got this exception when uploading a GIF or PNG, and doing \Image::load()

> imagecreatefromjpeg(): gd-jpeg: JPEG library reports unrecoverable error

Tried loads of different GIF/PNGs, all get the same error.

Then I updated fuel/core to 1.5/develop, as I noticed some updates had been made to the image class files, and then I started getting this exception.

>  COULD NOT GET THE IMAGE CHANNELS FOR:

Great. Looks like that [commit](https://github.com/fuel/core/commit/e8751dc35afd743382fd4390838757d927241671) didn't really achieve anything.

I can only assume that the problem is because the Image class seems to think that my PNG or GIFs are JPEGs, therefore is using the JPEG versions of the GD functions.

lukearmstrong:
Fixes #1225

I'm not sure if this is the best way of going about this,
or if the default image config "persistence" should be set to true
to prevent the second load, but here's one way of skinning the cat.

### DIFF
--- a/classes/image/driver.php
+++ b/classes/image/driver.php
@@ -845,7 +845,7 @@ abstract class Image_Driver
 	public function reload()
 	{
 		$this->debug("Reloading was called!");
-		$this->load($this->image_fullpath, false, $this->image_extension);
+		$this->load($this->image_fullpath);
 		return $this;
 	}
 


lukearmstrong:
This is the issue I was getting...

``` php
<?php

\Image::load('image.png')->save('image.jpg'); // Fails
\Image::forge(array('persistence' => true))->load('image.png')->save('image.jpg'); // Works
```

When persistence is set to false (the default), the save() function reloads the image (not sure why, as changes may have been made which is why it is saving?) but in doing so it forces the extension to be the same extension as the file you are saving. So when loading a PNG, and saving as a JPG, it forces the extension to be JPG when it reloads.

Anyway, this persistence stuff is seemingly a mystery as further down in the code I am doing a crop_resize() yet save isn't reloading the image otherwise the crop_resize() wouldn't have happened. Probably something to do with queue()

lukearmstrong:
@quetzyg does this fix your issue too?

https://github.com/fuel/core/pull/1206

quetzyg:
@lukearmstrong, believe it or not, I also though of doing that, but then I saw a minor issue. If your first call to load() uses the second and third arguments with different values from the default ones (false, false), they will not be set on reload().

Other than that, it's a better solution than the one I had proposed :)

@FrenkyNet, any comments?

PR #1235, opened 2012-12-08
Added connection support to migrations to support replicated databases

ihildebrandt:


ihildebrandt:
Thanks for the guidance.

PR #1236, opened 2012-12-09
Enable connection setting for migrations

ihildebrandt:
This is working without issue now. My problems were being caused by two things: The new migration config file was being written without an extension and the migration config file in config/ was being deleted.

### DIFF
--- a/classes/migrate.php
+++ b/classes/migrate.php
@@ -37,6 +37,11 @@ class Migrate
 	protected static $table = 'migration';
 
 	/**
+	 * @var string  database connection group
+	 */
+	protected static $connection = null;
+
+	/**
 	 * @var	array	migration table schema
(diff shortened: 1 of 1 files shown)


WanWizard:
I can't merge this, the extension issue has already been fixed elsewhere.

ihildebrandt:
Not entirely sure why all of these commits just got reported.

WanWizard:
this is due to the last commit, which merges the develop branch in.

Check https://github.com/fuel/fuel/wiki/Contributing on how you can prepair your feature branch for a pull request.

ihildebrandt:
Thanks for the link, got it cleaned up.

WanWizard:
Would you be so kind to document this new feature too?

PR #1238, opened 2012-12-12
File move & copy methods require very loose areas

iturgeon:
I tend to make my file areas pretty strict.  This becomes a problem with the methods in File that can copy/move/rename between two endpoints.

If I need to copy a file from /fuel/app/uploads to /public/assets/img I need to make an area that is too lenient to enforce any real restriction.

For instance:

`public static function rename($path, $new_path, $area = null)`

The area sent would need to contain $path and $new_path.  This may require that the area be the root of the entire project.

So my proposal is to add a second area argument to these methods that will allow the methods to operate in two strict areas.

This implementation should still support setting a single area making it backward compatible. The new area argument defaults to the first area.

I haven't tested this yet, just getting some feedback before I go off to bed.

Thoughts?

### DIFF
--- a/classes/file.php
+++ b/classes/file.php
@@ -438,16 +438,8 @@ class File
 	 */
 	public static function rename($path, $new_path, $source_area = null, $target_area = null)
 	{
-		$path_inst      = static::instance($source_area);
-		$new_path_inst  = static::instance($target_area ?: $source_area);
-
-		if ( ! $path_inst || ! $new_path_inst)
-		{
-			return false;
-		}
-
-		$path     = $path_inst->get_path($path);
-		$new_path = $new_path_inst->get_path($new_path);
+		$path = static::instance($source_area)->get_path($path);
+		$new_path = static::instance($target_area ?: $source_area)->get_path($new_path);
 
 		return rename($path, $new_path);
 	}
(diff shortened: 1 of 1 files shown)


WanWizard:
I can see the need for this, can you test it (especially the backward compatibility) and update the docs?

classes/file.php
```diff
@@ -456,13 +457,14 @@ public static function rename_dir($path, $new_path, $area = null)
 	 *
 	 * @param   string  path to file to copy
 	 * @param   string  new base directory (full path)
-	 * @param   string|File_Area|null  file area name, object or null for non-specific
+	 * @param   string|File_Area|null  source path file area name, object or null for non-specific
+	 * @param   string|File_Area|null  target path file area name, object or null for non-specific. Defaults to source_area if not set.
 	 * @return  bool
 	 */
-	public static function copy($path, $new_path, $area = null)
+	public static function copy($path, $new_path, $source_area = null, $target_area)
```
frankdejonge: Looks like the $target_area should also have a default of `null`.

iturgeon:
Ok, added default arg for copy per @FrenkyNet 

Here are the tests I wrote for this: https://gist.github.com/4350354

I found a nuance with how file sanitizes paths for an area:

`$path = static::instance($area)->get_path($path);`

File::instance() can return false; php will burp trying to call a method on a non-object, which in turn made unit testing difficult.

So I changed the code with the involved methods to die with more grace. There are more File::instance() calls that need to be updated if this is what we want to do.

Let me know if I need any more changes here before I modify the docs.

WanWizard:
I think the basis issue here is that for some reason instance() returns false when it should throw an exception if the requested instance does not exist. 

I think this bit of the code needs to be refactored, since I assume that this is done for a specific reason.

iturgeon:
I think you're right, throwing an error would be better.  I'll restore the code in those methods. 

I'll submit an issue to add the exception and finish this up.

PR #1240, opened 2012-12-14
Consider CR+LF in Validation

ka2n:
Replacing CR+LF with CR in order to mb_strlen return valid character length
 in validation methods (max_length, mix_length, exact_length).


WanWizard:
Closing this PR.

Apart from the lack of response, the current code is correct. If the value contains a CR+LF, both should be counted. If you don't want that, make sure the value to check is correct.

PR #1252, opened 2012-12-27
fixed Query_Builder::reset() - reset all fields used to compile the query

SlKelevro:
%subj%
The 'update' and 'delete' query builders weren't resetting some of their fields, e.g. $_join in the first one (it's how I noticed the bug). The bug is simple, so is the fix :)

PS. I see that it's been here from the v.1.0, strange that no one noticed

### DIFF
--- a/classes/database/query/builder/delete.php
+++ b/classes/database/query/builder/delete.php
@@ -61,7 +61,7 @@ class Database_Query_Builder_Delete extends \Database_Query_Builder_Where
 			// Get the database instance
 			$db = \Database_Connection::instance($db);
 		}
-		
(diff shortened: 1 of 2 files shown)


frankdejonge:
Could you request a pull for this on the 1.5/develop branch? We don't accept pull-request for released version? I'll close this one for now.

SlKelevro:
@FrenkyNet ok, will do
@WanWizard I used this one as an example ;) https://github.com/fuel/core/blob/1.5/develop/classes/database/query/builder/select.php#L451
Well ok then, I'll fix it too

PR #1261, opened 2013-01-08
Added Content-Length to the Response

tomschlick:
This now counts the string length of the response body and outputs
the Content-Length header accordingly.

Signed-off-by: Tom Schlick tom@tomschlick.com

### DIFF
--- a/classes/response.php
+++ b/classes/response.php
@@ -289,7 +289,7 @@ class Response
 
 		if($send_headers)
 		{
-			$this->set_header('Content-Length', strlen($body));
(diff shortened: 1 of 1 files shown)


tomschlick:
According to the php docs its been in since the first php 5 version so unless a user can specifically remove them I'd assume they are always available?

http://php.net/manual/en/function.mb-strlen.php

WanWizard:
I was thinking about the first sentence here: http://php.net/manual/en/mbstring.installation.php

tomschlick:
Don't know how I didn't see that...

PR #1268, opened 2013-01-12
added Mongo_Db::get_cursor()

tomschlick:
added a new method to directly get the mongodb cursor from the mongodb class
this is useful for things like data processing so you dont have to load the
whole dataset into memory. also switched the Mongo_Db::get() method to use
this method when it needs to communicate with the db to keep everything DRY.

### DIFF
--- a/classes/mongo/db.php
+++ b/classes/mongo/db.php
@@ -624,20 +624,33 @@ class Mongo_Db
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


WanWizard:
Can you document this new method? You know a lot better what it does than I. :smiley:

tomschlick:
Writing it now! :)

tomschlick:
PR here https://github.com/fuel/docs/pull/488

PR #1269, opened 2013-01-13
Does not strip extension when uri ends with a slash

hackoh:
``` php
// URL: http://localhost/example/index.html
echo Input::uri(); // return /example/index (correct)

// URL: http://localhost/example/1.2.3/
echo Input::uri(); // return /example/1.2 (incorrect)
```

Is this a bug ?
I implemented a branching process when the URI ends with a slash.

### DIFF
--- a/classes/input.php
+++ b/classes/input.php
@@ -179,18 +179,22 @@ class Input
 		// Deal with any trailing dots
 		$uri = rtrim($uri, '.');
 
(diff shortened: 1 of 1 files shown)


WanWizard:
Good catch! Not so sure about the regex though, they are quite expensive, I've come up with something else.

hackoh:
thanks

PR #1271, opened 2013-01-14
Add solution for delivery query strings when uri_segment is numeric

hackoh:
If want delivery query strings and want set uri_segment is numeric in slash path,
Then I should set config below.

``` php
// URL: http://localhost/test/2?keyword=pagination
        $config = array(
            'pagination_url' => '/test/{page}?'.http_build_query(Input::get()),
            'total_items' => 100,
            'per_page' => 10,
            'uri_segment' => 2,
        );

```

But I think better setting is below.

``` php
        $config = array(
            'pagination_url' => '/test/',
            'total_items' => 100,
            'per_page' => 10,
            'uri_segment' => 2,
            'query' => Input::get(),
        );
```

Some argue that there is a problem with the design URI, But, for example, I think when doing renovations add GET parameters to the existing management systems, this feature that is needed.

### DIFF
--- a/classes/pagination.php
+++ b/classes/pagination.php
@@ -130,6 +130,7 @@ class Pagination
 		'show_first'              => false,
 		'show_last'               => false,
 		'pagination_url'          => null,
+		'query'                   => array(),
 	);
 
 	/**
@@ -511,11 +512,17 @@ class Pagination
 				$this->config['pagination_url'] = rtrim($this->config['pagination_url'], '/').'/{page}';
 			}
 
+			if (strpos($this->config['pagination_url'], '{query_string}') === false and is_array($this->config['query']) and count($this->config['query']))
+			{
+				// if no query_string placeholder is present, add one to the end of the URL
+				$this->config['pagination_url'] .= '?{query_string}';
+			}
+
 		}
 
 		// return the page link
 		empty($page) and $page = 1;
-		return str_replace('{page}', $page, $this->config['pagination_url']);
+		return str_replace(array('{page}', '{query_string}'), array($page, http_build_query($this->config['query'])), $this->config['pagination_url']);
 	}
 
 }


WanWizard:
This leaves a question mark terminating the URL if no query is defined in the pagination array. It also doesn't work if the pagination_url already contains a query string.

It is not the task of the pagination class to construct URL's, it's job is to deal with pagination, which any URL you give it. I don't see any added value in this appoach, it's very application specific.

@FrenkyNet your opinion?

hackoh:
> This leaves a question mark terminating the URL if no query is defined in the pagination array

Then, It does not add question mark.
It sees config['query'] count.

``` php
if (strpos($this->config['pagination_url'], '{query_string}') === false and is_array($this->config['query']) and count($this->config['query']))
```

> It also doesn't work if the pagination_url already contains a query string.

Originally, If query string mix in the pagination_url, It do not work correctly.
example:

``` php
        $config = array(
            'pagination_url' => '/test/?foo=bar',
            'total_items' => 100,
            'per_page' => 10,
            'uri_segment' => 2,
        );
       // return http://localhost/test/?foo=bar/1, http://localhost/test/?foo=bar/2, http://localhost/test/?foo=bar/3
```

frankdejonge:
This will make the Pagination class responsible for creating urls, which is not it's job. The `Uri` class already has this functionality. For portability it's recommended to use that class to generate the `pagination_url`, so it would be super easy so make it have the arguments. Though I do understand the problem. We could create a {page} tag detection so query strings are still supported? @WanWizard ?

WanWizard:
Pagination doesn't care where the placeholder is, it does a dumb replace, and it already has this detection.

The real issue seems to be that the class has an issue with query strings when a URI segment is used for the page number. And if that's the case, then that should be fixed, and not worked around.

WanWizard:
This fix makes sure make_link() doesn't care what you define as URL, it will just put the page number in. 

If you specify the placeholder youself, it doesn't even care what the defined uri_segment is, it will just use it. If not, it will insert the page number at the location indicated by uri_segment. And it will preserve the query string.

quetzyg:
Just found out earlier that this PR messes up the pagination.

``` php
// Configuration
$pagination = Pagination::forge('default', array(
    'pagination_url' => Uri::create('users/listing'),
    'total_items' => $query->count(),
    'per_page' => 10,
    'uri_segment' => 3
));
```

Expected URL when paginating: **users/listing/[page_number]**
Actual URL when paginating: **users/[page_number]**

When dumping the Pagination instance config, we have a 10 element Array:
current_page (Integer): 2
offset (Integer): 10
per_page (Integer): 10
total_pages (Float): 2
total_items (Integer): 19
num_links (Integer): 5
uri_segment (Integer): 3
show_first (Boolean): false
show_last (Boolean): false 
pagination_url (String): "http://127.0.0.1/example.com/users/{page}" (41 characters)

The expected pagination_url should be: **http://127.0.0.1/example.com/users/listing/2**

hackoh:
Uri::create() receives relative URI from your application ROOT as an argument. 
uri_segment specifies the segment number of absolute URI. 

try this.

``` php
// Configuration
$pagination = Pagination::forge('default', array(
    'pagination_url' => Uri::create('users/listing'),
    'total_items' => $query->count(),
    'per_page' => 10,
//  'uri_segment' => 3
    'uri_segment' => 4
));
```

quetzyg:
@hackoh I did fiddle around and saw that putting the uri_segment + 1 solved the problem partialy. But then, the pagination links has the current page always stuck at the first two links. Can you reproduce it, or do you want more info?

hackoh:
I obtained the following result. 
In my case, it seems to operate correctly.

``` php
        $pagination = Pagination::forge('default', array(
            'pagination_url' => Uri::create('example.com/users/listing'),
            'total_items' => 100,
            'per_page' => 10,
            'uri_segment' => 4
        ));
        echo $pagination->render();
```

``` html
<div class="pagination">
    <span class="previous-inactive">
            <a href="#">&laquo;</a>

</span>
<span class="active">
            <a href="#">1</a>

</span>
<span>
            <a href="http://localhost:21013/example.com/users/listing/2">2</a>

</span>
<span>
            <a href="http://localhost:21013/example.com/users/listing/3">3</a>

</span>
<span>
            <a href="http://localhost:21013/example.com/users/listing/4">4</a>

</span>
<span>
            <a href="http://localhost:21013/example.com/users/listing/5">5</a>

</span>
<span>
            <a href="http://localhost:21013/example.com/users/listing/6">6</a>

</span>
<span class="next">
            <a href="http://localhost:21013/example.com/users/listing/2">&raquo;</a>

</span>

</div>
```

Can you show me detail code and result?

WanWizard:
@quetzyg,

The old (< 1.4) behaviour just took the pagination_url and glued the segments onto it. Which in your case just worked, because your page number was the last segment. A uri_segment of 500 would have probably worked.

However, in all other pages, where the page number isn't the last segment of the URL, or where the URL has query parameters, the old class produced rubbish.

quetzyg:
@hackoh I can easily understand that I have to account for the entire base URL as a segment, and I do get the same result as you. But if you go to the second, third, or nth pagination URL, what do you get? In my case, I get the same results as the first, but with a URL with a different pagination segment number. This is:

http://localhost:21013/example.com/users/listing
http://localhost:21013/example.com/users/listing/2
http://localhost:21013/example.com/users/listing/n

give same results.

@WanWizard in my case, the page number is always the last segment. But now that you mentioned using a 500 uri_segment, I've come to realise that in this particular case, the base URL may be getting accounted for twice. That is:

Should http://127.0.0.1/example.com/ be considered one, or two segments?

hackoh:
@quetzyg

I went to the second, third, and more, I got links correctly.
If you write like:

``` php
$pagination->current_page = 1;
```

then you get same results.

WanWizard:
'example.com' is 1 segment, the URI is exploded on the slash.

I think from a functional point of view, any segments part of the base URL should not be counted, or even treated by the URI class, as they are not part of the application.

If your configured base_url = "http://127.0.0.1:21013/example.com/", what does Uri::segments() return?

quetzyg:
@WanWizard I always set **base_url** to null, but setting it to **http://localhost/example.com/** gives me the same thing:

```
array(3) { [0]=> string(5) "users" [1]=> string(7) "listing" [2]=> string(1) "2" }
```

Which I find it correct.

WanWizard:
This PR is no longer relevant, and a new ticket has been opened for this issue: #1319. Let's continue there if needed.

PR #1277, opened 2013-01-15
Updated to allow to set full path to APPPATH, COREPATH and PKGPATH on testing

pirosikick:
In case to set full path to `<server name="core_path">` on phpunit.xml, test fails because `trim` function trims the first `/` of the path.

Signed-off-by: Hiroyuki Anai pirosikick@gmail.com


WanWizard:
Can you send this PR on 1.5/develop? We don't accept PR's on 1.4.

PR #1280, opened 2013-01-16
_make_link bug fix

hackoh:
1. parse_str is destructive. So can't use in ternary operator.
2. $url['port'] is not empty, then add ':' after $url['host'].
   (for example: http://localhost:8080)

### DIFF
--- a/classes/pagination.php
+++ b/classes/pagination.php
@@ -466,6 +466,16 @@ class Pagination
 			// break the url in bits so we can insert it
 			$url = parse_url($this->config['pagination_url']);
 
+			// parse the query string
+			if (isset($url['query']))
+			{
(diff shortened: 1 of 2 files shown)


hackoh:
Moreover, if $url['query'] is not held as array before

``` php
if (is_numeric($this->config['uri_segment']))
```

 when uri_segment is numeric, an error occurs in http_build_query.

kenjis:
Why don't you add unit tests?

WanWizard:
I think the issue with parse_str() is with the second parameter. That should not be there, it should return the result, not alter it.

edit: stupid remark, the guy who designed this function to use a by-ref instead of a return value should be ....

PR #1282, opened 2013-01-16
Added ability to use 'in' operator in a where query.

ihildebrandt:
Added logic to the _compile_conditions() method of the query builder to accept an 'in' operator with an array of values. 

Also added a few unit tests for good measure.


frankdejonge:
Looks like you started coding before testing, the query builder can already do this.

PR #1283, opened 2013-01-16
Update classes/fieldset.php

phil-lavin:
Related to #1278. Also fix add_after.


peter-fritchley:
Probably should have done that in the original. Whoops.

PR #1286, opened 2013-01-18
Add comment arg to  DBUtil::create_table().

takyam-git:
I want to set comment to table on DBUtil::create_table() method.
Because, [mroonga wrapper mode](http://mroonga.github.com/docs/userguide/wrapper.html) needs set engine on table comment.

like this...

``` sql
CREATE TABLE diaries (
  id INT PRIMARY KEY AUTO_INCREMENT,
  content VARCHAR(255),
  FULLTEXT INDEX (content)
) ENGINE = mroonga COMMENT = 'engine "innodb"' DEFAULT CHARSET utf8;
```

this is very simple fix. I added argument on DBUtil::create_table() method.

### DIFF
--- a/classes/dbutil.php
+++ b/classes/dbutil.php
@@ -109,9 +109,10 @@ class DBUtil
 	 * @param    string    $engine         storage engine overwrite
 	 * @param    string    $charset        default charset overwrite
 	 * @param    array     $foreign_keys   an array of foreign keys
+	 * @param    string    $comment        table comment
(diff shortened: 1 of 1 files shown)


WanWizard:
This is a platform specific change, which is not generic enough to go into the core.

You can easily work around this by using the $engine parameter, by passing `"mroonga COMMENT = 'engine \"innodb\"'"`

PR #1290, opened 2013-01-21
set data before checking for primary key in crud model's constructor

fenom:
I don't understand why primary key is checked before data are even set. How can existing record ever be detected?

### DIFF
--- a/classes/model/crud.php
+++ b/classes/model/crud.php
@@ -385,17 +385,17 @@ class Model_Crud extends \Model implements \Iterator, \ArrayAccess, \Serializabl
 	 */
 	public function __construct(array $data = array())
 	{
-		if (isset($this->{static::primary_key()}))
-		{
-			$this->is_new(false);
-		}
-
 		if ( ! empty($data))
 		{
 			foreach ($data as $key => $value)
 			{
 				$this->{$key} = $value;
 			}
+		}
+
+		if (isset($this->{static::primary_key()}))
+		{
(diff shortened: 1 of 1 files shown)


WanWizard:
This is because of a design issue in PHP.

If you use one of the database driver methods to create objects (for example mysql_fetch_object), the properties are actually set BEFORE the constructor is called, and not the other way around (as you would expect, since you instantate an object before you assign a property).

So it's there by design, and should absolutely stay.

fenom:
But why does that mean the constructor has to check for primary key before setting properties? If properties are already set before the constructor is called, what's the harm in running that piece of code to set properties first? If this must stay, then can you tell me how I am supposed to create an object of an existing record?

WanWizard:
If the object is populated before the constructor is called (like I described), it's a record from the database, and therefore it is not 'new' anymore.

With your change, the new state is set to false when the primary key is set, which does not mean a database record exists for the data in the object (since it was passed to the constructor in $data, including the primary key). 

So your change introduces a bug.

fenom:
So if you construct a new object manually and pass in the primary key in data, that is not intened to mean it's an existing record in the database?

WanWizard:
Correct. A record will only be created when you save() the model, at which time is_new() will be set to false. 

Passing a value for the PK to the constructor doesn't mean there's a record. You could have a table without auto_increment, so you HAVE to pass a PK value, even for new records...

PR #1291, opened 2013-01-22
Support localized language URLs

davispuh:
### About

This feature allows to make localized URLs. ie.
`http://domain/lang1/controller/method`
`http://domain/lang2/controller/method`

While I've seen other implementations which change Uri class this DOES NOT because it have several issues.
### Implementation

Uri class isn't changed and Lang class is improved properly -> all code is fully backward compatible and everything will work same as before for everyone who don't need such feature. No need to change existing code.
### Usefulness

I'm pretty sure there's people who would love such feature (including me).

And this implementation have several advantages.
- Easy to make same page links for different languages.
- Switching between languages user gets same page (in other language) without redirects.
- Backward compatible
- Lang class supports this feature making it in right place where it should be. And allows routing handle some specific language needs.
- Can be easily customizable if language need to be in other component than first (eg. last). While currently this can't be achieved with config, it's really easy to either directly edit lang class or just overload in app.
- SEO friendly
- Language in URL can be overridden with `Config::set('language',$lang)`
### Usage

To use this feature there are required few code changes, but nothing much.
1. all links which are supposed to be for same language must be replaced with
   `Lang::localized($uri)`, eg. `<a href="<?php echo Lang::localized('/blog/article/')?>"/>Article</a>`
   (Note: currently `localized` doesn't support schema or relative urls, and it expects that uri starts with `/` but it can be implemented if needed, now can do just `'https://domain'.Lang::localized('/blog/article/')`)
2. make links which change language. either use `Lang::localized('/blog/article/','ru')` or directly write `<a href="/ru">RU</a>`
3. in app `config.php` add available languages `'languages' => Array('en', 'ru')`
4. change `routes.php` config to also include localized routing.
#### `routes.php` config changes explained

Will need routes for both cases when language is included in URI and when isn't.
So will need to add another `_root_` ie. `$routes[$langs]` and all other routes prefixed with `$routes[(({$langs})/)?route]`

There's 4 possible configurations.
##### 1.

SEO friendly. Default language without any prefix. ie. `http://domain/controller` and all other languages `http://domain/lang/controller`. BUT `http://domain/default_lang/controller` returns 404

`config.php`: `'language' => 'en'` // specify which language will be served as default. without included in URL

`routes.php`

```
<?php

$languages = Config::get('languages');
unset($languages[0]);
$langs = implode('|', $languages);

return array(
'_root_' => 'welcome/index', // The default route
"({$langs})" => 'welcome/index', // The default route in other language
"(({$langs})/)?hello(/:name)?" => array('welcome/hello', 'name' => 'hello'),
);
```
##### 2.

Default language without any prefix. ie. `http://domain/controller`, BUT `http://domain/default_lang/controller` will still work.

everything same as previous, just without `unset($languages[0]);`
##### 3.

All languages `http://domain/lang/controller`. BUT `http://domain/controller` will return 404. There's exception to `_root_`, it will give default language (language_fallback).

`config.php`: `'language' => ''` // no default language, MUST be in URL
`'language_fallback' => 'en'` // Fallback language if language isn't in URL

`routes.php`

```
<?php

$langs = implode('|', Config::get('languages'));

return array(
'_root_' => 'welcome/index', // The default route
"({$langs})" => 'welcome/index', // The default route in other language
"(({$langs})/)?hello(/:name)?" => array('welcome/hello', 'name' => 'hello'),
);
```
##### 4.

Make your own routes, unlimited possibilities. Maybe you don't have some article in that language? No problem, just give in default language or give page informing that it's not available.
### PS

I've been using this feature already some while and it works great :)

Also btw I replaced some code which didn't make any sense...

```
   if ($language === null) {
      $languages = static::$fallback;
      array_unshift($languages, $language ?: \Config::get('language'));
      $language = reset($languages);
   }
```

firstly it was in several places exactly same and `$languages` wasn't used anywhere only these few lines. And that code part is used only when `$language` is `null`, but `null` is always `false`, so `$languages` always will be with prepended with `Config::get('language')` and so always `$language=Config::get('language')` when `$language` is `null`

### DIFF
--- a/classes/lang.php
+++ b/classes/lang.php
@@ -27,13 +27,38 @@ class Lang
 	public static $lines = array();
 
 	/**
+	 * @var  string default language when language is not specified in URI
+	 */
+	public static $default_language;
+
+	/**
 	 * @var  array  language(s) to fall back on when loading a file from the current lang fails
 	 */
 	public static $fallback;
 
 	public static function _init()
 	{
+		static::$default_language = \Config::get('language');
 		static::$fallback = (array) \Config::get('language_fallback', 'en');
+	}
+
+	/**
+	 * Returns currently active language.
+	 *
+	 * @return   string    currently active language
+	 */
+	public static function lang() {
+		$language = \Config::get('language');
+		if ($language == self::$default_language) {
+			$lang_segment = \Uri::segment(1);
+			if (!empty($lang_segment) and $lang_segment != false) {
+				$languages = \Config::get('languages');
+				if (!empty($languages) and array_search($lang_segment, $languages) !== false) {
+					$language = $lang_segment;
+				}
+			}
+		}
+		return $language;
 	}
 
 	/**
@@ -49,7 +74,7 @@ class Lang
 	public static function load($file, $group = null, $language = null, $overwrite = false, $reload = false)
 	{
 		// get the active language and all fallback languages
-		$language or $language = \Config::get('language');
+		$language or $language = self::lang();
 		$languages = static::$fallback;
 
 		// make sure we don't have the active language in the fallback array
@@ -145,9 +170,7 @@ class Lang
 	{
 		if ($language === null)
 		{
-			$languages = static::$fallback;
-			array_unshift($languages, $language ?: \Config::get('language'));
-			$language = reset($languages);
+			$language = self::lang();
 		}
 
 		// prefix the file with the language
@@ -199,9 +222,7 @@ class Lang
 	{
 		if ($language === null)
 		{
-			$languages = static::$fallback;
-			array_unshift($languages, $language ?: \Config::get('language'));
-			$language = reset($languages);
+			$language = self::lang();
 		}
 
 		return isset(static::$lines[$language]) ? \Str::tr(\Fuel::value(\Arr::get(static::$lines[$language], $line, $default)), $params) : $default;
@@ -222,9 +243,7 @@ class Lang
 
 		if ($language === null)
 		{
-			$languages = static::$fallback;
-			array_unshift($languages, $language ?: \Config::get('language'));
-			$language = reset($languages);
+			$language = self::lang();
 		}
 
 		isset(static::$lines[$language]) or static::$lines[$language] = array();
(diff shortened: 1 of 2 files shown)


frankdejonge:
Looks lik you've got your line endings set wrong. Can you fix that? It's impossible to see what's changed.

WanWizard:
I'm not sure this should be part of the framework, as it is application specific functionality. 

I wouldn't want my application to all of a sudden do something else just because I have a route that starts with /en/ (and which has nothing to do with a language, or which is handled a different way).

frankdejonge:
Also, I've had this same problem. What I did was use a route filter which detects a language, sets the config accordingly and strips the segment for normal routing to work. That was maybe 10 lines of code, super easy. Seeing both @WanWizard and I don't see this being a addition we'd like to see in the core I'm closing this pull. Hop by on IRC if you'd like to see my solution :)

davispuh:
sorry about that formatting, it was messed up only in last commit as I used another IDE. Anyway I removed that commit and added it again preserving original formatting.

And like I said, this doesn't break any existing code and you can have routes named whatever you want even `/en/` and nothing will change in usage. Even if you don't want use this feature because you already have some custom language support there's high chance that it still will work without conflicts or without need to change anything. (but of course if you want use this feature and you already have similar language support then changes might be needed). 

Because this feature are controlled with config option `'languages' => Array()` where you specify languages which will be used from URL, if this is empty array or null then this feature are simply disabled. Also all routing is controlled by YOU, by application. There's no magical URL rewrites. In fact all URLs are unchanged. Routes does all job.

While this might be considered as application specific, but isn't nearly ever feature such? not everyone need every feature, but that's why we've so many configuration options. While yes it would be useless if such feature would be needed only for couple of applications, but this is not the case. Also about "core". This isn't really core, changed are only Lang class. Does all applications are multilingual? Maybe I need only one language, then why there's such class in "core"? So it's right place where to do it. It's really simple if you don't need it zero configuration. But I'm sure there are applications which would benefit from this feature. Currently only way for application to use localized URls are with monkey patching or some hacks with possibility that it might break something else.

I suggest please review this feature more properly and test it with all your use cases. Because currently it just seems language support is not finished if you need take language from URL for every application you've to write some custom code.

davispuh:
Here can see changes
[Compare 1.6/develop vs patch](https://github.com/davispuh/core/compare/fuel:1.6/develop...davispuh:patch)

PR #1299, opened 2013-01-24
Remove PHP end tags

davispuh:
This is not really issue, but I think it should be good practice and also consistency with rest of framework.

While it's unlikely but possible that somehow accidentally space is after `?>` that would make it sent to browser and thus all next header changes would fail because headers already sent. And this kind of mistakes are very hard to track down, which file is causing it.

### DIFF
--- a/vendor/phpquickprofiler/console.php
+++ b/vendor/phpquickprofiler/console.php
@@ -94,5 +94,3 @@ class Console {
 	}
 
 }
-
-?>
--- a/vendor/phpquickprofiler/phpquickprofiler.php
+++ b/vendor/phpquickprofiler/phpquickprofiler.php
@@ -227,5 +227,3 @@ class PhpQuickProfiler {
 	}
 
 }
-
-?>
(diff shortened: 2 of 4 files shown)


WanWizard:
The Framework code doesn't have any PHP end tags, as per the coding standards.

Everything in vendor is (as the name implies), third party code, which is included as-is.

davispuh:
It's just silly, if you include it, then maintain and improve it. Otherwise could just point everyone to get it from original source. Where it probably will be updated more often.

davispuh:
actually indeed WHY phpseclib is bundled here? phpseclib is actively improved. But here's included old version. In upstream those end tags are already removed with added new fixes and functionality.

[phpseclib uses composer](https://packagist.org/packages/phpseclib/phpseclib) and can be also found at GitHub [ phpseclib/phpseclib](https://github.com/phpseclib/phpseclib)

frankdejonge:
@davispuh: we're making the move to composer and rome wasn't built in a day. So, get off your high horse. Helping is making improvements, it's not about complaining about insignificant stuff. We're in the progress of fixing this and are focussed on structurally tackling this, not patching fixes.

WanWizard:
PHPSecLib has been updated not to long ago. Their repo only has one branch, master, to which they commit. So for me, they don't have a stable branch on which we can rely on. So we stick to official releases.

Besides that, we don't need the entire package, only the Crypt and Hash functions. Which are based on mathematical algorithms, which do not change. None of the recently added features are relevant for the core (for which it is included).

PR #1301, opened 2013-01-25
Improve Lang class for easier extendability

davispuh:
Currently if anyone wants to extend Lang class and provide his own way how to determine which is currently active language, he's forced to overload most of Lang functions (load, save, get, set, delete)

This improvement adds function `Lang::get_lang()` which determines currently active language and will be used by all those previously mentioned functions. Thus if you need to change the way how active language is determined you can just overload this one function without need to change anything else.

It doesn't change functionality so it's fully compatible with current applications. Also it improves Lang class existing code.

``` php
if ($language === null)
{
     $languages = static::$fallback;
     array_unshift($languages, $language ?: \Config::get('language'));
     $language = reset($languages);
}
```

I mentioned this code part already in #1291 which does nothing more than `$language = \Config::get('language')` and also is used in several places.

### DIFF
--- a/classes/lang.php
+++ b/classes/lang.php
@@ -43,10 +43,7 @@ class Lang
 	 */
 	public static function get_lang() {
 		$language = \Config::get('language');
-		if (empty($language))
-		{
-			$language = self::$fallback[0];
-		}
+		empty($language) and $language = static::$fallback[0];
 		return $language;
 	}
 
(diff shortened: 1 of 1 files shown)


WanWizard:
Please use static instead of self, this needs late static binding for extensibility. Also, the code is not according to the coding standards, so please correct that too.

And can you add the new method to the documentation?

davispuh:
was editing from GitHub. sorry about `self`, just fixed. Could you tell more about what should be changed for coding standards?

Yeah sure, will do ;)

WanWizard:
See http://fuelphp.com/docs/general/coding_standards.html for the coding standards.
What immediately caught my eye was the curly brace that was not on a new line.

davispuh:
I've read it. Somehow missed that brace but I just don't see any other issues...

PR #1303, opened 2013-01-28
Fixed a weird is_scalar conflict

tomschlick:
In some cases if $checked = false but is boolean and the value of the
field is '0' the is_scalar would evaluate to true which obviously shouldn't happen.
making the statement two if clauses fixes this because it wont drop down into the
is_scalar block

### DIFF
--- a/classes/form/instance.php
+++ b/classes/form/instance.php
@@ -300,9 +300,12 @@ class Form_Instance
 			if ( ! is_array($checked))
 			{
 				// If it's true, then go for it
-				if (is_bool($checked) and $checked === true)
+				if (is_bool($checked))
 				{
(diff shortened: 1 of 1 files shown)


WanWizard:
This is odd, both statements are identical?

tomschlick:
No because in the first if the $checked is boolean but not === true it will then go to the elseif which could evaluate to true if the value of the field is something that is_scalar picks up on. 

In my change if its bool but not checked it just doesn't add anything which should be the proper behavior.

WanWizard:
Ah, missed the elseif, wasn't visible in the diff. That indeed changes things.

tomschlick:
Thanks!

PR #1304, opened 2013-01-30
Added ref (next, prev) attributes to the pagination links

pconcepcion:
Signed-off-by: Pablo de la Concepción Sanz pconcepcion@gmail.com

Yust added the rel attributes to the next and previous pagination links, the doc is updated also (in a separate pull Request)


WanWizard:
What is the reason for adding this?

pconcepcion:
Mainly because this **adds some semantic to the links**.

This can improve the navigation for crawlers (helping **SEO**) and browsers (that will understand better the relation between the pages) and can take advantage of it.

WanWizard:
Ok.

PR #1305, opened 2013-01-30
New FuelPHP Upload class compatibility problems

lukearmstrong:
I am currently working on a site that is due to go live in March so I am updating Fuel regularly.

Here is my original code that worked with Fuel 1.5/develop.

``` php
<?php

$config = array(
    'ext_whitelist'  => array('csv'),
    'path'           => APPPATH.'data/',
    'prefix'         => date('Y-m-d_H-i-s_')
);

\Upload::process($config);

if (\Upload::is_valid()) {
    \Upload::save();
} else {
    throw new \Exception("Invalid file uploaded, please try uploading a .csv file.");
}

$files = \Upload::get_files();
$data = file_get_contents($files[0]['saved_to'].$files[0]['saved_as']);
```

Updating to 1.6/develop causes this error when uploading a file.

```
call_user_func() expects parameter 1 to be a valid callback, function '/tmp/phpvJjKU0' not found or invalid function name
fuel/vendor/fuelphp/upload/src/FuelPHP/Upload/File.php:447
```

Straight away I noticed that line was using `call_user_func()` incorrectly, but also, it shouldn't have been getting there. Please bear with me whilst I submit the pull request to fix that issue.

I imagine there will be further issues with the `fuelphp\upload` class, as it doesn't appear to be complete. But this pull request, and my other pull request, at least make it work with my existing code.

I know that my pull requests are going to break uploading via FTP, if that worked to begin with.

### DIFF
--- a/classes/upload.php
+++ b/classes/upload.php
@@ -89,7 +89,7 @@ class Upload
 
 		// add the callbacks
 		$config['langCallback'] = '\\Upload::langCallback';
-		$config['moveCallback'] = '\\Upload::moveCallback';
+		$config['moveCallback'] = null;
 
 		// get an upload instance
 		static::$upload = new \FuelPHP\Upload\Upload($config);
--- a/classes/upload.php
+++ b/classes/upload.php
@@ -168,15 +168,26 @@ class Upload
 		foreach ($files as $file)
 		{
 			$data = array();
+
 			foreach ($file as $item => $value)
 			{
 				$item == 'element' and $item = 'field';
 				$item == 'tmp_name' and $item = 'file';
 				$data[$item] = $value;
 			}
+			
+			if (isset($data['path'])) {
+				$data['saved_to'] = $data['path'];
+			}
+			
+			if (isset($data['filename'])) {
+				$data['saved_as'] = $data['filename'];
+			}
+
 			$data['field'] = str_replace('.', ':', $data['field']);
 			$data['error'] = ! $file->isValid();
 			$data['errors'] = array();
+
 			$result[] = $data;
 		}
 


lukearmstrong:
Related to https://github.com/fuelphp/upload/pull/1

WanWizard:
moveCallback should return false, unless you have setup an FTP upload service. By setting it to null, you effectively disabled the option to FTP the file instead of a local disk move.

The error was caused by a missing argument, which you fixed in your other PR (Thanks!)

If it is not complete, I'd like to know what is missing. I tried to make it as compatible as possible, but obviously it's difficult to catch everything in the first commit. So the second commit is something I can merge, the first not. Can you amend the PR?

lukearmstrong:
The logic for determining if FTP should be used is flawed.

By simply checking if there is a callback set doesn't mean you want to use FTP.
https://github.com/fuelphp/upload/blob/master/src/FuelPHP/Upload/File.php#L445

The default `moveCallback` value is `null`, but it is changed here and cannot be configured?
https://github.com/fuel/core/blob/1.6/develop/classes/upload.php#L92

So, if you are not using FTP, then nothing happens as it just returns false.
https://github.com/fuel/core/blob/1.6/develop/classes/upload.php#L129

WanWizard:
The check is there because in 2.0 (fuelphp\upload is a 2.0 package that is backported to 1.x), you don't have the \Fuel\Core\Upload facade, so it doesn't have a value by default. 

The FTP function is typical for 1.x, and will not be present like that in 2.0.

That is why in 1.x the callback is always there, and when called, it will check if with_ftp is configured on not (which is something that happens outside of the upload package). The idea is that if you want an alternative move method, the upload package doesn't care how you want to move. This is for BC purposes.

lukearmstrong:
Fixed https://github.com/fuel/core/commit/44a55cef097cf4c47cce149964a16e40e1354cb2

PR #1306, opened 2013-01-30
Update classes/fieldset.php

peter-fritchley:
Allow array style form fields to be populated properly when the process method is passed a nested array.


WanWizard:
That is odd, I was sure that was already supported?

PR #1315, opened 2013-02-08
Fix bugs in FileClass

kobash:
[FileClass](https://github.com/fuel/core/blob/1.6/develop/classes/file.php) does not work if it is used with lock.
I think the problem is  'open_file()' & 'close_file()' , so I modified these functions.
Please confirm.


hackoh:
@kobash First, how about reading code agreement?

kenjis:
Yes. http://fuelphp.com/docs/general/coding_standards.html

kobash:
Oh.. Sorry for my careless.
I'll confirm and  try again.

Thank you for letting me know.

PR #1316, opened 2013-02-08
Fixed Debug::dump() when called from call_user_func_array()

hackoh:
I made shortcut function like below.

``` php
function d()
{
    $args = func_get_args();
    call_user_func_array('\\Debug::dump', $args);
}
```

But then, I got error.

``` html
ErrorException [ Notice ]: Undefined index: file

COREPATH/classes/debug.php @ line 57
```

``` php
// If being called from within, show the file above in the backtrack
 if (strpos($backtrace[0]['file'], 'core/classes/debug.php') !== FALSE)
```

Is this a bug or not?
If this is a bug, I attach a little draft fix.
Can you check it?

### DIFF
--- a/classes/debug.php
+++ b/classes/debug.php
@@ -53,15 +53,21 @@ class Debug
 		{
 			$backtrace = debug_backtrace();
 
-			// If being called from within, show the file above in the backtrack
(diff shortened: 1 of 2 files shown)


WanWizard:
This was a bit too complex.

I've commited a fix for it, can you change the PR to include the tests only?

PR #1317, opened 2013-02-08
Qualified PK's in Model_Crud

ihildebrandt:
Add ability to qualify primary key with the table name, use qualified primary keys in all 'where's

Signed-off-by: P. Isaac Hildebrandt king.isaac@gmail.com

### DIFF
--- a/classes/model/crud.php
+++ b/classes/model/crud.php
@@ -89,7 +89,7 @@ class Model_Crud extends \Model implements \Iterator, \ArrayAccess, \Serializabl
 	 */
 	public static function find_by_pk($value)
 	{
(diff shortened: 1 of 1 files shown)


WanWizard:
What is the use-case of this change? 

There is never more than one table involved in a query, so column names are always unique.

ihildebrandt:
I have run into instances of needing to join in a table on some of my queries. It was easier to use the built in find() method and build the joins in pre_find(), but If this is outside of the scope of this object then I'll move those queries into my class.

WanWizard:
It is poluting core code with something that is application specific, so it doesn't belong in the core.

PR #1321, opened 2013-02-12
Propose Arr::search()

hackoh:
I encountered the case where Arr::search() was needed.
Would you check?

### DIFF
--- a/classes/arr.php
+++ b/classes/arr.php
@@ -893,4 +893,57 @@ class Arr
 		$values = array_filter($arr, 'is_array');
 		return $all_keys ? count($arr) === count($values) : count($values) > 0;
 	}
+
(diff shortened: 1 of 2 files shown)


hackoh:
If this PR is adopted, I am ready to PR to docs.

WanWizard:
PR is ok. If you can create the PR for the docs, I'll merge them both.

WanWizard:
Thanks!

frankdejonge:
@WanWizard I'd be in favor of defaulting the recursive param to `true` it's because of recursion that this is handy, therefor defaulting makes sense. Agree?

WanWizard:
Agreed. Makes much more sense.

PR #1326, opened 2013-02-14
Made configurable first/last, next/previous markers

SCIF:
Pagination class didn't can to configure text of first/last and next/previous links. Fixed it.
Added new template values: first-marker, previous-marker, next-marker, last-marker.

### DIFF
--- a/classes/pagination.php
+++ b/classes/pagination.php
@@ -48,13 +48,13 @@ class Pagination
 	{
 		// old pre-1.4 mapping to new instance methods
 		static $mapping = array(
-			'get' => '__get',
-			'set' => '__set',
-			'set_config' => '__set',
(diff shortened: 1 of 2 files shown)


WanWizard:
Why only add it to the bootstrap template, and not to the default template? And can you also update the pagination docs?

SCIF:
Ok. I will do request with default and bootstrap templates :) Also if it's good idea — i will update docs.

WanWizard:
And I think it will be a good idea to use HTML entities (&raquo;) instead of the characters themselfs.

SCIF:
Don't know how to attach patch to existence pull request. Therefore — #1336

WanWizard:
#1336 got merged.

PR #1328, opened 2013-02-14
Validation post

ngarneau:
In the comments it says: 
"@param   array  input that overwrites POST values" 
but the POST data is never use. Now the method will use the POST data when input is null.

### DIFF
--- a/base.php
+++ b/base.php
@@ -364,26 +364,3 @@ if (!function_exists('http_build_url'))
 		;
 	}
 }
-
-/**
- * Loads in the classes used for the error handlers.  The class_exists() calls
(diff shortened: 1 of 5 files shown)


WanWizard:
This is not correct. 

If POST data would never been used, in the last two years validation would have been unusable (as everyone uses it to validate form input).

All methods in Validation will use the input() method to retrieve an input value. That will check $this->input first (to see if you have overridden a value manually), and if not found, will check the POST data by calling Input::param().

p.s. for next time, if you send a pull request, make sure you're branch is up to date, and rebase before you send the PR. See https://github.com/fuel/fuel/wiki/Contributing for some pointers on how to contribute.

PR #1330, opened 2013-02-15
Controller_Template: needless router() method duplication (or on purpose ?)

ghost:


WanWizard:
Well spotted. Controller has it, so Controller_Template will inherit it. It was too quick a fix I'm afraid.

PR #1332, opened 2013-02-15
Made compatability fixes to upload class

phil-lavin:
2 changes. Errors used to have keys for their number,message array. I've re-added these to try to maintain backwards compatibility.

save() takes parameters as noted in its comments. I've used the 2nd param to set the path of the file else it saves to '//'. This maintains compatibility with old code that calls \Upload::save(id, path)


WanWizard:
Well spotted, thanks!

PR #1333, opened 2013-02-17
Update classes/controller/rest.php

quetzyg:
Remove forgotten type hint from Fuel\Core\Controller\Rest::router()

### DIFF
--- a/classes/controller/rest.php
+++ b/classes/controller/rest.php
@@ -95,7 +95,7 @@ abstract class Controller_Rest extends \Controller
 	 * @param  string
 	 * @param  array
 	 */
-	public function router($resource, array $arguments)
+	public function router($resource, $arguments)
 	{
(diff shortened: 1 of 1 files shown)


WanWizard:
Why that? This type hint has always been there, no need to remove it. It only had to be removed from Controller, because it was never there in the past...

quetzyg:
Well, because the Controller_Rest class extends the Controller class, and I get notices about declaration compatibility when using the Hybrid/Rest Controller.

```
Runtime Notice!

ErrorException [ Runtime Notice ]: Declaration of Fuel\Core\Controller_Rest::router() should be compatible with Fuel\Core\Controller::router($resource, $arguments)
```

WanWizard:
Ah, chips, didn't think of that.

quetzyg:
:+1:

PR #1337, opened 2013-02-18
Changed validation rules with string length to use str class

dmyers:


WanWizard:
Did you test this? My Str class doesn't have a method called 'len'?

it-can:
Should this not be \Str::length()

dmyers:
@WanWizard Good call, that's what I get for staying up so late coding, thanks! I updated the original.

Thanks @IT-Can that was the correct function.

PR #1342, opened 2013-02-19
Fix csrf_autoload bug. It be check only 'POST', 'PUT' and 'DELETE' http methods.

takyam-git:
Currently check allmethod on csrf_autoload is true.  
It will be catch \SecurityException when I access to some pages by http GET method.

this pull request check http methods and run check_token() only post, put and delete method.

### DIFF
--- a/classes/security.php
+++ b/classes/security.php
@@ -54,7 +54,8 @@ class Security
 		static::$csrf_old_token = \Input::cookie(static::$csrf_token_key, false);
 
 		// if csrf automatic checking is enabled, and it fails validation, bail out!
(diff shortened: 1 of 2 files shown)


WanWizard:
Can you add this to the docs too?

takyam-git:
I send pull request to fuel/docs now.
https://github.com/fuel/docs/pull/514

WanWizard:
This can no longer be merged as it conflicts with your other change. I've added it manually.

takyam-git:
thanks!

WanWizard:
I've made generate_token() general (not related to csrf) and public, as it can be used in other situations where you need a random token.

PR #1343, opened 2013-02-19
Change logic on CSRF token generate.

takyam-git:
Currently logic of generate token can be inferred.
because, uniqid() uses microtime.

I tried use more secure logic that used salt.


takyam-git:
@WanWizard 
Thank you for review.
I added 2 commits.

I created new method generate_token(). It use 'sha512' or 'md5'.
And we will be able to extend logic of generate token.

PR #1346, opened 2013-02-19
Added first/last inactive generation

SCIF:
Made first/last links configurable. It allows not to dissapear links if we are already at first/last page, if needed. By default templates are empty – for backward compability.

Also i replace {uri} with # in inactive url templates. That allows to configure link of inactive pages, not hardcoded # as before. It brakes bc a little. I think that really small number of people will get broken urls of inactive pages, if apply my changes. I can add sanitizing of inactive template to remove {uri} from inactive template but think that it's very weird and clumsy.

### DIFF
--- a/classes/pagination.php
+++ b/classes/pagination.php
@@ -142,23 +142,27 @@ class Pagination
 		'first'                   => "<span class=\"first\">\n\t{link}\n</span>\n",
 		'first-marker'            => "&laquo;&laquo;",
 		'first-link'              => "\t\t<a href=\"{uri}\">{page}</a>\n",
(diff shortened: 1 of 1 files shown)


WanWizard:
This indeed breaks BC, and can't be merged like this. 

You need to keep the str_replace() as is, because in existing applications, {uri} might still be used in templates or in custom code. If the template doesn't contain {uri} but something else, the replace will simply do nothing.

Otherwise I don't have a problem with this change. If you undo the replace change, and update the documentation, I can merge this.

SCIF:
#1347

PR #1347, opened 2013-02-20
Inactive patterns now have not {uri}. Added first/last inactive generation.

SCIF:
Backward fully compatible. Changed template for using "#" instead of "{uri}", that is more clear. Now behavior of pagination is fully configurable.

Suggest to declare deprecation of using "{uri}" in inactive links and remove support of them in 1.7. (But it's just suggestion).


SCIF:
Correct PR is #1348

PR #1350, opened 2013-02-20
Easy way to switch default connection.

SDpower:
Hi.
## supply a Easy way to switch default connection.

As we know all the DB kind obeject use \Database_Connection::instance() to connection.
for example:
  A,B modules Use master,slave connection.
A switch connection to master just:

``` php
\Database_Connection::$default = "master"
```

And B switch connection to slave just:

``` php
\Database_Connection::$default = "slave"
```

end than modules work out must give back it.

``` php
\Database_Connection::$default = "default"
```
## Bad thing is.

User ma be very sure hie use which one  connection.

### DIFF
--- a/classes/database/connection.php
+++ b/classes/database/connection.php
@@ -49,7 +49,7 @@ abstract class Database_Connection
 		if ($name === null)
 		{
 			// Use the default instance name
-			$name = \Config::get('db.active');
+			$name = static::$default;
 		}
 
 		if ( ! isset(static::$instances[$name]))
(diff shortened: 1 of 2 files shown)


WanWizard:
This will break all existing applications, so it's a no-go. The $default property is a remainder of the past, and must be removed. 

If you want to change the default DB, either change the active value in the config file (if it's permanent), or use `\Config::set('db.active', 'somethingelse')`.

SDpower:
OK!
It's my miss.
tks!!

PR #1354, opened 2013-02-21
Improved driver handling of class names to use ucwords

dmyers:


WanWizard:
Are you expecting driver names with underscores? If not, then calling Str might be expensive.

dmyers:
@WanWizard  Ah I didn't think about that. I was just thinking to get them all consistent.

I did notice the `Request` class uses `Inflector::words_to_upper()` though which made me wonder if they should all use that.

WanWizard:
words_to_upper() is using ucwords, and is therefore not MB safe. It also supports spaces, and replaces them with underscores, which makes it a bit of an odd choice for unifying a classname...

dmyers:
@WanWizard Oh yeah good point. I thought about changing that driver too to be consistent, but wasn't quite sure. The space checking did look strange for classes.

I just thought the Str class was the defacto standard to make the framework multibyte safe across the framework.

WanWizard:
Yes, but the question is if you're going to create controller names and method names in UTF-8? Does PHP even support that, a class called ひらがな or 平仮名 ?

dmyers:
Do you want to go without the Str case and just use ucwords? I just noticed `fuel.php` uses `Str::ucwords()` with the always load classes config.

Maybe we change the `Request` too? Or is that pretty dangerous?

WanWizard:
I think this is going the wrong way. 

Problem is that currently the Fuel autoloader is (partly) case sensitive. Which is why we struggle to come up with ways to make sure the classnames are "standard". But in PHP, class names are case insensitive, which is confusing for a lot of people. It also means you'll get in to trouble if you use two different cases for the same class in your code, as the autoloader sees it as two different classes.

This really has to be thought through, and perhaps we should leave it as this, and instead focus on 2.0 development.

PR #1355, opened 2013-02-21
Fixed Event_Instance::unregister() bug

indigofeather:
in https://github.com/fuel/core/pull/1344 I just followed the annotation. 
But the target to unregister is incorrect. 
**This will unset all events.** It's my fault.

### DIFF
--- a/classes/event/instance.php
+++ b/classes/event/instance.php
@@ -102,7 +102,7 @@ class Event_Instance
 		{
 			if ($callback === null)
 			{
-				$this->_events = array();
+				unset($this->_events[$event]);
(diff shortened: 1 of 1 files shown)


classes/event/instance.php
```diff
@@ -100,9 +100,9 @@ public function unregister($event, $callback = null)
 	{
 		if (isset($this->_events[$event]))
 		{
-			if ($callback === null)
+			if (is_null($callback))
```
frankdejonge: Why use a function when you can do it without?

indigofeather:
updated!

PR #1365, opened 2013-03-05
Added $_mass_whitelist/blacklist to Model_Crud

huglester:
For example I have Model_Customer to work with and it's ok.

Now I need Model_Customer_Api which would extend the Model_CUstomer, but I want only fields like:

``` php
    protected static $_mass_whitelist = array(
        'first_name',
        'last_name',
    );
```

to be changed from Api model, but not others. so I pass only those fields.
It will work only for ->set($input);

For example: 
$input = array('first_name' => 'test', 'password' => 'wont_be_save');

So like $customer = Model_Customer_Api::find_by_pk(1);
$customer->set($input):
$customer->save();

will save only the 'fist_name'.
But if we will use the:
$customer->password = 'will_be_save';
it'll still work.

I think you understand the use case.

Also made the mass_blacklist, so we can blacklist for example 'username' field, where api won't be able to change the username.

Thanks,
huglester

### DIFF
--- a/classes/model/crud.php
+++ b/classes/model/crud.php
@@ -46,6 +46,16 @@ class Model_Crud extends \Model implements \Iterator, \ArrayAccess, \Serializabl
 	// protected static $_properties = array();
 
 	/**
+	 * @var  array  $_mass_whitelist  The table column names which will be set while using mass assignment like ->set($data)
+	 */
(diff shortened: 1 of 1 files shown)


tomschlick:
:+1:

PR #1366, opened 2013-03-06
Sorting array containing objects

hackoh:
This is proposing a solution about sorting array containing objects.
(For example, an array containing ORM instances.)
But, I think this approach creates few overhead.
How do you think?

### DIFF
--- a/classes/arr.php
+++ b/classes/arr.php
@@ -677,7 +677,15 @@ class Arr
 
 		foreach ($array as $k => $v)
 		{
-			$b[$k] = static::get($v, $key);
+			try
+			{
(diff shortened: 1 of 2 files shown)


WanWizard:
@FrenkyNet What do you think? Not too sure personally...

tomschlick:
If you implement \ArrayAccess on your objects you can use them as arrays just fine which eliminates the need to convert the objects to arrays manually. What problems were you experiencing in doing so (if that was the case)? I believe the Arr::sort already handles this just fine...

hackoh:
@tomschlick 
Oh, I had overlooked that point.
My stupid worry was solved by your advice. 
Thanks!

PR #1370, opened 2013-03-07
Fix Fuel::STAGE const name and value

kenjis:


WanWizard:
Question is if you call the environment 'staging', shouldn't the constant not also be called STAGING ?

It will possibly break BC, so we have to document this.

kenjis:
If we can change it to `Fuel::STAGING`, I want to do.

WanWizard:
Ok. I think it's more consistent. I'll update the documentation

kenjis:
I'm sorry I forgot my sign. So I amended the commit with fixing the const name.

PR #1371, opened 2013-03-08
Change Config::load() process

hackoh:
Change to check whether each array are assoc or not.
As a result of this change, the following settings can be.

``` php
\Config::load(array(
    'test' => array(
        'foo' => 'bar',
        'fuel' => 'php',
        'multi' => array(
            'key' => 'value',
            'numbers' => array(1, 2),
        ),
        'array' => array(1, 2, 3),
    )
));

\Config::load(array(
    'test' => array(
        'foo' => 'boo',
        'multi' => array(
            'numbers' => array(),
        ),
        'array' => array(1, 9),
    )
));

print_r(\Config::get('test'));

// Array
// (
//     [foo] => boo
//     [fuel] => php
//     [multi] => Array
//         (
//             [key] => value
//             [numbers] => Array
//                 (
//                 )
//         )
//     [array] => Array
//         (
//             [0] => 1
//             [1] => 9
//         )
// )
```

### DIFF
--- a/classes/config.php
+++ b/classes/config.php
@@ -107,17 +107,17 @@ class Config
 		// Serialize no assoc arrays.
 		$serialize = function (&$array) use (&$serialize)
 		{
-			foreach ($array as $k => $v)
+			foreach ($array as $k => &$v)
 			{
 				if (is_array($v))
 				{
 					if (\Arr::is_assoc($v))
 					{
-						$serialize($array[$k]);
+						$serialize($v);
 					}
 					else
 					{
-						$array[$k] = array(md5('_serialized_'.$k) => true, 'content' => serialize($v));
+						$v = array(md5('_serialized_'.$k) => true, 'content' => serialize($v));
 					}
 				}
 			}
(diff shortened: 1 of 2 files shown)


WanWizard:
This is incorrect merge behaviour, and will break lots of applications that expect multi-dimensional arrays to be merged properly, instead of being overwritten.

hackoh:
@WanWizard 
In Config class, I think no one hoping that non-assoc arrays will be merged.
I think, should precede process of loading config correctly than process of merging array correctly.

This approach can help for some problems.
For example, `security.output_filter`, `whilelist_classess`, and more.
Don't you think so?

WanWizard:
Those issues are already addressed in the Security class.

I do have applications that require correct merging of config files (in modules and packages).

hackoh:
> I do have applications that require correct merging of config files (in modules and packages).

If possible, can you show me one?

The example about which I worry:

``` php
// app/config/config.php
return array(
    'fuel' => 'php',
    'foo' => array(1, 2, 3)
);
```

``` php
// app/config/development/config.php
return array(
    'foo' => array(4, 5)
);
```

then, I expect

``` php
echo Config::get('fuel'); // php
print_r(Config::get('foo')); // array(4, 5)
```

But results are not so.
returned `array(1, 2, 3, 4, 5)`.
Is this behavior correct?
I don't think so..

hackoh:
And, I changed Config_File too. (I was absentminded..)
https://github.com/hackoh/core/compare/1.6/develop...1.6/change_config_load

WanWizard:
In your example, I expect

```
array(1, 2, 3, 4, 5)
```

which is the correct way of merging indexed arrays.

frankdejonge:
@hackoh: in many of FuePHP's classes we're using it in this way. It's a design choice we've made and we stand by it. It'll be dreadful for BC if we change that now and has been very beneficial in the past. It's made configs flexible and extendible. To add to it, what you're proposing here is extremely expensive and will slow a lot of apps down as a result of this change.

PR #1375, opened 2013-03-11
adding a stack trace to profiler , makes it easier to find duplicate / s...

lukepolo:
adding a stack trace to profiler , makes it easier to find duplicate / slow queries

### DIFF
--- a/classes/profiler.php
+++ b/classes/profiler.php
@@ -47,13 +47,14 @@ class Profiler
 		return static::$profiler ? static::$profiler->display(static::$profiler) : '';
 	}
 
-	public static function start($dbname, $sql)
+	public static function start($dbname, $sql,$stacktrace)
 	{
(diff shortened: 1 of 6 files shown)


WanWizard:
Use \Fuel::clean_path() to make the filenames look "nice", it also deals with core and package paths.

classes/profiler.php
```diff
@@ -47,13 +47,14 @@ public static function output()
 		return static::$profiler ? static::$profiler->display(static::$profiler) : '';
 	}
 
-	public static function start($dbname, $sql)
+	public static function start($dbname, $sql,$stacktrace)
```
frankdejonge: can you add a space before `$stacktrace`?

WanWizard:
hmmm, this seems to break all DB profiling,,,

PR #1380, opened 2013-03-19
View class keep the data used at the latest rendering

hackoh:
When rendering a View of the multilayer, the value set in the controller does not reach the end of the rendering process.
If you pass the value of the lower layer to the View, We have to verbose description as the second argument of View::forge.
Or, there is also a hack like the following.

``` php
<body>
   <div id="container">
       <?php echo View::forge('next/view', $__data) ?>
   </div>
</body>
```

But I do not think that the way is correct.
It is because malfunction will be caused if `'__data'` is set to View. ..and more reasons.

This change enables the following.

``` php
<body>
   <div id="container">
       <?php echo View::forge('next/view', true) ?>
   </div>
</body>
```

### DIFF
--- a/classes/view.php
+++ b/classes/view.php
@@ -38,6 +38,11 @@ class View
 	protected static $global_filter = array();
 
 	/**
+	 * @var  array  Keep the data used at the latest rendering
+	 */
+	protected static $last_data = array();
+
+	/**
 	 * @var  array  Current active search paths
 	 */
 	protected $request_paths = array();
@@ -108,9 +113,9 @@ class View
 		{
 			$data = get_object_vars($data);
 		}
-		elseif ($data and ! is_array($data))
+		elseif ($data !== true and $data and ! is_array($data))
 		{
-			throw new \InvalidArgumentException('The data parameter only accepts objects and arrays.');
+			throw new \InvalidArgumentException('The data parameter only accepts objects and arrays and true.');
 		}
 
 		$this->auto_filter = is_null($filter) ? \Config::get('security.auto_filter_output', true) : $filter;
@@ -122,6 +127,9 @@ class View
 
 		if ($data !== null)
 		{
+			// If the second parameter is true, restore the last data.
+			$data === true and $data = static::$last_data ?: array();
+
 			// Add the values to the current data
 			$this->data = $data;
 		}
@@ -560,6 +568,9 @@ class View
 		{
 			throw new \FuelException('You must set the file to use within your view before rendering');
 		}
+
+		// Store the last data
+		static::$last_data = $this->data;
 
 		// combine local and global data and capture the output
 		$return = $this->process_file();


hackoh:
Test is here, https://github.com/hackoh/core/commit/51c4a67c3f755a61c1f8b236fa7a5b4fc3aaf63d
But I do not know where do I put the view file for the test..

WanWizard:
This is quite a hack too, and not needed. In a view file, you can access the current View object via $this, so the correct usage should be

``` php
<body>
   <div id="container">
       <?php echo View::forge('next/view', $this->get()) ?>
   </div>
</body>
```

hackoh:
@WanWizard 
No.
Rendering process is not in scope of view instance.
I proposed other approach in https://github.com/fuel/core/issues/1382

WanWizard:
In what case does it not work? "No" doesn't really help me much.

A view file is included within the local scope of a method of the View class, and this method is called on a View object, so $this should always be available, and refer to the current View object you're in?

hackoh:
I'm sorry for my weak explanation.

"$this can be used in anonymous functions." from PHP5.4.

http://php.net/manual/en/functions.anonymous.php

My development environment is 5.3.2..

hackoh:
As you say, I was confirmed that in an environment of PHP5.4 is accessible as $this.

kenjis:
By the way, why do you need to call another view with passing all data in a view?

hackoh:
@kenjis 
Examples are already present in the crud view generated by the scaffold.

``` php
// in app/views/admin/***/create.php and edit.php
<?php echo render('admin/***/_form'); ?>
```

In _form.php, It refers the post directly.
If values transmitted transparently from the controller , did not become such an implementation.

kenjis:
@hackoh 
It is not true. Input::post() is used for getting new user input data when editing.

``` php
// in _form.php
<?php echo Form::input('name', Input::post('name', isset($form) ? $form->name : ''), array('class' => 'span6')); ?>
```

`$form` in above is passed with View::set_global in a controller, so `render('admin/***/_form')` doesn't need any params.

hackoh:
@kenjis 

I have a bit of opinion about the View::set_global, and the implementation of accessing to the Input from the View, but it is to derail from the discussion of this PR.

I need to get current view instance, PHP 5.4 can do so, But 5.3 can not do so.

WanWizard:
I never use set_global, I (probably) have the same opinion as you.

When I was using 5.3.x, I passed $__data to the sub-view, but I've been on 5.4.x for quite some time, and since then I've been using $this (been doing that so long I forgot that only works in 5.4+).

I'm considering upping the requirement for Fuel2 to PHP 5.4, so we can use other new features too...

kenjis:
Probably you two don't like set_global().

If you want to get current view instance in PHP 5.3,
your https://github.com/fuel/core/pull/1382 seems to be better.

PR #1382, opened 2013-03-20
Store the active View instance

hackoh:
This is other approach of https://github.com/fuel/core/pull/1380

The rendering process is not in scope of view instance. (PHP 5.4 <)
So can not access as `$this`.
And can not current view instance delivery to rendering scope as $this.
(If a View instance is passed as $this, it will not operate normally by behavior of PHP. )

I think, the approach passing as a variable is contamination of variable in scope of rendering.

This approach makes it possible to access an instance of the current view.

``` php
<body>
   <div id="container">
       <?php echo View::forge('next/view', View::active()->get()) ?>
   </div>
</body>
```

If you agree to this approach, I'll prepare tests and docs.

### DIFF
--- a/classes/view.php
+++ b/classes/view.php
@@ -38,6 +38,11 @@ class View
 	protected static $global_filter = array();
 
 	/**
+	 * @var  View   Holds the global active view instance
+	 */
+	protected static $active = false;
+
+	/**
 	 * @var  array  Current active search paths
 	 */
 	protected $request_paths = array();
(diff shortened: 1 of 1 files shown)


frankdejonge:
@hackoh this is done by design. All the data needed in the view, should be passed to it, therefor access to the view object itself should not be necessary. You could however set a self-referencing param in the viewdata. Or maybe look into viewmodels if you need more logic in view generation.

WanWizard:
The view is rendered inside a closure, that is why $this doesn't work in PHP < 5.4. As a workaround, you can access the entire data array using $__data, and pass that on to the client view.

PR #1385, opened 2013-03-23
Changed filename to be optional when saving image

dmyers:
Before if you wanted to use the same image you would have to pass the file path twice.

Before:

``` php
$image = Image::load('fuel.png')
$image->resize('10', '10');
$image->save('fuel.png');
```

After:

``` php
$image = Image::load('fuel.png')
$image->resize('10', '10');
$image->save();
```

### DIFF
--- a/classes/image/driver.php
+++ b/classes/image/driver.php
@@ -614,8 +614,13 @@ abstract class Image_Driver
 	 * @param   string  $permissions  Allows unix style permissions
 	 * @return  array
 	 */
-	public function save($filename, $permissions = null)
+	public function save($filename = null, $permissions = null)
 	{
+		if (empty($filename))
+		{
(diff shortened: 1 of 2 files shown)


classes/image/driver.php
```diff
@@ -614,8 +614,12 @@ public function grayscale()
 	 * @param   string  $permissions  Allows unix style permissions
 	 * @return  array
 	 */
-	public function save($filename, $permissions = null)
+	public function save($filename = null, $permissions = null)
 	{
+		if (empty($filename)) {
```
frankdejonge: Can you put the `{` on a newline, it's our coding style.

dmyers: Whoops, good catch. I always forget, I have like 3 styles in my head from personal to work to fuel.

PR #1388, opened 2013-03-23
Just to be consistent :)

huglester:


dmyers:
Ha! I just wasn't 100% sure if it would break or not. :+1:

PR #1390, opened 2013-03-25
Create multidimensional arrays from Database_Result::as_array

Altreus:


Altreus:
@FrenkyNet Pointed out a) it's a bit inefficient and b) it breaks (null, 'column')

I think b) can be generalised across the n-ary output to allow null at any place

a) could be solved with not using Arr; but I didn't spot any significant slow-down so I will concentrate on compatibility first.

Altreus:
This is closed by https://github.com/fuel/core/pull/1391, which is considerably better. :)

PR #1391, opened 2013-03-25
Refactor Database_Result::as_array to be n-ary for n dimensions

Altreus:
Supersedes https://github.com/fuel/core/pull/1390

A refactor of this function allows it to accept any number of arguments and produce N-1 dimensions of array. 

The original four behaviours - (null,null) (null,colname) (colname,null) (colname,colname) - are preserved.

The method extends itself to allow _null_ to mean "all" and _colname_ to use that column's value.

An extra example or two in the doc also show this.


frankdejonge:
@Altreus we've talked this through, but this is not something we want to see built into the framework. Luckily it's easy to extend the class and use that for yourself.

PR #1397, opened 2013-04-11
Fix log folders permission

kenjis:
`mkdir` mode is modified by the current umask.

### DIFF
--- a/classes/log.php
+++ b/classes/log.php
@@ -52,9 +52,19 @@ class Log
 
 		if ( ! is_dir($filepath))
 		{
-			$old = umask(0);
-			@mkdir($filepath, \Config::get('file.chmod.folders', 0777), true);
-			umask($old);
+			$filepath_year = \Config::get('log_path').date('Y/');
+			$permission = \Config::get('file.chmod.folders', 0777);
+			
+			if (is_dir($filepath_year))
+			{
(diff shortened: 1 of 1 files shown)


WanWizard:
The problem with umask is that it isn't thread-safe, and the new PHP installs become more and more multi-threaded. It is safer to replace the mode by a separate chmod call after the mkdir.

kenjis:
Thanks for your comment. I've changed.

WanWizard:
Where did this extra mkdir come from? The filepath comes from the config, it's possible it doesn't start with the year at all. 

And perhaps the @ should be removed (it is very slow), and put the mkdir in a try/catch block to capture errors if there are no rights to create the path. Now it silently fails, which can't be a good thing.

kenjis:
See https://github.com/fuel/core/blob/1.6/develop/classes/log.php#L51

``` php
$filepath = \Config::get('log_path').date('Y/m').'/';
```

I did not change anything. It starts with the year, and the month.

Removing @ seems to be good idea.

WanWizard:
Sorry, missed that. Been fixing stuff for two days strait, can't see properly anymore. :laughing: 

I'll merge and add the try/catch.

kenjis:
Thank you. Great job!

PR #1400, opened 2013-04-15
fix documnet root access for uri with extension

takecore:
This is Windows issue for documnet root access.
If uri access has extension.
For example 'http://test.com/index.html'.

```
$uri_info['dirname'] === '\';
```

This commit is fix of document root access for uri with extension.


WanWizard:
Thanks for finding it! 

Going to solve it differently though, never liked that pathinfo() in there (it's not a path, it's a URI).

PR #1404, opened 2013-04-19
Added Validation option us_phone_number to validation class

chris-roerig:
This will remove all non numeric characters, count the total digits and compare it to the passed length argument (default is 10).

This allows US phone numbers such as:
(000) 000-0000
000-000-0000
0000000000
000 000 0000
or any other format that includes the number of digits required.

Note: Its up to the programmer to format the number after the info has been submitted and before it is saved.

### DIFF
--- a/classes/validation.php
+++ b/classes/validation.php
@@ -969,4 +969,18 @@ class Validation
 		}
 		return \Arr::get($parsed, 'error_count', 1) + ($strict ? \Arr::get($parsed, 'warning_count', 1) : 0) === 0;
 	}
+
+	/**
+	 * Checks if input can be formated as a US phone number.
+	 * 
+	 * @param	string|int
+	 * @param	int 	The number of expected numeric digits.
+	 * @return	bool
+	 */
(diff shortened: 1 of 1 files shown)


WanWizard:
There are several things wrong with this.

First, the algorithm used has nothing to do with a US phone number, it just strips non-digit characters from the input string. Second, it's not a validation method (it doesn't validate anything), it transforms the input string. And third, it returns an integer, which means there are limits to the length, and it will convert "0000000001" to 1, which means it doesn't represent the input anymore.

This rule is not generic but application specific (i.e. it suits your particular application purpose). This PR can therefore not be accepted.

chris-roerig:
I think you might be misunderstanding the intent of this code. 
1. It returns bool and is not meant to return any manipulated input. In other words, its not supposed to return a formated number because we already have Num::format() for that.
2. In the US (and probably Mexico/Canada) phone numbers can be represented in many ways yet the common theme is they always have the same number of digits. What this method does is counts the digits and looks for the appropriate amount.

I hope this makes my intention a bit clearer.

PR #1405, opened 2013-04-23
prevent invalid XML node names

benharold:
Format::to_xml() method can return invalid XML if the node name starts with a number or if it contains only numeric and other non-alpha characters. To fix, we remove any non-alphanumeric characters before checking to see if the key can be used to instantiate a new DOMElement.

Example of invalid results before this patch:
oil c
$array('30:0' => 'thing');
print_r(\Format::forge($array)->to_xml());

### DIFF
--- a/classes/format.php
+++ b/classes/format.php
@@ -145,15 +145,15 @@ class Format
 
 		foreach ($data as $key => $value)
 		{
-			// no numeric keys in our xml please!
-			if (is_numeric($key))
+			// replace anything not alpha numeric
+			$key = preg_replace('/[^a-z_\-0-9]/i', '', $key);
+
+			// no invalid keys in our xml please!
+			if ( ! static::_valid_xml_key($key))
 			{
 				// make string key...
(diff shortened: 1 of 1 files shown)


frankdejonge:
I lean towards a "won't fix" here, a dev is responsible for creating the node names, so when doing that incorrectly he didn't do his job ;) although this fix isn't /bad/ it creates n times more function calls and object creations, which could easily be avoided. Agree @WanWizard ?

WanWizard:
@FrenkyNet agree.

benharold:
Core methods such as `\Arr::flatten()` can result in invalid key names. What if, instead of calling the function that tests object creation, we just use a regex to test for validity? After a bit of research, I came up with this regex to validate XML node names: `[a-zA-Z_][a-zA-Z0-9_\.-]*`

WanWizard:
Core methods can not result in invalid key names. Invalid input to Format::to_xml() can result in invalid key names.

If you as a developer feed the output of a "core method" (or any other piece of PHP code) into to_xml() without validation, they you're providing invalid input. You should validate, and make sure it doesn't happen.

PR #1409, opened 2013-04-26
add before_view() method to ViewModel

takyam-git:
The `ViewModel::before()` method excuted on `ViewModel::forge()`.  
`before()` can't get params of set after forge.

``` php
example: <?php
//on controller
$this->template->content = ViewModel::forge('foo/bar');
$this->template->content->set('param_a', $param_a);

//on class View_Foo_Bar
public function before()
{
    Debug::dump($this->param_a); // OutOfBoundsException [ Error ]: View variable is not set:param_a
}
```

The `before_view()` method can do it.

### DIFF
--- a/classes/viewmodel.php
+++ b/classes/viewmodel.php
@@ -111,9 +111,14 @@ abstract class ViewModel
 	}
 
 	/**
-	 * Executed before the view method
+	 * Executed before the view method on forge
 	 */
 	public function before() {}
+
+	/**
+	 * Execute before the view method
+	 */
+	public function before_view() {}
 
 	/**
 	 * The default view method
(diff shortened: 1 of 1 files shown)


WanWizard:
Isn't the actual problem here that the before() is called in the Viewmodel constructor, instead of before rendering the view? It seems more logical to me to move the before() call...

takyam-git:
Yes, I think that too, but I feel uneasiness of backward compatibility.
If there's no problem move `before()`, then I want to do it.

WanWizard:
I don't see an immediate issue, so if you can move it, please do.

takyam-git:
I sent new pull request #1412,  please see it.

PR #1415, opened 2013-04-30
Changed view file name property to be public

dmyers:


WanWizard:
I don't see any reason why, you're not supposed to manipulate this from outside the class.

dmyers:
Most classes allow data to be set and retrieved. Why can't I retrieve the filename passed to this object when I set the filename beforehand? When passing around objects this is important.

PR #1416, opened 2013-04-30
Changed view get data method to be public

dmyers:


WanWizard:
I don't see any reason why, you're not supposed to access this from outside the class.

dmyers:
Most classes allow data to be set and retrieved. Why can't I retrieve the data passed to this object when I set the data beforehand? When passing around objects this is important.

frankdejonge:
@dmyers wouldn't viewmodels help you here? It's a more object oriented way of going about views, you can add getters and setters for you view objects which would allow you to pass it around and make editing of viewdata possible?

PR #1417, opened 2013-05-01
set fallback of lang::get() for tabular fieldset label

PWesterdale:
Currently if there is no Lang set for the label the tabular fieldset returns false and no label is displayed.
I've amended this behaviour and provided the label from the tabular fieldset model as the third lang::get parameter so it always falls back to what you are passing it if no Lang alternative is found.

### DIFF
--- a/classes/fieldset.php
+++ b/classes/fieldset.php
@@ -600,7 +600,7 @@ class Fieldset
 				{
 					continue;
 				}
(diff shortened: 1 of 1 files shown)


WanWizard:
Ah, thanks, totally missed that. 

We use a Lang extension in our app framework that returns missing language strings as bold-black-on-fluorecent-yellow span's, so you can quickly see you're lang file is missing an entry. So we never have an empty return value.

PR #1425, opened 2013-05-07
Added build_query_string() to URI class

phil-lavin:
As discussed in #fuelphp. Designed for use like this...

``` html
<a href="<?= \Uri::build_query_string(\Input::get(), ['foo'=>'bar']) ?>">Foo</a>
```

### DIFF
--- a/classes/uri.php
+++ b/classes/uri.php
@@ -235,7 +235,7 @@ class Uri
 	 * @param   array    ...
 	 * @return  string
 	 */
-	public static function build_query_string(array $params)
+	public static function build_query_string($params)
 	{
 		$params = array();
 
--- a/classes/uri.php
+++ b/classes/uri.php
@@ -228,6 +228,26 @@ class Uri
 		return $url;
 	}
 
+	/**
+	 * Builds a query string based on a number of merged arrays
+	 *
+	 * @param   array    Array to merge
+	 * @param   array    ...
+	 * @return  string
+	 */
+	public static function build_query_string()
+	{
+		$params = array();
+
+		foreach (func_get_args() as $arg)
+		{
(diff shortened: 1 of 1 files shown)


classes/uri.php
```diff
@@ -228,6 +228,26 @@ public static function base($include_index = true)
...
 	}
 
+	/**
+	 * Builds a query string based on a number of merged arrays
+	 *
+	 * @param   array    Array to merge
+	 * @param   array    ...
+	 * @return  string
+	 */
+	public static function build_query_string()
```
frankdejonge: Could you add one type hinted param without a default value in the method signature? Like so: `string(array $params)`

phil-lavin: Done :)

phil-lavin:
Yeh, except @FrenkyNet asked me to add that. Want the type hinting removed?

phil-lavin:
At which point, it somewhat voids the point of having an argument. Want the whole argument removed?

frankdejonge:
@WanWizard I thought it would only accept arrays as query strings are associative as well, we could remove the type hinting to allow string input too. My point was to ensure input in the first place.

WanWizard:
No, the idea was to be able to add strings as well, if you want to drop that then the typehint is ok, but the string-to-array conversion can be removed.

Reason for wanting this is that we use GET variable a lot as toggle, so the presence of the variable is used, and not a value, like http://example.org?featureA&featureB&featureC

phil-lavin:
@WanWizard Removed the hint

WanWizard:
Do we have docs for this method? And preferably a unit test?

frankdejonge:
@phil-lavin any updates on this?

PR #1434, opened 2013-05-13
add default extension if there is not one set

PWesterdale:
The docs say 'The path to the config file, relative to the config directory. Do not include the file extension (".php" is assumed).'
which currently is not correct. If you do not specify a file extension none gets set so this sets the default if one is not provided.

### DIFF
--- a/classes/config.php
+++ b/classes/config.php
@@ -150,6 +150,7 @@ class Config
 
 		$info = pathinfo($file);
 		$type = 'php';
+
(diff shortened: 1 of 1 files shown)


indigofeather:
Dear @WanWizard 
This change makes migration system broken, it will generate  <code>migrations.php.php</code>.

WanWizard:
Has already been reverted.

indigofeather:
Thank you very much!

PR #1436, opened 2013-05-15
Fixup hardcoded *nix paths, some spaces around concatenation

AlbinoGeek:
Fixed a few hardcoded *nix directory separators that would stop `oil` from finding any migrations with glob patterns in Windows.

### DIFF
--- a/tasks/migrate.php
+++ b/tasks/migrate.php
@@ -75,9 +75,9 @@ class Migrate
 				foreach (\Config::get('module_paths') as $path)
 				{
 					// get all modules that have files in the migration folder
-					foreach (glob($path . '*/') as $m)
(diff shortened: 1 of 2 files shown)


tasks/migrate.php
```diff
@@ -100,9 +100,9 @@ public function __construct()
 				// get all packages that have files in the migration folder
 				foreach (\Config::get('package_paths', array(PKGPATH)) as $p)
 				{
-					foreach (glob($p . '*/') as $pp)
+					foreach (glob($pp.'*'.DS) as $pp)
```
frankdejonge: $pp?

AlbinoGeek: Many usages of $p and $pp, sorting through which one is which now.... -_-

PR #1437, opened 2013-05-15
Fixed 'Undefined index: session_id' on line 132 or session redis class

jared2501:
So this is my first ever pull request with github (slash git) and I have no idea what I'm doing, so here goes:

I was receiving an 'undefined index session_id on line 132 of session/redis.php class' error. Traced to to the fact that we can see on line 99:

$this->keys = array();

and then on line 132:

$this->keys['previous_id'] = $this->keys['session_id'];

So therefore we try and access the 'session_id' key of the keys array, despite not setting anything in the keys at all.

Compare this to the session/memcached.php class, where these two lines of code at 132 and 133 aren't there.

I suggest removing them, making the method definition very similar to the memcahced equivalent.

### DIFF
--- a/classes/session/redis.php
+++ b/classes/session/redis.php
@@ -3,7 +3,7 @@
  * Part of the Fuel framework.
  *
  * @package    Fuel
- * @version    1.7
+ * @version    1.6
  * @author     Fuel Development Team
  * @license    MIT License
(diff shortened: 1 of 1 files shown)


frankdejonge:
Was just fixed in another pull request :) thanks!

jared2501:
Oh yeah I see now! Cheers

PR #1438, opened 2013-05-16
Avoid PHP error in rest controller

Art4:
Avoid PHP error if key does not exist.


dmyers:
I think its best to use `Arr::get($array, 'key', false)` in cases like that.

huglester:
@dmyers imho Arr is too heavy in this scenario. Probably isset() is even faster than array_key_exists();

WanWizard:
Agreed. But I can live with array_key_exists(), as it's only a single check, and the difference between that and isset() is minimal.

PR #1444, opened 2013-05-21
Fixed problem where inaccurate server time setting could cause Memcached...

nickautomatic:
... caches to expire unexpectedly

(for expiration times less than 30 days, use a relative expiration time, so cache will expire at the correct time even if server time is incorrect)

Signed-off-by: nickautomatic hello@lightvesselautomatic.org

### DIFF
--- a/classes/cache/storage/memcached.php
+++ b/classes/cache/storage/memcached.php
@@ -237,13 +237,19 @@ class Cache_Storage_Memcached extends \Cache_Storage_Driver
 		$key = $this->_get_key();
 
 		$payload = $this->prep_contents();
-
+		
+		// Calculate relative expiration time (eg. 60s)
+		$expiration = !is_null($this->expiration) ? $this->expiration - time() : 0;
+		
+		// if expiration value is less than 30 days, use relative value, otherwise use unix timestamp:
+		$expiration = $expiration <= 2592000 ? (int) $expiration : (int) $this->expiration;
+		
 		// write it to the memcached server
-		if ($this->memcached->set($key, $payload, ! is_null($this->expiration) ? (int) $this->expiration : 0) === false)
+		if ($this->memcached->set($key, $payload, $expiration) === false)
 		{
 			throw new \FuelException('Memcached returned error code "'.$this->memcached->getResultCode().'" on write. Check your configuration.');
 		}
-
+		
 		// update the index
 		$this->_update_index($key);
 


nickautomatic:
For more background / detail about the issue this tweak fixes, see the discussion here:
http://fuelphp.com/forums/discussion/comment/14304

WanWizard:
@FrenkyNet your opinion? I have mixed feelings, fixing a server configuration error in framework code...

nickautomatic:
For what it's worth, I see this not so much as a fix as (hopefully) an improvement to the robustness of the driver: it took a _lot_ of head-scratching to get to the bottom of this problem (by which point I'd already had to write a custom driver to get around this), but with this tweak the cache would have worked fine even on my servers. (Also, I think it's arguable at what point a slightly inaccurate clock becomes a "server configuration error": as it is, _any_ discrepancy at all between the clock used by Memcached and the clock on the server running Fuel will affect the cache's expiration time).

One other thing to note: my set-up (on which the clock mismatch occurred) is on - as far as I can tell - fairly standardly configured Amazon EC2 virtual servers + Elasticache (why their clocks are wrong I don't know), so it's possible that this problem is not a particularly obscure or uncommon one.

Anyway, just thought I'd elaborate on why I submitted it. If you're still not convinced, no problem!

huglester:
I'd like to comment here, since I'm a big user of cache class :)

I think such an improvement should be added to the core, it will make moving fuel acros different server less headache.

Also I don't see an overhead here so let's hope Frenky will like that too :)

WanWizard:
I understand why you would want this, and I don't really have a problem with the change as it doesn't change functionality.

However, It is against my principles to make the change because it allows you to be "sloppy" with the servers time settings. A server should ALWAYS use NTP and run on time. If not you will have all kinds of issues when communicating with other systems. This is an example, but you'll also run into weird problems with cookies and sessions.

@nickautomatic, you spend a lot of time trying to find the cause of this weird behaviour, you don't want to know the amount of time I have spent trying to fix people's session issues, none of it would have been needed if the server was setup correctly.

nickautomatic:
I totally agree with the general principle that server time should be configured correctly. However, in practice not everyone is aware of (or even perhaps - on cheaper hosting - has access to?) their server's time settings (I'm a fairly experienced web developer and it still, as I said, took me a painful amount of time to get to the bottom of this one...)

Also, the change I've made is not in itself a hack to allow sloppy server time settings per se: it just sets expiry time on Memcached in the alternative way (of only two possible ways) - ie. using a relative rather than absolute time for shorter expiry times - and thereby avoids a serious problem where server time _is_ set wrongly while leaving behaviour unchanged when everything is configured correctly.

I agree that hacks are generally to be avoided, but if there are 2 possible ways of doing something, and one of those ways avoids one extra problem at no real extra cost, surely that's the better option? :)

PR #1446, opened 2013-05-22
Check for HTTP_X_FORWARDED_PROTO in Input::protocol

tomschlick:


frankdejonge:
@tomschlick don't we need HTTP_X_FORWARDED_PORT here as well? To be complete. Can you add that too?

WanWizard:
I want to look at this, these headers are easily spoofed.

WanWizard:
p.s. Just noticed that somewhere along the line other X-headers have been sneaked into the Input class. Altough possible with less impact, I think needs to be reviewed, as I see it as a security risk.

As to this header, we use a custom header that our reverse proxy injects, and which is checked in the virtualhost definition of the webserver, which will turn it back into HTTPS if required. This mechanism gives you full control, and is completely transparent for the web application.

WanWizard:
HTTP_X_FORWARDED_PORT is a non-standard header. Who uses that, and why?

WanWizard:
I'll add a config key to allow X- headers to be included in checks, so you can enable them when you know you're behind a proxy that injects them.

PR #1457, opened 2013-05-31
replace non-alphanum characters before checking for numeric keys

benharold:
As discussed in a previous pull request: https://github.com/fuel/core/pull/1405 as well as forum post: http://fuelphp.com/forums/discussion/12119/best-way-to-avoid-invalid-xml-formatting-in-api-output


frankdejonge:
@WanWizard looks good to me, check?

PR #1465, opened 2013-06-20
Reusing one memcached connection

tomschlick:
instead of making 1 connection per Cache::get/set/delete call this saves the
memcached connection as a static variable to avoid DOSing your memcached
server if you have many cache calls on the same page

### DIFF
--- a/classes/cache/storage/memcached.php
+++ b/classes/cache/storage/memcached.php
@@ -30,7 +30,7 @@ class Cache_Storage_Memcached extends \Cache_Storage_Driver
 	/*
 	 * @var  Memcached  storage for the memcached object
 	 */
-	protected $memcached = false;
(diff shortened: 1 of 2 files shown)


dmyers:
I'm kind of surprised `Cache` doesn't have an `instance()` method and work off that as I would imagine this would affect the other cache storage classes as well.

WanWizard:
The best solution (for the future) would be to separate the two, and have specific backend storage drivers, like you have with DB (where indeed a common instance is shared amongst everything using that backend).

WanWizard:
@tomschlick can you do the same for Redis?

tomschlick:
Updated Redis to do the same.

WanWizard:
Thanks!

PR #1466, opened 2013-06-26
Add more ways of getting the CSRF token depending on the Request Method

abread:
I'm developing a RESTful API and to use the methods PUT and DELETE with CSRF Autoload feature I needed this little "hack".
If Security::check_token() fails to get a token from Input::post(), Input::json() and the optional parameter it will try again with Input::put() if "PUT" is the request method, it will try again with Input::delete() if the request method is "DELETE" as well.

Just sharing this to stop others from over thinking the "CSRF verification failed" error when they just need something like this.

### DIFF
--- a/classes/security.php
+++ b/classes/security.php
@@ -265,17 +265,7 @@ class Security
 	 */
 	public static function check_token($value = null)
 	{
-		$value = $value ?: \Input::post(static::$csrf_token_key, \Input::json(static::$csrf_token_key, 'fail'));
-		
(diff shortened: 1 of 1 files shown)


WanWizard:
This PR can not be accepted. Please have a look at the coding standards in the documentation.

WanWizard:
Oops, missed your update of the PR. Sorry...

abread:
It's fine, I took almost two weeks to update it after the suggestions you sent... You took just a day.

PR #1470, opened 2013-07-03
Arr Class add method, insert value(s) into a random iteration array

takecore:
Insert value(s) into a random iteration array

### DIFF
--- a/classes/arr.php
+++ b/classes/arr.php
@@ -552,18 +552,11 @@ class Arr
 	 */
 	public static function insert_random(array &$original, $values)
 	{
-		if (is_array($values))
-		{
-			foreach ($values as $value)
-			{
-				$offset = array_rand($original);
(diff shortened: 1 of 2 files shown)


WanWizard:
@FrenkyNet ?

frankdejonge:
I don't really see the use-case. The `$values` parameter could be type-casted so there's no double code. So if that's fixed it can go in when there are docs supplied.

takecore:
@FrenkyNet 
I'm sorry do not write the use-case.

here is a use-case example.

I has 2 models which Model_Article and Model_Advertisement.
I want to display both an array.
and insert Advertisements array to Articles array irregularly.
so, one foreach, then display both contents.

thanks reading.

hackoh:
@takecore 
I think that this is a very exceptional use-case.
It is not the responsibility range of the Core.

PR #1471, opened 2013-07-03
Added new csrf method to form class

dmyers:
Docs: https://github.com/fuel/docs/pull/566

What do you think?


frankdejonge:
@dmyers code looks good, but the docs are a bit confusing. It _is_ a static method and the param section should be removed. Also the description is for another method. Copy-Paste error? Can you fix this?

PR #1476, opened 2013-07-09
add \Str::random('uid')

benharold:
Add random UID generation to string class.

http://en.wikipedia.org/wiki/Universally_unique_identifier#Definition

### DIFF
--- a/classes/str.php
+++ b/classes/str.php
@@ -308,11 +308,13 @@ class Str
 				break;
 
 			case 'uid':
-				return sprintf('%s-%s-%s-%s-%s',
+			    $pool = array('8', '9', 'a', 'b');
+				return sprintf('%s-%s-4%s-%s%s-%s',
 					static::random('hexdec', 8),
 					static::random('hexdec', 4),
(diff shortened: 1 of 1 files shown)


frankdejonge:
@benharold looks good, could you also update the docs?

Art4:
I would prefer to call it with "uuid" (see RFC 4122 https://tools.ietf.org/html/rfc4122)

``` php
\Str::random('uuid');
```

huglester:
Good point at uuid

WanWizard:
Agree with @Art4. If changed and docs are present we can merge it.

benharold:
After reviewing Wikipedia and the RFC, it looks like the code needs a little tweaking to conform to the standard. I'll post an update with docs when I get a chance.

frankdejonge:
@benharold thanks, we'll keep an eye out for it.

WanWizard:
Can you change this to UUID as requested, and send a PR to update the docs?

benharold:
Done!

PR #1480, opened 2013-07-11
Correct assignment of $valid_login from auth response

jverb:
Corrects the 'if' condition that was improperly assigning the result of the instanceof check to the $valid_login variable when a boolean was returned from the auth function and not the return boolean itself.


WanWizard:
This is already corrected in 1.7/develop. We do not accept PR's on master branches, these are release versions.

PR #1482, opened 2013-07-12
Changed default image alt attribute to be empty instead of filename

dmyers:
What do you guys think of this? I can't really see the benefit of setting the alt to the filename.

Here's an example that was bugging me in my app when the image didn't exist.

screen shot 2013-07-11 at 6 59 25 pm

### DIFF
--- a/classes/html.php
+++ b/classes/html.php
@@ -78,7 +78,7 @@ class Html
 			$src = \Uri::base(false).$src;
 		}
 		$attr['src'] = $src;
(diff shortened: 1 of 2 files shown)


huglester:
Agree on this. no point setting alt to filename, empty alt is far better.

WanWizard:
The idea behind this is that an empty or missing alt attribute is a bad idea. It is meant to contain something useful for people that can't see the image itself, either because they use a renderer that doesn't show the image, or because they use a screen reader (for the blind). Defaulting to the filename is better then no alt value.

An exception to this are decorative images which don't add to the content, for those you can explicitly pass an empty string as alt text (in which case the filename will not be used).

PR #1489, opened 2013-07-19
Change method visibility from private to public on useful methods

benswinburne:
Having attempted to extend this class to add some bespoke functionality, not having access to these methods has proven unhelpful. Is there a reason they're private?

None of the other drivers contain private methods.


WanWizard:
Methods should never be private, as it makes extending a nightmare. Image was a third-party addition, perhaps that is why it slipped through.

dmyers:
:+1:

PR #1495, opened 2013-07-31
[Cache] Redis class is in Fuel\Core, not \

romaintb:
this makes Redis backend for Cache work on php54 (also tested on php53)

I don't really know why it even worked on php53

### DIFF
--- a/classes/cache/storage/redis.php
+++ b/classes/cache/storage/redis.php
@@ -57,7 +57,7 @@ class Cache_Storage_Redis extends \Cache_Storage_Driver
 			// get the redis database instance
 			try
 			{
-				static::$redis = \Redis::instance($this->config['database']);
+				static::$redis = Redis::instance($this->config['database']);
 			}
 			catch (\Exception $e)
 			{


WanWizard:
All Fuel\Core classes are aliased to global to make them extendible. If it doesn't work for you, you have another problem. 

Which is not 5.4, I've been developing on that for ages, now 5.5, and still no problem.

You may have a PHP extension loaded that defines the Redis class, in which case the alias fails...

romaintb:
Ok, uninstalled php54-redis and it works, the thing is if you have php-redis installed, the official Redis class is picked up as \Redis.
This can be a problem for a lot of us, and I am impressed no one ever hit that.

WanWizard:
We know that, and we've had a few questions about it.

I think I've implemented a work-around. I've renamed the class to Redis_Db, changed the cache and session driver to use that, and aliased Redis_Db to Redis if the PECL extension isn't loaded.

This should keep existing applications that use Redis (and therefore not have the extensions loaded) running, and allow people that do need the PECL extension to still use the cache and session driver.

romaintb:
works perfectly ever since

WanWizard:
Thanks for the feedback.

romaintb:
thank you for your time :)

PR #1496, opened 2013-08-01
Method get_path() added

Art4:


Art4:
This is helpful for different use cases, i.e. attaching the file to an email

``` php
$email->attach( $file->get_path() );
```

or get file infos in dev

``` php
print_r File::file_info( $file->get_path() );
```

frankdejonge:
Looks good, can you ad docs for this as well?

PR #1498, opened 2013-08-05
Backup date default timezone before using date_default_timezone_set in Date unit test.

fbnfgc:
Forcing date default timezone in unit test is not a good idea because it is breaking environment configuration in other unit tests.
This commit reverts original date default timezone before exiting the function test_format in Fuel\Core\Test_Date.


WanWizard:
Agreed, that should never be done like that, tests should not change the environment.

PR #1507, opened 2013-09-03
Update record on DB Session write

andrewwheal:
If you have a shutdown handler which wants to write to the session then you get problems with duplicate `session_id`s when using the DB session driver because the record was not being set and it was trying to create a duplicate session record in the DB

### DIFF
--- a/classes/session/db.php
+++ b/classes/session/db.php
@@ -185,6 +185,9 @@ class Session_Db extends \Session_Driver
 			{
 				// then update the cookie
 				$this->_set_cookie(array($this->keys['session_id']));
+
+				// and update the record
+				$this->record = \DB::select()->where('session_id', '=', $session['session_id'])->from($this->config['table'])->execute($this->config['database']);
 			}
 			else
 			{


frankdejonge:
@SicoAnimal is there a reason that you're not listening to the shotdown event provided by the Event class?

andrewwheal:
We use this code `\Event::register('shutdown', array($this, 'save'));` is there a better way?

WanWizard:
Why have a separate handler? All sessions are processed/updated via the shutdown event, that is already in place?

andrewwheal:
We have a shutdown handler which saves a e-commerce basket to the DB and then adds the ID of the quote to the session so the we can retrieve it in the next request. We use a shutdown handler so that we can manipulate the quote where ever needed and know that it will always be saved at the end of the request.

We run the below code to set the required information to the session:

```
\Session::set("basket.{$this->_name}.quote_id", $this->quote()->id);
\Session::write();
```

The write was necessary because the session was already being written before our code.

UPDATE: everything worked fine when using the cookie driver

WanWizard:
So the actual issue is about execution order? Do you autoload the session, or start it manually?

For this reason I've implemented reverse execution of shutdown events in the past, to make sure the event defined first is executed last, exactly to avoid the issue of the session being closed too early. 

It looks like this mechanism was removed somewhere when the event class was refactored, because I can't find that code anymore in the 1.7 codebase...

WanWizard:
Correction: they are reversed, the logic for that has been moved to the instance, and controlled by the 4th parameter of trigger(). 

Perhaps you can check in which sequence they are executed, and also check when the session event is registered, and when you register your shutdown event...

andrewwheal:
I put some logs in on our live site, so below is the order things are executed:

```
ERROR - 2013-09-03 11:10:36 --> whealer - Register Shutdown: Session
ERROR - 2013-09-03 11:10:36 --> whealer - Register Shutdown: Basket
ERROR - 2013-09-03 11:10:39 --> whealer - Fuel\Core\Session_Db::write
ERROR - 2013-09-03 11:10:39 --> whealer - PropShop\Core\Basket::save
ERROR - 2013-09-03 11:10:39 --> whealer - Fuel\Core\Session_Db::write
```

UPDATE: `Basket::save` is logged just before it calls `Session::write()`

WanWizard:
And you autoload the session (always load)?

andrewwheal:
Nope, session is not in the always load config

WanWizard:
Ok, so it's caused by loading the session later then registering the other shutdown events, causing the frameworks' internal shutdown methods to be called after the ones you register in the code.

I'll have to think about it, perhaps we need an "app_finished" event you can hook into, instead of "shutdown".

PR #1510, opened 2013-09-06
Bootstrap 3.0 upgrade

sagikazarmark:
Pagination config upgraded to have `bootstrap3` template.

### DIFF
--- a/config/pagination.php
+++ b/config/pagination.php
@@ -65,7 +65,7 @@ return array(
 	),
 
 	// Twitter bootstrap 3.x template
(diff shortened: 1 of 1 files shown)


sagikazarmark:
Do not merge this! This is just for test and to open a new issue. Oil is not yet finished.
I try to keep the same structure, but there are some modifications in the framework which must be applied.

sagikazarmark:
Ok, OIL is updated, it is working now with BS 3.0. PR added for oil as well.

PR #1511, opened 2013-09-07
Fixed arb char property access error for whitelisted classes

marcdebeer:
I seemed to be receiving an arb exception which raised an error when adding various classes to the security.whitelisted_classes directive. Not entirely sure of the cause, or that it may be related to my file encoding; however adding a trim() to the $k variable on line 229 seems to have resolved it by stripping arbitrary characters.

ErrorException [ Error ]: Cannot access property started with '\0' in COREPATH/classes/security.php @ line 230

fuel-property-access

### DIFF
--- a/classes/security.php
+++ b/classes/security.php
@@ -226,7 +226,7 @@ class Security
 
 			foreach ($value as $k => $v)
 			{
-				$value->{$k} = static::htmlentities($v, $flags, $encoding, $double_encode);
+				$value->{trim($k)} = static::htmlentities($v, $flags, $encoding, $double_encode);
 			}
 		}
 		elseif (is_object($value))


dm:
Watch your whitespace modifications; what's just a one line diff turned into the whole file being changed! :frowning: :-1:

classes/security.php
```diff
@@ -1,412 +1,412 @@
...
+			}
+		}
+		elseif ($value instanceof \Iterator or get_class($value) == 'stdClass')
+		{
+			// Add to $already_cleaned variable
+			$already_cleaned[] = $value;
+
+			foreach ($value as $k => $v)
+			{
+				$value->{trim($k)} = static::htmlentities($v, $flags, $encoding, $double_encode);
```
dm, line 229: The only line that matters in the diff is lost within the whitespace changed... :p

WanWizard:
You should not try to work around an issue in your code by modifying the core.

Your object has illegally named properties, so that is what needs fixing.

marcdebeer:
Apologies for the whitespace... never noticed when comitting! Thanks WanWizard... what properties are you referring to? I've briefly gone through the classes, but can't seem to find any obvious illegal property names.

See http://bin.fuelphp.com/snippet/view/oJ

dm:
Double check your property names, you're using `__get()` and `__set()` and need to make sure that you're verifying both that the property exists; and if using binary names, that they don't begin with `NUL` (nor \* BTW): https://github.com/php/php-src/blob/master/Zend/zend_object_handlers.c#L504

I agree with @WanWizard that this is part of your code being wrong; although the trim being in this assignment wouldn't hurt either ;)

At the very list you should use `trim()` when assigning if you use bin names...

```
$property = trim($property);
$this->$property = $val;
```

WanWizard:
trim() removes whitespace, tabs and a null byte from the beginning and end of a string.

In this case, the string is an object property name. If that contains something that can be trimmed, your in trouble with your object. So make sure such property names can not be created. A trim in the security class would let you get away with sloppy coding, and would waste cpu cycles for those who don't.

marcdebeer:
It appears as though the problem is related to the type of properties in the class (public, private, protected). I added a var_dump($k) just above the line 229 and then changed the type of my $items property in the Navbar_Items class to test:

public $items = array() outputs string(5) "items"
private $items = array() outputs string(25) "\000Admin\Navbar_Items\000items"
protected $items = array() outputs string(8) "\000*\000items"

I've tried doing a trim() in my magic __get() method, but that has no affect at all. By the looks of out, the security class doesn't accommodate private and protected properties.

I'm working around this at the moment by using a public definition for my $items property. But surely this is something that then should be addressed in Fuel? I can't seem to think of any other workaround methods at this point to implement in my code as dm and WanWizard have stated.

Help? Please?

WanWizard:
This looks like an object cast to array (using `(array) $object`), which you can't do.

PR #1521, opened 2013-09-18
Changed to allow replication connection

hackoh:
This change allows replication connection.

When you want to use replication feature, you can set the flag in db.php.

``` php
<?php
return array(
    'default' => array(
        'connection'  => array(
            'dsn'        => 'mysql:host=localhost;dbname=default',
            'username'   => 'dbuser',
            'password'   => 'dbpass',
        ),
    ),
    'db1' => array(
        'type'        => 'pdo',
        'connection'  => array(
            'dsn'        => 'mysql:host=localhost;dbname=db1',
            'username'   => 'dbuser',
            'password'   => 'dbpass',
        ),
        'identifier'   => '`',
        'table_prefix' => '',
        'charset'      => 'utf8',
        'enable_cache' => true,
        'profiling'    => false,
    ),
    'db2' => array(
        'type'        => 'pdo',
        'connection'  => array(
            'dsn'        => 'mysql:host=localhost;dbname=db2',
            'username'   => 'dbuser',
            'password'   => 'dbpass',
        ),
        'identifier'   => '`',
        'table_prefix' => '',
        'charset'      => 'utf8',
        'enable_cache' => true,
        'profiling'    => false,
    ),
    'replication' => array(
        'enable' => true,
        'master' => array('db1'),
        'slave' => array('db1', 'db2'),
    )
);
```

Only Database_Query_Builder_Select uses the slave connection.
And other queries use the master connection.

Respectively, specify the multiple connection name, the destination is a random lottery.

I want your opinion.

### DIFF
--- a/classes/database/connection.php
+++ b/classes/database/connection.php
@@ -18,6 +18,16 @@ namespace Fuel\Core;
 abstract class Database_Connection
 {
 	/**
+	 * @var string Cache of the name of the master connection
+	 */
+	protected static $_replication_master;
+
+	/**
+	 * @var string Cache of the name of the slave connection
+	 */
+	protected static $_replication_slave;
+
+	/**
 	 * @var  array  Database instances
 	 */
 	public static $instances = array();
@@ -33,20 +43,37 @@ abstract class Database_Connection
 	 *     // Create a custom configured instance
 	 *     $db = static::instance('custom', $config);
 	 *
-	 * @param   string $name   instance name
-	 * @param   array  $config configuration parameters
+	 * @param   string $name     instance name
+	 * @param   array  $config   configuration parameters
+	 * @param   bool   $writable when replication is enabled, whether to return the master connection
 	 *
 	 * @return  Database_Connection
 	 *
 	 * @throws \FuelException
 	 */
-	public static function instance($name = null, array $config = null)
+	public static function instance($name = null, array $config = null, $writable = true)
 	{
 		\Config::load('db', true);
 		if ($name === null)
 		{
-			// Use the default instance name
-			$name = \Config::get('db.active');
+			if (\Config::get('db.replication.enable') and ($master = \Config::get('db.replication.master', array())) and ($slave = \Config::get('db.replication.slave', array())))
+			{
+				if ($writable)
+				{
+					is_null(static::$_replication_master) and static::$_replication_master = \Arr::get($master, array_rand($master));
+					$name = static::$_replication_master;
+				}
+				else
+				{
+					is_null(static::$_replication_slave) and static::$_replication_slave = \Arr::get($slave, array_rand($slave));
+					$name = static::$_replication_slave;
+				}
+			}
+			else
+			{
+				// Use the default instance name
+				$name = \Config::get('db.active');
+			}
 		}
 
 		if ( ! isset(static::$instances[$name]))
--- a/classes/database/connection.php
+++ b/classes/database/connection.php
@@ -18,14 +18,9 @@ namespace Fuel\Core;
 abstract class Database_Connection
 {
 	/**
-	 * @var string Cache of the name of the master connection
+	 * @var string Cache of the name of the readonly connection
 	 */
-	protected static $_replication_master;
-
-	/**
-	 * @var string Cache of the name of the slave connection
-	 */
-	protected static $_replication_slave;
+	protected static $_readonly;
 
 	/**
 	 * @var  array  Database instances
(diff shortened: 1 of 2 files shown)


WanWizard:
This requires more thought.

Currently similar behaviour is build into ORM and Model_Crud, and it should be either there or in the low-level driver, but not both. So if this is accepted, ORM and Model_Crud most be looked at, because I'm not sure that functionality can be removed without breaking existing apps.

Furthermore, you could have a lot of database definitions in your db.php, so "mis-using" a definition for something completely different is not a good idea, it should be a per-definition setting, for example:

``` php
<?php
return array(
    'db1' => array(
        'type'        => 'pdo',
        'connection'  => array(
            'dsn'        => 'mysql:host=localhost;dbname=db1',
            'username'   => 'dbuser',
            'password'   => 'dbpass',
        ),
        'identifier'   => '`',
        'table_prefix' => '',
        'charset'      => 'utf8',
        'enable_cache' => true,
        'profiling'    => false,
        'readonly'     => array('db1', 'db2'),
    ),
    'db2' => array(
        'type'        => 'pdo',
        'connection'  => array(
            'dsn'        => 'mysql:host=localhost;dbname=db2',
            'username'   => 'dbuser',
            'password'   => 'dbpass',
        ),
        'identifier'   => '`',
        'table_prefix' => '',
        'charset'      => 'utf8',
        'enable_cache' => true,
        'profiling'    => false,
    ),
);
```

@FrenkyNet, what do you think? We also need something similar in 2.0.

hackoh:
> because I'm not sure that functionality can be removed without breaking existing apps.

Unless "enable" is set as "true", there is no influence in existing applications. 
Moreover, I think that I chose low-risk-way, after reading many codes about DB classes. 
For example, I also considered about "transactions" in the Orm Model.

Only in Database_Query_Builder_Select, connection with the slave DB is made.

hackoh:
@WanWizard 
I changed the config format like your suggest.

frankdejonge:
@hackoh are you on IRC right now? I'd like to discuss this.

hackoh:
@FrenkyNet 
Sure.
But, I cannot do it not much long. 
Please let me know the information about the IRC server.

kenjis:
#fuelphp IRC channel on freenode.

frankdejonge:
@WanWizard FYI, Model_Crud doesn't do any replication logic I think....

hackoh:
@kenjis 
Thank you for your information.

WanWizard:
@kenjis Both Model_Crud and ORM support different connections for read and write, but no "load-balancing" reads over multiple slaves.

@hackoh 

What I meant is that if this will be functionality provided by the low-level driver, it should not be part of the ORM and Model_Crud code, the driver should be configured properly and it should be handled there instead of in two places. So it has to be removed from the ORM Model and Model_Crud.

hackoh:
@WanWizard 
Did you mean about '$_write_connection' and '$_connection' right?
If some connection name specified in Model::$_write_connection (or $_connection) , It uses that preferential.
This means that the high-level layer can overwrite a setup of the low-level driver.

hackoh:
This function operates only when DB connection name is not given.

WanWizard:
Hmm, I'm not sure I agree with that. 

If you setup your application to use a master/slave setup for your database backend, you do that for a specific purpose, which is performance. I'm not sure the application architect would be pleased to see that carefully designed setup is not used because of alternate config at a higher level...

kenjis:
This PR seems to be good, but I also not sure to be merged.

@hackoh 
By the way, why don't you use http://www.php.net/intro.mysqlnd-ms.php ?

hackoh:
@kenjis 
mysqlnd is a driver of concrete more. 
Originally, DB class is not conscious of the type of RBDMS. 
The function of this PR is realized the behavior for the replication by more abstract layer.

hackoh:
@WanWizard 
Understood.
Then, how's this?

1.
When the connection name is specified, load-balancing is carried out according to setting of 'readonly', too. 
Like a following commit.

https://github.com/hackoh/core/commit/e0a7f0d4e520a712565b0aa45d9ea87a68cd59bd

2.
Remove _write_connection from ORM and Model_Crud.

3.
Remove all codes which cares about whether the connection is writable from ORM and Model_Crud.

4.
_connection enables change of a connection place for every model. 
Since it is very useful, so it leaves this.

WanWizard:
Thought about it. Since 1.7 is going to be the last 1.x release, 2.0 has a new database layer and will have a new ORM, I think we should leave the ORM and Model_Crud as-is, to avoid BC issues.

Can this PR be pulled as is? Lost track about with the different changes.

hackoh:
Do you mean that I PR the branch of https://github.com/hackoh/core/commit/e0a7f0d4e520a712565b0aa45d9ea87a68cd59bd ?

PR #1528, opened 2013-09-23
Add subset() method to Arr

Altreus:
We found ourselves needing a version of filter_keys that created them - useful for forms.

Will create docs if there are no objections to the method.

We considered writing it in terms of Arr::merge but we decided to keep it simple for the 90% case of handling input data.

### DIFF
--- a/classes/arr.php
+++ b/classes/arr.php
@@ -1209,7 +1209,7 @@ class Arr
 	}
 
 	/**
-	 * Return the subset of the array by the supplied keys.
+	 * Return the subset of the array defined by the supplied keys.
 	 *
 	 * Returns $default for missing keys, as with Arr::get()
 	 *
--- a/classes/arr.php
+++ b/classes/arr.php
@@ -1219,12 +1219,15 @@ class Arr
 	 *
 	 * @return  array  An array containing the same set of keys provided.
 	 */
-	public function subset($array, $keys, $default=null)
+	public function subset($array, $keys, $default = null)
 	{
 		$ret = [];
-		foreach ($keys as $key) {
+
+		foreach ($keys as $key)
+		{
 			$ret[$key] = \Arr::get($array, $key, $default);
 		}
+
 		return $ret;
 	}
 }
--- a/classes/arr.php
+++ b/classes/arr.php
@@ -1207,4 +1207,24 @@ class Arr
 		// return the value or the key of the array entry the next key points to
 		return $get_value ? $array[$keys[$index+1]] : $keys[$index+1];
 	}
+
+	/**
+	 * Return the subset of the array by the supplied keys.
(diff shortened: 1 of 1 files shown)


classes/arr.php
```diff
@@ -1207,4 +1207,24 @@ public static function next_by_value($array, $value, $get_value = true, $strict
...
+	 * @param   array    $keys     list of keys (or indices) to return
+	 * @param   mixed    $default  value of missing keys; default null
+	 *
+	 * @return  array  An array containing the same set of keys provided.
+	 */
+	public function subset($array, $keys, $default=null)
+	{
+		$ret = [];
+		foreach ($keys as $key) {
+			$ret[$key] = \Arr::get($array, $key, $default);
```
frankdejonge: this should be static:: instead of \Arr

classes/arr.php
```diff
@@ -1207,4 +1207,24 @@ public static function next_by_value($array, $value, $get_value = true, $strict
...
+	 * @param   array    $array    the array containing the values
+	 * @param   array    $keys     list of keys (or indices) to return
+	 * @param   mixed    $default  value of missing keys; default null
+	 *
+	 * @return  array  An array containing the same set of keys provided.
+	 */
+	public function subset($array, $keys, $default=null)
+	{
+		$ret = [];
+		foreach ($keys as $key) {
```
frankdejonge: The bracket should be on a new line and a empty line before the foreach.

classes/arr.php
```diff
@@ -1207,4 +1207,24 @@ public static function next_by_value($array, $value, $get_value = true, $strict
...
+			$ret[$key] = \Arr::get($array, $key, $default);
+		}
+		return $ret;
```
frankdejonge, line 1231: There should be a new line before this return statement.

frankdejonge:
Looks good, can you update your code with the feedback I gave and request a pull for docs?

classes/arr.php
```diff
@@ -1207,4 +1207,24 @@ public static function next_by_value($array, $value, $get_value = true, $strict
...
+	 *
+	 * Returns $default for missing keys, as with Arr::get()
+	 *
+	 * @param   array    $array    the array containing the values
+	 * @param   array    $keys     list of keys (or indices) to return
+	 * @param   mixed    $default  value of missing keys; default null
+	 *
+	 * @return  array  An array containing the same set of keys provided.
+	 */
+	public function subset($array, $keys, $default=null)
```
frankdejonge: $array and $keys should both be typehinted as array and the = should have spaces around it.

Altreus:
Docs PR incoming...

Altreus:
https://github.com/fuel/docs/pull/585

PR #1531, opened 2013-09-25
Module optional bootstrap.php load

sagikazarmark:
I know this is something that has been decided in the past, but as you can see it is optional and I think this could help a lot of people. What do you think?

### DIFF
--- a/classes/module.php
+++ b/classes/module.php
@@ -96,6 +96,12 @@ class Module
 			$ns  => $path.'classes'.DS,
 		), true);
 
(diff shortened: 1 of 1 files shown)


WanWizard:
Forging a finder instance and doing a lookup for a file that in our case would never exist is very expensive, given the fact that our apps have several dozens of modules...

If you want this, extend the Module class and add it to your version of load(), not that complicated. 

So no.

sagikazarmark:
Understood

PR #1532, opened 2013-09-25
In Format::forge()->to_csv(), escape enclosure by a double enclosure

felixgilles:
... not a backslash

First, in the [CSV RFC, section 2.7](http://tools.ietf.org/html/rfc4180#section-2) (it's not a specification of an Internet standard but it documents the format used for Comma-Separated Values)

_If double-quotes are used to enclose fields, then a double-quote appearing inside a field must be escaped by preceding it with another double quote.  For example:_

```
 "aaa","b""bb","ccc"
```

I've tested opening a file generating by Format->to_csv() in Excel 2000 on Windows and Open-office on Ubuntu, in the two cases, columns containing double-quote escaped by a backslash are not well parsed.

Sample:

```
"aaa","b""bb","ccc"
"aaa","b\"bb","ccc"
"ddd","eee","fff"
"ggg","h\"hh","iii"
```

Result in Open-office:

```
aaa    b"bb    ccc             
aaa    b\bb"   ccc"ddd"    eee fff"ggg"    h\hh"   iii
```

First line is OK. Lines 2, 3, 4 are merged.

Result in Excel 2000:

```
aaa    b"bb    ccc
aaa    b\bb"   ccc
ddd    eee     fff
ggg    h\hh"   iii
```

First line is OK. The result is better for lines 2 and 4, but their second column are not well parsed.

### DIFF
--- a/classes/format.php
+++ b/classes/format.php
@@ -193,12 +193,11 @@ class Format
 		$newline = \Config::get('format.csv.newline', "\n");
 		$delimiter or $delimiter = \Config::get('format.csv.delimiter', ',');
 		$enclosure = \Config::get('format.csv.enclosure', '"');
-		$escape = \Config::get('format.csv.escape', '\\');
 
 		// escape function
-		$escaper = function($items) use($enclosure, $escape) {
-			return array_map(function($item) use($enclosure, $escape){
-				return str_replace($enclosure, $escape.$enclosure, $item);
+		$escaper = function($items) use($enclosure) {
+			return array_map(function($item) use($enclosure){
+				return str_replace($enclosure, $enclosure.$enclosure, $item);
 			}, $items);
 		};
 


WanWizard:
The reason for having this configurable is that the escape character heavily depends on the platform used to import the CSV file.  You can change it by simply changing the format config file (format.csv.escape key) to a double quote.

This PR removes that functionality and switches to a hard-coded double quote, and that is a bad idea.

felixgilles:
No, I don't talk about import, but of **export** !

If I change the format.csv.escape config, it will change also for the import, and I don't want to change the import behaviour.
The escaping of enclosure by a double enclosure character is well parsed by all platform (I've test with Open Office on a Mac, same result that in Ubuntu and Windows).

The actual behaviour of to_csv(), with column containing double-quote and default configuration, is not well parsed by majors tools on all OS.

WanWizard:
In that case, it would probably be better to change the config file, so you can configure import and export seperately, because my concern is still valid.

I have an app that exports data that needs to be loaded into a Sybase database, and that requires the backslash as escape character.

felixgilles:
Can you do this test:

```
\Debug::dump(str_getcsv('"aaa","b""bb","ccc"', ',', '"', '\\'));
\Debug::dump(str_getcsv("'aaa','b''bb','ccc'", ',', "'", '\\'));
\Debug::dump(str_getcsv('"aaa","b""bb","ccc"', ',', '"', '/'));
```

Result:

```
Variable #1:
  (Array, 3 elements) ↵
     0 (String): "aaa" (3 characters)
     1 (String): "b"bb" (4 characters)
     2 (String): "ccc" (3 characters)

Variable #1:
  (Array, 3 elements) ↵
     0 (String): "aaa" (3 characters)
     1 (String): "b'bb" (4 characters)
     2 (String): "ccc" (3 characters)

Variable #1:
  (Array, 3 elements) ↵
     0 (String): "aaa" (3 characters)
     1 (String): "b"bb" (4 characters)
     2 (String): "ccc" (3 characters)
```

The native method str_getcsv transforms two consecutive $enclosure characters in one single $enclosure character, whatever the $escape character.

In Sybase export case, the need of escaping with backslash is not for CSV but for Sybase.

WanWizard:
We're discussing export here, from a FuelPHP application, in CSV format, which has to go to an external application. So what get_strcsv() does isn't really relevant, that is about importing a CSV.

Point remains that from_csv() must be able to import both files that use the "" or \" to escape the delimiter. And to_csv() must be able to export both. 

And the current code does that, but defaults to the \ as escape character, which as I understand it is the wrong default in your eyes. The only optimisation I see here is to split the config into an import and export one, so you can define different delimiters and/or escape characters for both operations.

felixgilles:
Yes, I wrote those samples of get_strcsv() to point that the default escaping is two consecutive enclosure characters, and so that the export system has to export with this default behaviour too.

And, I repeat, escape enclosure character with backslash is not well parsed by a major part of softwares handling CSV, and it's one of the main goal of a export in CSV format. It's not just my point of view.

WanWizard:
I get that, its just that changing the default breaks BC, so I have to make a decision whether or not to accept that.

felixgilles:
Another sample

```
$list = array (
    array('aaa', 'b"bb', 'ccc', 'dddd'),
);

$fp = fopen(APPPATH.'cache'.DS.'file.csv', 'w');

foreach ($list as $fields) {
    fputcsv($fp, $fields);
}

fclose($fp);
```

In the file, b"bb becomes "b""bb"

I understand that it's a problem for BC, but the actual behaviour is a bug.

WanWizard:
Now you've lost me. 

What exactly is a bug? If you change the escape character in the format.php config from the backslash to the double quote, it will create exactly that output?

I would not call that a bug, but maybe a poor choice of default value. And I can't say why Phil picked this default when he made the class...

felixgilles:
Excuses me, english is not my natural language, so I'm not clear.

It would be nice that to_csv() works exactly the same that fputcsv(). fputcsv() has not a parameter for escape character, it use two consecutive enclosure. 

Explode format.csv.escape in two key, one for export and one for import, and make double-quote the default value for the export case, is an another way to solve the issue. And it's more flexible that the native fputcsv().

Do you wan't that I modify my PR in this way ?

I suppose Phil chose this default because is the default for str_getcsv(). But it seem that the escape parameter of str_getcsv() is not for espacing the enclosure.

WanWizard:
Your english is a lot better than my french :wink: 

No worries, I already made the changes for a separate import and export config. I only need to decide whether or not to change the default.

felixgilles:
Perfect, a very good synthesis !!

Thanks a lot !

PR #1537, opened 2013-10-01
Add escape tags flag to config of format

hackoh:
Current to_xml() cannot output the node containing tags. 
It is because all nodes are escaped compulsorily. 

This feature allows a user to decide whether to escape tags or not.

If $escape_tags is specified false, nodes containing tags will be outputted with CDATA.

### DIFF
--- a/classes/format.php
+++ b/classes/format.php
@@ -117,9 +117,10 @@ class Format
 	 * @param   mixed        $data
 	 * @param   null         $structure
 	 * @param   null|string  $basenode
+	 * @param   null|bool    whether to escape tags in node
 	 * @return  string
 	 */
-	public function to_xml($data = null, $structure = null, $basenode = null)
+	public function to_xml($data = null, $structure = null, $basenode = null, $escape_tags = null)
 	{
 		if ($data == null)
 		{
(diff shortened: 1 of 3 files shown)


hackoh:
@WanWizard 

> Default value should be false, as that is the current behaviour.

Current behaviour is 'escape all tags'. isn't it?

WanWizard:
The current code doesn't escape anything, so the default behaviour in the new code should be the same.

hackoh:
See this manual.
http://www.php.net/manual/en/simplexmlelement.addchild.php#112204

SimpleXMLElement::addChild escapes automatically "<" and ">" in node to `&lt;` and `&gt;`.

Please do test_to_xml_escape_tags() on current code.

WanWizard:
Ah, with "escaped" you mean encoded. 

I thought you referred to CDATA escaping, which is what it should not do by default. But that is not the case I see.

Can you document this too, so it can be merged?

hackoh:
@WanWizard 

Oh..
I did the big misapprehension. 

That behavior is 'encoding'.
It is not 'escaping', like you say.

So I did change the meaning of this flag to 'whether to use CDATA'.
How do you think?

> Can you document this too, so it can be merged?

I will do, if you approve of the above-mentioned proposal.

WanWizard:
$use_cdata is definetely better.

PR #1540, opened 2013-10-07
Add $format option to \Cli::color for underline

Altreus:
Did it this way because other formats are available in SGR. Didn't fancy trying to merge all the various parts of the SGR escape codes, but this naive approach is heading towards bloat. I expect there's a termcap library or suchlike available for PHP that we could use in future with composer.


emlynwest:
No documentation for this? :wink:

PR #1543, opened 2013-10-10
Fix typo in Upload::register()

kenjis:


WanWizard:
This will break BC.

kenjis:
@WanWizard 

What do you mean?

There is no 'after_validate'.
https://github.com/fuelphp/upload/blob/master/src/Fuel/Upload/Upload.php#L68

In 1.7/develop, `Upload::register('validate', $callback)` causes error.

WanWizard:
ah, sorry, overlooked that. You're absolutely right.

PR #1545, opened 2013-10-15
overhead of calling a array_push function

Onuma0519:
Note: If you use array_push() to add one element to the array it's better to use $array[] = because in that way there is no overhead of calling a function.
http://www.php.net/manual/en/function.array-push.php


frankdejonge:
@Onuma0519 thanks! Merged :+1:

PR #1548, opened 2013-10-16
Allow objects to sanitize themselves

phil-lavin:


phil-lavin:
Related to https://github.com/fuel/orm/pull/310

WanWizard:
Isn't it overly complex to use a callback, instead of just a method in the object?  If the object wants to support callbacks, fine, it can implement it. This will set a closure as sanitizer on every object that comes along?

And finaly, not really Fuel coding standards compliant code...  ;-)

WanWizard:
Noticed this issue is still open. 

As of 1.7 Objects can implement \Sanitization, and when implemented, the Security class will signal the object that it needs to provide sanitized data instead of trying to sanitize the object. This has been implemented for both Orm and Model_Crud.

Issue #1442, opened 2013-05-21
The format class doesn't support XML namespaces.


lhoucine:
am using the Format class to parse a gdata xml response from youtube Api, the response contain many prefixed xml elements like media:thumbnail or media:title ..., 
the issue is that elements are ignored and not returned when using Format::to_array()  in the array result!!

WanWizard:
See http://fuelphp.com/forums/discussion/12195/class-format-ignoring-xml-prefixed-elements for a use case.

hackoh:
How about this?

https://github.com/hackoh/core/commit/8bfe8b8587e5cdb8abaa74ca81805f89bdafcdc6

``` php
// config/app/format.php
array(
    'xml' => array(
        'ignore_namespaces' => true,
    ),
)


// Get namespace attributes
$xml = '<?xml version="1.0" encoding="utf-8"?>
<xml xmlns="http://www.w3.org/2005/Atom" xmlns:media="http://search.yahoo.com/mrss/" xmlns:app="http://www.w3.org/2007/app"><article><title>test</title><app:title>app test</app:title></article></xml>';

$data = Format::forge($xml, 'xml')->to_array();

/*
array(
    '@attributes' => array(
        'xmlns' => 'http://www.w3.org/2005/Atom',
        'xmlns:media' => 'http://search.yahoo.com/mrss/',
        'xmlns:app' => 'http://www.w3.org/2007/app',
    ),
    'article' => array(
        'title' => 'test',
        'app:title' => 'app test',
    )
);
*/

```

WanWizard:
this assumes that the namespace prefix is fixed "xmlns", which is not the case?

hackoh:
are there defferent cases?

WanWizard:
The example in the issue report mentions the namespace "media" for example.

hackoh:
This fixed all namespaced-prefix for example "xmlns:media" and "media:thumbnail"

WanWizard:
Can you verify with the output from http://gdata.youtube.com/feeds/api/videos/HIpQ4ah5XTY?v=2 ?

hackoh:
``` php
/*

array(15) {
  ["@attributes"]=>
  array(5) {
    ["xmlns"]=>
    string(27) "http://www.w3.org/2005/Atom"
    ["xmlns:media"]=>
    string(29) "http://search.yahoo.com/mrss/"
    ["xmlns:gd"]=>
    string(32) "http://schemas.google.com/g/2005"
[...]

PR #1549, opened 2013-10-17
Change for parsing namespaced-xml

hackoh:
PR for https://github.com/fuel/core/issues/1442#issuecomment-26481677

### DIFF
--- a/classes/format.php
+++ b/classes/format.php
@@ -27,11 +27,6 @@ class Format
 {
 
 	/**
-	 * @var  array|mixed  input to convert
-	 */
-	protected $_data = array();
-
-	/**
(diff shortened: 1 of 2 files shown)


hackoh:
I will write this doc, ASAP.

frankdejonge:
@hackoh could you more the vars above the forge function? We always have vars first, then methods.

hackoh:
@FrenkyNet 

Really?

IIRC, static vars are put before static methods, and instance vars are put before instance methods. aren't they?

WanWizard:
Order is:
class constants, static properties, static methods, instance properties, constructor, magic methods, public methods, protected methods (think I have all).

PR #1550, opened 2013-10-17
Improve error reporting for CLI

Altreus:
I added the option for \Debug::backtrace to accept the backtrace to dump, which might help with that pesky thing where it reports bootstrap.php all the time

### DIFF
--- a/classes/file.php
+++ b/classes/file.php
@@ -843,7 +843,7 @@ class File
 
 			if ($delete)
 			{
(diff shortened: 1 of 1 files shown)


Altreus:
Don't accept this yet - for some reason it's possible for a stack frame to have no file

Altreus:
See https://github.com/fuel/core/pull/1634

PR #1559, opened 2013-10-26
Translate validation message to Japanese.

koyhoge:
Why only lang/en? May I add a other language?


WanWizard:
We have decided not to do that, as that would suggest Fuel is supported in that other language, while 99% of all messages the framework produces are hardcoded in English (and no mechanism exists to translate it).

For Fuel v2 everything is translatable, through the https://github.com/fuelphp/translations package. When Fuel v2 reaches beta stage, we will provide an online translation application, and will invite people to volonteer for translator duty. ;)

PR #1561, opened 2013-10-27
Update upload.php

57u:
Changed default path_chmod from 777 to 775

### DIFF
--- a/config/upload.php
+++ b/config/upload.php
@@ -73,7 +73,7 @@ return array(
 	'create_path'		=> true,
 
 	// permissions to be set on the path after creation
-	'path_chmod'		=> 0777,
+	'path_chmod'		=> 0775,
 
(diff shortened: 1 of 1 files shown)


WanWizard:
We don't accept PR's on master branches, only on the latest develop branch.

And why would you change this default? It will make the file not writable for the application in certain installations. If you require a different value, use your application configuration to override the default.

57u:
Hi Harro, 

Yes i realised immediately after doing a pull request, i thought i was on 1.8/dev.

Reason for my change was its a default of 777 - is this the wisest? Even though we can change that in our config, is it the safest way..?

WanWizard:
That is always up for debate. 

Do you want a default that always works out of the box, or do you want a default that doesn't. For the moment, we're in camp A (which is also why cookie sessions are default, since they always work).

PR #1563, opened 2013-10-29
Change the condition of db connection

hackoh:
This change allow you to connect to slave with following query.

``` php
\DB::query('SELECT * FROM `items`')->execute()
```

### DIFF
--- a/classes/database/query.php
+++ b/classes/database/query.php
@@ -52,6 +52,11 @@ class Database_Query
 	protected $_as_object = false;
 
 	/**
+	 * @var  bool  Whether this query will write
+	 */
+	protected $_is_writing = null;
+
+	/**
 	 * Creates a new SQL query of the specified type.
 	 *
 	 * @param string $sql   query string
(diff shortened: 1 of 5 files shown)


WanWizard:
What is the added benefit of this? Where is compile() called with an argument? 

Also, this may only be beneficial when the master/slave are two different DB engines, which is a very unlikely setup for master/slave replication. Do you have a use-case for this?

hackoh:
It is important to have changed the condition of whether to connect with a slave to the following condition. 

``` php

    if (is_null($this->_is_writing))
    {
      $this->_is_writing = ($this->_type != \DB::SELECT and strtoupper(substr(ltrim($this->_sql,'('), 0, 6)) != 'SELECT');
    }
```

So, it allows you to connect to slave with following code.

``` php
\DB::query('SELECT * FROM `items`')->execute();
```

WanWizard:
Isn't this not overly complex?

Why not simply move https://github.com/fuel/core/blob/1.8/develop/classes/database/query.php#L261 up to the beginning of the method, and then use $this->_type to see if it's a SELECT or not?

hackoh:
@WanWizard 

> Why not simply move https://github.com/fuel/core/blob/1.8/develop/classes/database/query.php#L261 up to the beginning of the method, and then use $this->_type to see if it's a SELECT or not?

I worry about following use case.

> Also, this may only be beneficial when the master/slave are two different DB engines, which is a very unlikely setup for master/slave replication. Do you have a use-case for this?

WanWizard:
Closed to to merge conflict. If this is still an issue, please feel free to update the PR.

PR #1569, opened 2013-11-06
Get driver class from the root namespace

sagikazarmark:
Look for the class in the root namespace, so it can be extended and replaced


frankdejonge:
@sagikazarmark strings are already from the root namespace, no need to prefix it. In fact, a lot of times autoloading problems occur when string classnames have namespace separator prefixes.

PR #1576, opened 2013-11-19
Remove format 'html' in Rest controller

kenjis:
In fact, 'html' is not supported.

And Rest controller returns string 'Array' if array is passed on 1.8/develop.

On 1.7/master an error occurs:

```
Fuel\Core\PhpErrorException [ Warning ]:
strpos() expects parameter 1 to be string, array given
DOCROOT/index.php @ line 78
```

### DIFF
--- a/classes/controller/rest.php
+++ b/classes/controller/rest.php
@@ -55,7 +55,6 @@ abstract class Controller_Rest extends \Controller
 		'jsonp'=> 'text/javascript',
 		'serialized' => 'application/vnd.php.serialized',
 		'php' => 'text/plain',
-		'html' => 'text/html',
 		'csv' => 'application/csv',
 	);
 


WanWizard:
Well, it is, but it's dealt with in the Reponse::__toString() method. If you want to return HTML, you should not return an array (how to convert that to HTML?).

The index.php should probably use

```
// Render the output
$response->body((string) $response);
```

to have __toString() do it's magic...

WanWizard:
Implemented: https://github.com/fuel/fuel/commit/4c004fa1b57114fdc580345b81389c283c102a70

kenjis:
I don't want to return HTML.

http://fuelphp.com/docs/general/controllers/rest.html

> Supported Formats
> html - a simple HTML table

But Rest controller does not return HTML table. I thought Rest controller returned HTML table even if I passed an array.

Your fix returns the same result as Fuel 1.6. It returns var_dump() results of an array.

What I want to say is:
- Why var_dump() results?
- Documentation is not match the behavior

WanWizard:
Just read your discussion in IRC, so now I have a bit of context.

I don't know why it is documented like this. I went through the history, but the Rest controller never had logic to generate HTML tables. Perhaps Phil intended to build this in, but never did. 

The var_dump() is there so that if you test your Rest method with a browser, at least you'll see something other then "Array". Making it into an HTML table is going to be very difficult, an array can contain anything.

huglester:
I think there used to be something like: Format::toHtml() or something like this (if I remember correct).
It could be somehow linked to this not implemented feature.

WanWizard:
As far as I know that has never existed either. I can't find any trace of it.

kenjis:
I think nobody needs HTML table output, because there has been no bug report about it.
So fixing Documentation is enough?

WanWizard:
Agreed. 

I decided to leave the html format in the code (I have use-cases where a string return value is used), changed the docs description to a more generic string format (instead of a table), and added a note that if you return an array when your output format is html, you'll get a var_dump() string, because array's can't be cast to string.

PR #1582, opened 2013-11-26
Arr::get can return keys including dots

sagikazarmark:


WanWizard:
@FrenkyNet ok? This needs to be changed in the v2 code too then.

huglester:
probably isset() would be faster by some nano seconds and with same benefit?

sagikazarmark:
What about null values?

huglester:
I am not following something maybe?

``` php

>>> $a = array('a' => 1, null => 3);
>>> isset($a[null]);
true
>>> array_key_exists(null, $a);
true
```

and if I have no null values, both commands return some error from cli

frankdejonge:
@huglester, that's null index not null values, and he's correct.

WanWizard:
$key will never be null there, a null value has been checked and dealt with before you get here.

This PR is about keys, not about values, so I'd have to say I don't understand the remark either.

sagikazarmark:
```
$array = ('some.key' => null);

//returns false
isset($array['some.key']);

//returns true
array_key_exists($array, 'some.key');
```

So the isset check for the array ELEMENT (identified by the key) returns false, if the value is null.

WanWizard:
Hmm... must have been sleeping...

PR #1598, opened 2013-12-12
Apply chmod on every Config_File->save(), remove specific chmod on crypt config

felixgilles:
The specific `chmod` on the `crypt.php` config file is an issue: the name of the crypt config file is not necessarily `crypt.php`.

With moving the `chmod` in `Config_File->save()`, the `chmod` will be executed for crypt config file, but also for all others config files.

### DIFF
--- a/classes/config/file.php
+++ b/classes/config/file.php
@@ -189,7 +189,13 @@ abstract class Config_File implements Config_Interface
 			mkdir($path['dirname'], 0777, true);
 		}
 
-		return \File::update($path['dirname'], $path['basename'], $output);
+		$return = \File::update($path['dirname'], $path['basename'], $output);
+		if ($return)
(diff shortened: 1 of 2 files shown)


WanWizard:
Good find!

Savageman:
This introduced an E_Warning ([ Warning ]: chmod(): Operation not permitted) issue when saving a file we don't own.

Chmod is only allowed for the owner of a file. We could either check for posix_geteuid() == fileowner() or use the @ operator.
I'm not fan of the latter, but this is a native call so I think it's ok. But the first solution relies on the POSIX extension to be enabled (which should be by default but may also be disabled...).

WanWizard:
Fixed.

Savageman:
Thanks :)

PR #1602, opened 2013-12-24
Error concerning session.http_header_name configuration

iturgeon:
The configuration and docs all indicate this is a typeo, 'header_header_name' should be 'http_header_name' here.


InsiteFX:
Fix for Header Error

fuel/core/config/session.php

This needs to be changed from:

'header_header_name' => 'Session-Id',

To:

'http_header_name' => 'Session-Id',

Or it will give a Notice Error!

PR #1610, opened 2014-01-12
Nested transaction feature.

ose1955:
Only first level is the real start/commit. All nested transactions only increment/decrement the transaction depth counter. Rollback resets a counter and rollbacks the transaction.

All rules are implemented in Database_Connection.php. Drivers (pdo, mysql, msqli) only implement pure transaction functions (begin_transaction, commit, rollback).

### DIFF
--- a/classes/database/connection.php
+++ b/classes/database/connection.php
@@ -100,6 +100,18 @@ abstract class Database_Connection
 	protected $_instance;
 
 	/**
+	 *
+	 * @var bool $_in_transation allows transactions
+	 */
+	protected $_in_transaction = false;
+	
+	/**
+	 *
+	 * @var int Transaction nesting depth counter
+	 */
+	protected $_transaction_depth = 0;
+	
+	/**
 	 * @var  resource  Raw server connection
 	 */
 	protected $_connection;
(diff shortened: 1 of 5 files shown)


WanWizard:
This doesn't add much, other then allowing you to repeatingly call start_transaction() without error messages.

If nested transactions were implemented, they should really work. Which means for MySQL (and PostgreSQL), savepoints should be implemented.

WanWizard:
This gives you pointers as to how that works: http://www.yiiframework.com/wiki/38/how-to-use-nested-db-transactions-mysql-5-postgresql/

Note that you have to deal with this in the drivers, not globally like you do now. There could be multiple drivers active at any given point, or multiple instances of the same driver (for example with connections to different databases). Each should have their own tracking of the transaction nesting status...

ose1955:
Dear Harro,
thank you for the detailed answer and useful link.

Indeed, savepoints is an elegant decision. However, it has one disadvantage in the proposed implementation. If, suppose, the SQL query failed on the any nested level, the transaction is rolled back to its savepoint and does not closed. For other levels all is OK and a script can continue to work. Meanwhile, transaction should not be commited in any case as partially completed and real rollback should be taken.
Except of savepoint the algorithm is similar to mine.

> > This doesn't add much, other then allowing you to repeatingly call start_transaction() without error messages.

It also allows to commit all nested levels at once and rollback all nested levels from the failed point. Actually, that is what I await from this feature.

> > Note that you have to deal with this in the drivers, not globally like you do now. There could be multiple drivers active at any given point, or multiple instances of the same driver (for example with connections to different databases). Each should have their own tracking of the transaction nesting status...

I am confused a little bit. When connection with a database is established, the instance of the Database_Connection together with a given driver is created. Every new instance has its own driver that is unchanged for the connection lifetime. So, the nested transaction tracking system can be implemented on the connection level, not driver one. Yes, it is "global" within an instance but that is correct because it is isolated from other instances (and drivers). Besides, this is much easier to implement than repeating almost the same code throughout of drivers.

Regards, Sergey.

WanWizard:
Whether or not a complete rollback to level 0 is needed depends on the code. Even if the current level fails and need to be rolled back, it could well be the higher levels can be committed safely, for example because the code will handle the exception. 

So I don't agree with the assumption that a fail somewhere means you have to rollback everything.

With this change, everything is essentially part of one big translation, so rollback is an all-or-nothing. Which is the same as not maintaining the level at all, but just an 'in-transaction' status.

So I stick to my point of either addressing this how it should be, or leave it as is.

ose1955:
> > Whether or not a complete rollback to level 0 is needed depends on the code. Even if the current level fails and need to be rolled back, it could well be the higher levels can be committed safely, for example because the code will handle the exception. 

Sounds reasonable. Will add savepoints.

WanWizard:
Thanks.

You have to look at rollback() too, that might need an option to rollback the last level, or rollback everything. Also, not every RDBMS supports savepoints, which means in the PDO driver you need to check for mysql or pgsql PDO drivers before using it. Otherwise you'll break the PDO driver when using it with MSSQL (doesn't support nested transactions, they are  NOOP's). Afaik Oracle also supports savepoints.

ose1955:
I have added savepoints. 
- Class Database_connection has three protected methods: `set_savepoint()`, `release_savepoint()` and `rollback_savepoint()`. 
- They returns `null` that means that a driver does not support savepoints.
- If a driver supports savepoint it owerrites these methods, if not - does nothing.
- If savepoint is supported, transaction methods use savepoints to manage nested transactions.
- If savepoint is not supported, methods deal with nested transactions as with one global transaction;
- The `rollback_transaction()` method has an option to rollback the last level or rollback everything.

Please check.

ose1955:
All your remarks have been fixed. Please check.

WanWizard:
Thanks. 

Does this need any updates to the docs before merging? I think the new argument (rollback_all) needs to be documented?

ose1955:
Hello!
I don't think so. The "rollback_all" already commented in  DB::rollback_transaction() and  Database_Connection::rollback_transaction() methods. There are a lot of things that do not documented but well-commented in the code and this is enough, at least, for me;)

WanWizard:
The parameter isn't mentioned in the docs: http://docs.fuelphp.com/classes/database/db.html so I can't merge this yet.

ose1955:
Documentation is fixed. Please check https://github.com/fuel/docs/pull/638

PR #1613, opened 2014-01-15
Improves error logging in Database_MySQLi_Connection class

iankberry:


WanWizard:
I can not merge it like this. 

Besides that, it probably needs the same treatment as in the MySQL driver, the password can get exposed in the error message.

iankberry:
Aside from my MySQL driver issue, is there something else that I should have done differently? Sorry, my first pull request :-)

Also I’m not aware of any MySQL generated errors which include the password, but I could be wrong on that.

WanWizard:
The issue was with the fact that your PR contains a complete branch merge. See https://github.com/fuel/fuel/wiki/Contributing

If you have a SQL syntax error on the connect, PHP returns an error message that contains the password. This issue was already addressed in all other messages related to connecting.

PR #1616, opened 2014-01-20
html_tag should not self-close all tags

jett-dave:
Fixed html_tag to account for cases where there's no content in an element and the element is not a void element as per the HTML specification.  In this case, the element cannot be self-closed and requires a closing tag.

List of valid self-closing tags was pulled from http://www.w3.org/html/wg/drafts/html/master/syntax.html#void-elements


jett-dave:
WanWizard - apologies for this - I've moved the void elements array into the html class and updated the html_tag function accordingly.

PR #1618, opened 2014-01-24
fixed Content-Type and Content-Length for FCGI and NGINX setup

umireon:
\Input::server('Content-Type') should return the value of Content-Type field if available.
However, \Input::server('Content-Type') always returns null with lighttpd and mod_cgi.

In \Input::headers(), \Input::server() is called when getallheaders() is missing.
\Input::server() should be called with an underscored name, but actually with an hyphened name.
For example, \Input::server('Content-Type') is wrong and must be \Input::server('CONTENT_TYPE').
That is as well about Content-Length.

This bug should reproduce for FCGI and NGINX setup or other setups that do not have getallheaders().

Signed-off-by: Udagawa Kaito umireon@gmail.com

### DIFF
--- a/classes/input.php
+++ b/classes/input.php
@@ -504,8 +504,8 @@ class Input
 					$headers[$key] = $value;
 				}
 
-				$value = static::server('CONTENT_TYPE') and $headers['Content-Type'] = $value;
-				$value = static::server('CONTENT_LENGTH') and $headers['Content-Length'] = $value;
+				$value = static::server('Content_Type', static::server('Content-Type')) and $headers['Content-Type'] = $value;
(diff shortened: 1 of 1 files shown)


WanWizard:
Good find! 

Unfortunately, this breaks BC for platforms that do use the dash instead of the underscore. You can use this to capture both. Can you update the PR?

```
$value = static::server('Content_Type', static::server('Content-Type')) and $headers['Content-Type'] = $value;
$value = static::server('Content_Length', static::server('Content-Length')) and $headers['Content-Length'] = $value;
```

umireon:
Thank you for your advice, @WanWizard !

This branch was updated to accept both the Content_Type style and the Content-Type style, as @WanWizard suggested.

WanWizard:
Thanks!

PR #1622, opened 2014-01-27
Htmlentities for row data

pachico:
At representation level, row data should be "htmlentitized" since data could contain html codes.
Example: if you try to log "&currency=EUR", quick profiler will display "¤cy=EUR".


WanWizard:
Implemented in 440a1034d17868c49181ec96832f8d66a7e04372

PR #1632, opened 2014-02-13
Add filesize check to _get() method

hackoh:
Unfortunately, when the size of the cache file is 0 bytes in some trouble, error occurs in the following lines. 

``` php
$payload = fread($handle, filesize($file));
```

At that time, file lock of previous lives, it falls deadlock.

I experienced. ;)

### DIFF
--- a/classes/cache/storage/file.php
+++ b/classes/cache/storage/file.php
@@ -323,6 +323,10 @@ class Cache_Storage_File extends \Cache_Storage_Driver
 		{
 			return false;
 		}
+		if ( ! filesize($file))
+		{
+			return false;
+		}
 
 		$handle = fopen($file, 'r');
 		if ( ! $handle)


frankdejonge:
@hackoh when does this happen? Every session payload is serialised in a containing object, so the contents of a cache file (in theory) can never be empty. @WanWizard, right?

hackoh:
@FrenkyNet 

``` php
protected function _set()
    {
        ...

        // wait for a lock
        while ( ! flock($handle, LOCK_EX));

        // truncate the file
        ftruncate($handle, 0);

        // write the session data

                // maybe, HERE!
        fwrite($handle, $payload);

        //release the lock
        flock($handle, LOCK_UN);

        // close the file
        fclose($handle);

        // set the correct rights on the file
        chmod($file, \Config::get('file.chmod.files', 0666));

        return true;
    }
```

hackoh:
But, I could not track error line yet.

hackoh:
Points:
- filesize was 0 byte
- permission was 0644

So, the error occured between following lines.

``` php
 ftruncate($handle, 0);
```

``` php
chmod($file, \Config::get('file.chmod.files', 0666));
```

WanWizard:
@FrenkyNet Cache data is serialized, and prefixed with a metadata string. So the payload can never be zero length.

@hackoh can you reproduce it? if so, any idea what the return values are of ftruncate/fwrite/flock/fclose?

hackoh:
@WanWizard 
Sorry, I can not reproduce it.
Because, I have noticed that the file is 0 byte unawares. 
Other singular points, it is that the permission of the file was 644. 

I think that there are two problems. 

```
1: 0 byte cache file may be generated. 
(It is unknown whether it is a problem of logic and whether it is an external factor) 
```

```
2: Then, the 2nd argument of fwrite will be 0 and the error occurred.
So, the file lock is not unlocked.
```

I think that "1:" cannot happen in case of a normal system.

WanWizard:
The second argument of fwrite() is the payload, which can (or should) never be "" ?

hackoh:
@WanWizard 
Oh, sorry.

I mean that the problem "2:" happened at "fread" in _get() method.
sorry for my typo..

PR #1640, opened 2014-02-21
Fixed a bug when running PHPUnit with views being loaded multiple times.

drewsonne:
When running phpunit tests and creating a controller, if I use a dataProvider and call the same controller again and again with different data, I get errors ("Cannot redeclare my_function_name()") with functions defined in the global scale.

### DIFF
--- a/classes/view.php
+++ b/classes/view.php
@@ -237,7 +237,7 @@ class View
 			try
 			{
 				// Load the view within the current scope
-				include $__file_name;
+				include_once $__file_name;
 			}
 			catch (\Exception $e)
 			{


WanWizard:
That include just loads the View file, that should not contain code?

drewsonne:
The project I'm building test cases for has code in there as view helpers. I appreciate that by FuelPHP view docs there's no mention of class/function declaration in the views, but they're being used as view helpers, to avoid copy/paste code associated specifically with the presentation layer.

WanWizard:
We believe code in a view is bad practice. 

If you need generic helpers, you can define them outside the views, either as static class methods, or as traditional functions. If they are specific helpers for that view, you should move the processing to a Viewmodel (Presenter pattern), and abstract it from the View. Ideally, it should only contain HTML, and some simple if or loop structures.

pendexgabo:
I agree with @WanWizard . The separation between views and anything else is pretty clear (and the view implementation is clean)...so if your patch is applied you won't be able to reuse (AKA `forge` again) such view since was already included...

PR #1642, opened 2014-02-27
Only look in child fieldsets when $flatten is true

Altreus:
`$flatten` appears to decide whether we should recursively treat child fieldsets' fields as our own; the behaviour of `$fieldset->field('name')` should be the same as `$fieldset->field()['name']` - which is only valid for newer PHPs.

### DIFF
--- a/classes/fieldset.php
+++ b/classes/fieldset.php
@@ -412,8 +412,13 @@ class Fieldset
 			return $fields;
 		}
 
-		if ( ! array_key_exists($name, $this->fields) and $flatten)
+		if ( ! array_key_exists($name, $this->fields))
 		{
+			if ( ! $flatten)
+			{
+				return false;
+			}
+
 			foreach ($this->fieldset_children as $fieldset)
 			{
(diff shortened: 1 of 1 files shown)


emlynwest:
What exactly is the issue here?

Altreus:
If you add a child fieldset with a particular field name, say `ids[]`, you can't then add a field with that name to the parent fieldset, because it will eventually do `$this->field('ids[]')`

This is not consistent with `$this->field()['ids[]']`, which would return a different result (i.e. not there), and is a perfectly valid thing to do.

This solution therefore only recurses into child fieldsets when the method call has `$flatten = true`

emlynwest:
When you say "a child fieldset" you do mean a child field? I am just trying to find out what the actual problem here so I can better assess your change.

Altreus:
No I mean a child fieldset

```
$fieldset = \Fieldset::forge('parent');
$child = \Fieldset::forge('child');
$child->add('ids[]', null);
$fieldset->add($child);
$fieldset->add('ids[]', null);  // error
```

Issue #1641, opened 2014-02-22
Uri::main() execution problem on HMVC request


mitsu7788:
\Uri::main() returns the active URI on HMVC context when main request URI is an empty string (i.e. http://mydomain/)

codeChap:
Try: http://fuelphp.com/docs/classes/uri.html#/method_current

WanWizard:
@mitsu7788,

Is it fixed if you change L#155 to `is_null($uri) and $uri = static::string();` ?

mitsu7788:
@codeChap 

Thank you for your suggestion.

@WanWizard  

Yes, my problem has been resolved.

mitsu7788:
Sorry, I found one side effect.

If form action is empty string, Form::open() doesn't generate current url.

my workaround:  form/instance.php

109 > if( ! array_key_exists('action', $attributes) or $attributes['action'] === null)

to

109> if( ! array_key_exists('action', $attributes) or $attributes['action'] === null or $attributes['action'] === '')

PR #1643, opened 2014-03-02
Uri::main() fixing has a side effect

mitsu7788:
I found that Uri::main() amendment #1641 has one side effect.  If form action has an empty string, the generated url gets wrong.


mitsu7788:
Sorry, this might not be side effect.

I have fixed my code to be \Form::open(array('action' => null)) not \Form::open(array('action' => ''))

WanWizard:
It was one. Which is now fixed.

PR #1645, opened 2014-03-05
Added unit test for Router to check for correctly finding controller

abzainuddin:
whether using underscores or namespaces. Two function are
added to ease testing.


WanWizard:
Can you have a look at #1649 ? It looks like your tests fail when using PHPUnit v4.

PR #1646, opened 2014-03-05
Recurse fieldset validation

Altreus:
By doing this, subclasses of Fieldset get the opportunity to
validate themselves from the input, rather than having the parent
do all the work and make possibly false assumptions.

### DIFF
--- a/classes/fieldset.php
+++ b/classes/fieldset.php
@@ -556,12 +556,18 @@ class Fieldset
 	 */
 	public function repopulate()
 	{
-		$fields = $this->field(null, true);
+		$fields = $this->field(null);
 		foreach ($fields as $f)
 		{
 			// Don't repopulate the CSRF field
 			if ($f->name === \Config::get('security.csrf_token_key', 'fuel_csrf_token'))
 			{
+				continue;
+			}
+
+			if ($f instanceof Fieldset)
+			{
+				$f->repopulate();
 				continue;
 			}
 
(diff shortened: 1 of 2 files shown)


Altreus:
Good thing this wasn't merged because there was epic fail in it.

`$errors` was being silently created - should've been `$this->errors` - which meant I didn't notice I'd accidentally inverted the logic in the first place!

I also added a change so that subclasses of Validation can be given the opportunity to isolate the parts of `$input` they actually care about - useful for multidimensional input, where the Validation instance knows which key is relevant.

Altreus:
This may have thrown up more issues so don't merge this yet

Altreus:
Discovered an issue with all of this:

```
$fields = $this->field(null, true, false);
```

This exists in `populate` but not `repopulate`. Validation also doesn't use this third parameter. Not sure why only `populate` uses this but the changes in the other two places now recurse into sub-fieldsets but this one doesn't - and I don't think it can.

So I'm not quite sure where to go from here! I don't think we even use `populate` so I don't have an idea of how to use it or test it.

Ideas welcome

WanWizard:
What is the status on this?

If you don't use populate(), how do you initially populate a fieldset form from an ORM object in case of "edit"? It's what it is there for.

Altreus:
Um it's a bit back-burner at the moment. I'm not sure how we do things without `populate()` - perhaps ORM calls it for us, which is why I can't find it.

Alternatively we do use it, but I've not come across it yet. I'm working on some fieldsets at the moment so I can test the use of `populate()`

WanWizard:
@Altreus ping?

PR #1648, opened 2014-03-06
\Log::_init() - Extended determinig name and location of logfile by new ...

feldsam:
...config "log_filename"

log_filename = 'fuelphp.log' - for loging is used only this one file in log_path dir.
log_filename = null - use original way

I made this extension, because I want use linux logrotation

Signed-off-by: Kristián Feldsam feldsam@gmail.com

### DIFF
--- a/classes/log.php
+++ b/classes/log.php
@@ -51,9 +51,21 @@ class Log
 		try
 		{
 			// determine the name and location of the logfile
-			$rootpath = \Config::get('log_path').date('Y').'/';
-			$filepath = \Config::get('log_path').date('Y/m').'/';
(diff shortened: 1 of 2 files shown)


WanWizard:
This also needs:
- a new entry log_filename in fuel/core/config/config.php (default null)
- updated documentation

As documentation, I would take the approach that you can use this key to define the log filename. If you don't, a filename will be generated (and describe the current mechanism). Best place for this clarification is probably the section on 'output'.

feldsam:
Hello, would you complete the documentation? My english is not best.

WanWizard:
Will do.

PR #1659, opened 2014-03-19
fixing class name comparison in SSH2.php

ghost:
Noticed get_class() is returning 'phpseclib\crypt_rsa' instead of 'crypt_rsa' only.


WanWizard:
Correct, we added the namespace.

PR #1672, opened 2014-04-03
Deleted ! on IF condition to work as expected with default values

jmmendivil:
If start or/and end wildcats are TRUE (like default values) concatenate corresponding chars to create regex string.

### DIFF
--- a/classes/mongo/db.php
+++ b/classes/mongo/db.php
@@ -537,12 +537,12 @@ class Mongo_Db
 		$value = (string) trim($value);
 		$value = quotemeta($value);
 
-		if ($enable_start_wildcard !== TRUE)
+		if ($enable_start_wildcard == TRUE)
 		{
 			$value = '^' . $value;
 		}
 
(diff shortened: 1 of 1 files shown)


WanWizard:
Eh, the comments clearly state:

> If set to anything other than TRUE, a starting line character "^" will be prepended to 
> the search value, representing only searching for a value at the start of a new line

which this what this change negates. So this breaks compatibility.

jmmendivil:
You are right, but the the docs say:

> When set to true the beginning of the string must match

That means it must create a proper regex (append/prepend) if it set to TRUE, besides, the syntax default values and docs are set to TRUE, does it must match by default?.

WanWizard:
I agree there's a discrepancy between code, comments and docs. Question is what should be the truth?

I would suggest changing the code from $enable... to $disable..., so the code and comments are logical again, yet the logic doesn't change. And change the documentation accordingly. 

Any other solution would break all existing code that uses this, which is what I want to avoid at all costs.

PR #1673, opened 2014-04-08
Bugfix - Prevent infinite recursion

joni2back:
When the recursive method list the files to make sure that the folder is empty (list_files), the delete_dir list again the self dir and the parent, and recursion is generated.

Logging:

``` php
function delete_dir($filepath) {
  Logger::log('Deleting $item: ' . $filepath);
  ...
  if ($list !== false and count($list) > 0) {
    foreach ($list as $item) {
       Logger::log('Deleting $item: ' . $item);
  ...
```

Output:

```
[2014-04-08 11:07:37] logger.INFO: Deleting $item: /folder/.. [] []
[2014-04-08 11:07:37] logger.INFO: Deleting $item: /folder/.. [] []
[2014-04-08 11:07:37] logger.INFO: Deleting $item: /folder [] []
[2014-04-08 11:07:37] logger.INFO: Deleting $item: /folder [] []
[2014-04-08 11:07:37] logger.INFO: Deleting $item: /folder/.. [] []
[2014-04-08 11:07:37] logger.INFO: Deleting $item: /folder/.. [] []
[2014-04-08 11:07:37] logger.INFO: Deleting $item: /folder [] []
[2014-04-08 11:07:37] logger.INFO: Deleting $item: /folder [] []
[2014-04-08 11:07:37] logger.INFO: Deleting $item: /folder/.. [] []
[2014-04-08 11:07:37] logger.INFO: Deleting $item: /folder/.. [] []
[2014-04-08 11:07:37] logger.INFO: Deleting $item: /folder [] []
[2014-04-08 11:07:37] logger.INFO: Deleting $item: /folder [] []
[2014-04-08 11:07:37] logger.INFO: Deleting $item: /folder/.. [] []
[2014-04-08 11:07:37] logger.INFO: Deleting $item: /folder/.. [] []
[2014-04-08 11:07:37] logger.INFO: Deleting $item: /folder [] []
```

### DIFF
--- a/classes/ftp.php
+++ b/classes/ftp.php
@@ -453,16 +453,17 @@ class Ftp
 
 		if ($list !== false and count($list) > 0)
 		{
-			if (preg_match('/\/\.\.|\/\.$/', $item)) {
-				continue;
-			}
 			foreach ($list as $item)
 			{
 				// If we can't delete the item it's probaly a folder so
 				// we'll recursively call delete_dir()
 				if ( ! @ftp_delete($this->_conn_id, $item))
 				{
-					$this->delete_dir($item);
+					// don't recurse into current of parent directory
+					if ( ! preg_match('/\/\.\.|\/\.$/', $item)) 
+					{
+						$this->delete_dir($item);
+					}
 				}
 			}
 		}
--- a/classes/ftp.php
+++ b/classes/ftp.php
@@ -453,6 +453,9 @@ class Ftp
 
 		if ($list !== false and count($list) > 0)
 		{
+			if (preg_match('/\/\.\.|\/\.$/', $item)) {
+				continue;
+			}
 			foreach ($list as $item)
 			{
 				// If we can't delete the item it's probaly a folder so


WanWizard:
Interesting, CodeIgniter, which also uses Phil's code, has the same  bug. ;-)

I'm not to happy with the regex through, they tend to be quite expensive, so its probably better to check only before it will start to recurse, so it doesn't run for every item, something like

``` php
// If we can't delete the item it's probaly a folder so
// we'll recursively call delete_dir()
if ( ! @ftp_delete($this->_conn_id, $item))
{
    // don't recurse into current of parent directory
    if ( ! preg_match('/\/\.\.|\/\.$/', $item))
    {
        $this->delete_dir($item);
    }
}
```

Also, please keep the coding standards in mind when you submit a PR, save you from getting rejected...

joni2back:
Sorry for the mistake, the condition should be as you mentioned (in the old commit the $item var is not initialized)

PR #1676, opened 2014-04-12
Adding url_suffix fix for uri to url create method

gubnota:
`strlen($suffix) > 4` for two letter **.tld**, since one symbol plus slash plus point plus two letter tld is **5** (e.g. '**.ru/2**').
I discovered original issue when trying to get Route for one-symbol uri, url_suffix was added only for two symbol.

Imho, It would be better to change checking «whether uri has suffix» logic as:

``` php
if (substr($url, -1) != '/' and (($suffix = \Config::get('url_suffix')) == false or !strrpos($uri,$suffix)))
```

### DIFF
--- a/classes/uri.php
+++ b/classes/uri.php
@@ -167,7 +167,7 @@ class Uri
 		$url .= ltrim($uri, '/');
 
 		// Add a url_suffix if defined and the url doesn't already have one
-		if (substr($url, -1) != '/' and (($suffix = strrchr($url, '.')) === false or strlen($suffix) > 4))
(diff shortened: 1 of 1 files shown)


kenjis:
Why don't you write unit tests?

WanWizard:
Looking at it again, the logic is wrong too. The question there is not "does the URL already have the suffix", the question that needs to be answered is "does the URL have any suffix"?

To be clear, if you have defined ".html" as suffix, something like `Uri::create('test.jpg')` should not result in 'test.jpg.html' which it does with your change...

PR #1677, opened 2014-04-15
Fix the browsercap url. It has moved.

omoon:


WanWizard:
We don't accept PR's on a master branch.

This has already been fixed in develop, expect a 1.7.2 maintenance release within a week which will fix the master too.

omoon:
oh, sorry.

PR #1684, opened 2014-05-11
Fixing "save_info" behavior on themes

Destreyf:
When using save_info it actually writes the file into fuel/app/config/{theme info filename} instead the respected theme's directory, if you look at line 619 (load_info function) you'll notice it passes array($theme) instead of array($theme['name']) this causes an inconsistent behavior between reads and writes.

### DIFF
--- a/classes/theme.php
+++ b/classes/theme.php
@@ -657,7 +657,7 @@ class Theme
 			throw new \ThemeException(sprintf('Could not find theme "%s".', $theme['name']));
 		}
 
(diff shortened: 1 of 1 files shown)


Destreyf:
This also applies to 1.8/develop

kenjis:
You must send PR only to 1.8/develop, or never merged.

See https://github.com/fuel/fuel/blob/1.7/master/CONTRIBUTING.md

WanWizard:
Already fixed in 1.8/develop, so no need to send a PR.

PR #1685, opened 2014-05-11
Fixing "save_info" behavior on themes

Destreyf:
When using save_info it actually writes the file into fuel/app/config/{theme info filename} instead the respected theme's directory, if you look at line 619 (load_info function) you'll notice it passes array($theme) instead of array($theme['name']) this causes an inconsistent behavior between reads and writes.

This is applicable to 1.7.X as well.

### DIFF
--- a/classes/theme.php
+++ b/classes/theme.php
@@ -657,7 +657,7 @@ class Theme
 			throw new \ThemeException(sprintf('Could not find theme "%s".', $theme['name']));
 		}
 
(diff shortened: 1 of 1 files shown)


WanWizard:
This fix is already present, part of a two-year old commit. See https://github.com/fuel/core/blob/1.8/develop/classes/theme.php#L660 I even went back as far as v1.2, and all of them are correct.

On what code are you working?

Destreyf:
1.7, and it even happens on 1.8

PR #1696, opened 2014-06-05
Cli extras

Altreus:
PHP's output buffer doesn't work if you fwrite directly to STDERR and STDOUT, but I wanted to capture the output of scripts I ran with \Oil\Refine.

Also I wanted to not have CSI codes in there if I did that.

### DIFF
--- a/classes/cli.php
+++ b/classes/cli.php
@@ -495,7 +495,7 @@ class Cli
 			static::$STDERR = $fh;
 		}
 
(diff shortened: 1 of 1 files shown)


WanWizard:
Docs ?

Altreus:
Docs! https://github.com/fuel/docs/pull/647

While writing them I realised how much more useful it would be if the methods returned the old resource.

PR #1702, opened 2014-06-08
Update phpquickprofiler.php

mklemarczyk:
Include once directive was added.
Fatal error: Call to undefined function displayPqp() in /var/www/clients/client0/web1/private/fuelphp-1.7.1/fuel/core/vendor/phpquickprofiler/phpquickprofiler.php on line 226


WanWizard:
Closed due to no response.

PR #1703, opened 2014-06-10
Ensure tests clean up any changes made to the environment.

sjamaan:
This is desirable because changes to the environment can influence
the pass/fail status of other tests, which makes them dependent on
the order in which they are executed.  This matters when running
subsets of tests with --group, or when using phpunit-randomizer,
for example.

### DIFF
--- a/tests/debug.php
+++ b/tests/debug.php
@@ -20,12 +20,22 @@ namespace Fuel\Core;
  */
 class Test_Debug extends TestCase
 {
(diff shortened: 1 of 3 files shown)


WanWizard:
Can you close this and send the PR in on 1.8/develop?

sjamaan:
ok, see https://github.com/fuel/core/pull/1706

PR #1705, opened 2014-06-10
fix bug, to show exception message correctly

KenjiOhtsuka:
I modified classes/database/connection.php to show exception message correctly.

Please check and merge.

### DIFF
--- a/classes/database/connection.php
+++ b/classes/database/connection.php
@@ -72,7 +72,7 @@ abstract class Database_Connection
 
 			if ( ! isset($config['type']))
 			{
-				throw new \FuelException('Database type not defined in "{$name}" configuration or "{$name}" configuration does not exist');
+				throw new \FuelException("Database type not defined in \"{$name}\" configuration or \"{$name}\" configuration does not exist");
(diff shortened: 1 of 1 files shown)


WanWizard:
Thanks. But unfortunately the wrong fix, we don't like embedded variables (it looks like old kahana code), we prefer single quotes and concatenation.

KenjiOhtsuka:
Oh, I'm so sorry.

WanWizard:
No problem, you want to change it (using `'text'.$var.'text'` instead, and send it a new PR?

KenjiOhtsuka:
Thank you.
I'll make that code to 

```
'Database type not defined in "'.$name.' configuration or ...
```

And I guess code at line 70

```
$config = \Config::get("db.{$name}");
```

should be

```
$config = \Config::get('db.'.$name);
```

, and also at line 59.

I'll try.

WanWizard:
`'Database type not defined in "'.$name.'" configuration or ...` (missing second double quote)

And yes, that is a good find too!

PR #1710, opened 2014-06-14
modify error message, etc.

KenjiOhtsuka:
I modified strings to show correct error, and so on.

ref
https://github.com/fuel/core/pull/1705


WanWizard:
Thanks!

PR #1712, opened 2014-06-17
Database method list_tables is supported by Database_PDO_Connection

memememomo:


WanWizard:
Does this work / did you test this on EVERY database platform supported by PDO?

memememomo:
Sorry, I considered only MySQL. I understand why list_tables is not supported.

kenjis:
Why don't you run the logic only in case of MySQL?

WanWizard:
You'll get very ugly code. 

It would be a lot better (but maybe a bit pointless in this stage of v1) to create separate PDO based drivers for each of the platforms supported by PDO (Database_PDOMysql_Connection extends Database_PDO_Connection, Database_Sqlite_Connection extends Database_PDO_Connection, Database_PDOMSSQL_Connection extends Database_PDO_Connection, and so on).

kenjis:
@WanWizard 

Ugly code is better than nothing.
What you say is right, but v1 is getting closer to the end. 

The code of v1 PDO is already ugly. Why don't you care some a bit more added?

The method does not work for two years.

Issue #1704, opened 2014-06-10
Database method list_tables() is not supported by PDO driver


Art4:
Using DB::list_tables() with PDO driver fails with an Exception:

`Fuel\Core\FuelException [ Error ]: Database method Fuel\Core\Database_PDO_Connection::list_tables is not supported by Fuel\Core\Database_PDO_Connection`

After a code sight it seems to me [it was never implemented](https://github.com/fuel/core/blob/73d4a0fe64bdaf58d1b92040da99abe846a796dc/classes/database/pdo/connection.php#L311).

WanWizard:
Correct. Which is why it throws an exception. 

And the reason is very simple: the DB actions to perform to list all tables are highly dependent on the RDBMS platform in use. And since the PDO driver is a generic driver, you can not make assumptions about the underlying RDBMS.

If you have a specific target platform in mind you want to access through PDO, extend the PDO driver, and overload the list_tables() method with SQL for that specific platform.

PR #1714, opened 2014-06-18
Add DB::list_tables() for PDO_MYSQL

kenjis:
Related: https://github.com/fuel/core/pull/1712

### DIFF
--- a/classes/database/pdo/connection.php
+++ b/classes/database/pdo/connection.php
@@ -310,7 +310,40 @@ class Database_PDO_Connection extends \Database_Connection
 	 */
 	public function list_tables($like = null)
 	{
+		if (strtolower($this->driver_name()) == 'mysql')
+		{
+			return $this->list_tables_mysql($like);
+		}
+
 		throw new \FuelException('Database method '.__METHOD__.' is not supported by '.__CLASS__);
+	}
+
+	/**
+	 * List tables for PDO_MYSQL
+	 * 
+	 * @param string $like
+	 * @return array
+	 */
+	protected function list_tables_mysql($like = null)
+	{
+		$query = 'SHOW TABLES';
+
+		if (is_string($like))
+		{
+			$query .= ' LIKE ' . $this->quote($like);
+		}
+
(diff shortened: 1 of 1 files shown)


WanWizard:
Still very ugly, and creating a new class that extends isn't a lot of additional effort.

Art4:
@kenjis If the method wasn't implemented for two years it seems to me, nobody needed it before. I agree with @WanWizard.

In my case #1704 I had an unique use for a list of all tables. My first thought was that someone had forgotten to implemend the method. After understanding why this isn't implemented yet I wrote a simple workaround like this one, used it and deleted it.

WanWizard:
The main reasons nothing decent has been implemented up until now are:
- this code came from Kohana, so it takes effort to dive into it and understand it
- to do it right, DBUtil functionality needs to be moved to the driver too (as it is platform dependent code)
- don't forget the majority uses the MySQL drivers, and not the PDO ones

In case of a MySQL platform, I don't see the benefit of using PDO over MySQLi (from a framework user point of view, not from a PHP point of view), and most users will feel the same.

To address this issue, new drivers must be created, preferrably for each of the supported platforms, that start out as "extends Database_PDO_Connection" and nothing more. Then via overloading platform specific solutions can be added for list tables and columns. In a next step, custom Query drivers can be added, for example to work around the lack of support for "OFFSET" and "LIMIT". And in a final step, classes could be added to the drivers to replace the static DBUtil code.

Now, who is going to pick up this task, given the fact that the v2 database layer is already finished?

kenjis:
Creating new classes that extend is a lot of additional effort. :-)

DB::list_tables() is the last one which is not implemented.
v1 database code is legacy, which has no tests. Nobody wants to refactor it now.

So I prefer this ad hoc solution, even if it is not clean. There are already `driver_name()` if statements in the same file.

WanWizard:
Yes, very annoying. Because all others that didn't want to put the effort in...

WanWizard:
Better this than nothing...

PR #1719, opened 2014-06-25
[FIELDSET] Add possibility to build a field without template

jhuriez:
I want to use Fieldset Field for validation, populate, etc. But in some case, i don't want to render the field with the Fieldset template.

### DIFF
--- a/classes/fieldset/field.php
+++ b/classes/fieldset/field.php
@@ -475,7 +475,7 @@ class Fieldset_Field
 	 *
 	 * @return  string
 	 */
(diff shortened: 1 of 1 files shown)


emlynwest:
I'd recommend looking at v2 fieldset/validation as the whole mess of fieldset -> field ->validation has been separated out in v2

WanWizard:
You can disable a fieldset field to avoid it being rendered.

WanWizard:
Functionality already exists, using field()->disable().

PR #1720, opened 2014-06-25
Update crypt.php

sarciszewski:
This is probably pedantic, but every other implementation uses ===


WanWizard:
it is a numeric comparison, and given that === is slighly faster, this change is logical.

PR #1722, opened 2014-06-30
Add Input::is_post()

Onuma0519:
Because the branch of the POST method is commonly used in the application.

### DIFF
--- a/classes/input.php
+++ b/classes/input.php
@@ -540,4 +540,14 @@ class Input
 			static::$put_patch_delete = array();
 		}
 	}
+	
+	/**
+	 * Whether it is POST method
+	 * 
+	 * @return boolean
+	 */
+	public static function is_post()
+	{
+		return static::server('REQUEST_METHOD') === 'POST';
+	}
 }


WanWizard:
What is the added benefit over `if (Input::method() == 'POST')` or `if (Input::post())` ?

For me this doesn't add anything, and while it may address your use case, you can easily argue why not an `is_get()`, which that is far more common?

Onuma0519:
Input :: method() ,  using the Request class. 
I would like to complete in the Input class. 
Input :: post(),  if there are no parameters, it becomes false

WanWizard:
The request class is the only reliable place for the current request method. Your solution isn't. 

If you don't care, you can always extend the Input class and add the method yourself. I can not be added to the framework, as it will not always return the correct request method.

WanWizard:
Problem is that $_SERVER is global. If your main request (the one done by the browser) is a POST, $_SERVER will reflect that, and your is_post() will return true.

However, if your controller then does an HMVC request, that request is a GET (unless you make it a post). And is_post() will still return true, while Input::method() will return 'GET'.

This is what I mean by "not reliable".

PR #1723, opened 2014-07-03
phpseclib upgrade

xrobau:
This upgrades phpseclib to the latest version, and hopefully allows for easy tracking of future changes in the upstream repository.

Changes:
1 - Raw import of upgrade (plus add required licence information, which was missed on the original import)
2 - Add Namespaces
3 - Remove manual loader, rely on autoloader.


WanWizard:
I hate to be a party pooper, but:
- 3 would not have been needed, class_exists() will call the autoloader and returns true if found, so the includes will be never be executed
- the pdkdf2() method is now removed, breaking all Auth

xrobau:
3 is needed - otherwise it tries to load an existing file from DOCROOT, and can't. I was going to manually add __DIR__ to everything that it's trying to include, but I thought relying on the already existing autoloader is good.

The other side is that adding __DIR__ to it means I could submit it back upstream and resolve a potential future  issue there.

Looking at your second item now.

xrobau:
Oh. Duh. class_exists() does not automatically add the current namespace. That's why I had to remove it.

xrobau:
Replaced that old implementaton. Sorry, I didn't see it used anywhere, which is why I didn't put it back in. (Probably should be using Crypt->password anyway)

WanWizard:
It should, but it is used in the Auth package as default password hasher.

WanWizard:
And more issues: https://github.com/fuel/core/issues/1728

WanWizard:
I have tested setPassword(), but it's PBKDF2 implementation isn't compatible with the old one, so it has to be present for backward compatibility.

kenjis:
> 3 - Remove manual loader, rely on autoloader.

causes #1728 

`AES.php` includes `Rijndael.php`. `Rijndael.php` includes `Base.php`. `Base.php` defines constant like `CRYPT_MODE_CTR`.

WanWizard:
Fixed, this was a bug in the latest PHPSecLib implementation. Not all classes make sure Crypt_Base is loaded before the constants are used.

it-can:
Hi I updated a project of mine that uses the \Crypt::decode() \Crypt::encode() functions... But it seems some code has changed in the Crypt class, so now my decoding doesn;t work anymore...
Is this a bug?

codeChap:
What does the error look like?

it-can:
It doesn't decode old encoded data... there is no code error btw

WanWizard:
It seems the PHP5 branch of PHPSecLib is one bit pile of :shit: :-1:  

We'll have to figure out what it is, and see if we fix it or revert it...

xrobau:
I'm happy to run with this, but, being that I don't have any test data to
work with, can you paste some?  Eg ('this should work, but doesn't) and
I'll take ownership of it.

WanWizard:
It can be easily tested: In 1.7.1, do

``` php
// this echo's a base64 string
echo \Crypt::encode('some string');
```

and then in 1.7.2 or 1.8/develop, do

``` php
// this should say 'some string'
echo \Crypt::decode('<output of that echo here>');
```

WanWizard:
This works in <= 1.7.1:

``` php
echo \Crypt::encode('some string'),'<br />';
echo \Crypt::decode('vkEynSr9uZAlE-pAHhyLN3lrZlltMHpieGU3cnFnQ3loS05jODZiSFNMcnJHS28zQ0J3NjNyYWZlNk0');
```

If you do the same in >= 1.7.2, you get

```
MMTqtQEJ-mzfjUo-jB_9VnYzeGN0Q3h0c1hFeHFlZWpHOUJYRnJEdndLM05hdWFxdVdkdnNxbGpvSDg
```

and the decode fails.

note: make sure app/config/crypt.php is the same in both tests!

WanWizard:
And Crypt uses this from PHPSecLib:

``` php
static::$crypter = new Crypt_AES();
static::$hasher = new Crypt_Hash('sha256');
```

What I think the problem might be: since 5.5, PHP has built-in methods for hashing and crypt, and I think PHPSecLib's PHP5 branch uses those if present. 

And I already noticed that PHP's built-in PBKDF2 method (which is used in Auth) returns different output than our software implementation, which was checked and validated by a security company and confirmed as a correct algorithm implementation.

xrobau:
Ok, I can confirm that the same input produces different output - encoding "some string" produces "TgHj6NH--Ac49AEm8vhcMXhTQkJlelBNcFhYOXdhekg0bFJJZzJxS3lCVFF4TjRjcUJNNVhKNDh4dnc" with phpseclib from 1.7.2, whilst I get "N1Y6IyliWBn8f_aTvKCXeDdyYXFtUUtXc09MX0wyQVRERnA2N1pTTm0za3h3MThvaGZwYmFDRHFGQkE" from 1.7.1

Investigating now.

xrobau:
I opened ticket #1739 to track this.

xrobau:
And it's now fixed! yay!

PR #1736, opened 2014-07-15
HttpException namespace fix

taha:
Locally when I start a fresh Fuel project, I got the error below when attempting to visit 404 pages. Turns out, since `HttpNotFoundException` was extended from `\HttpException`, rather than `HttpException`, which is the class from the PHP [HTTP module](http://php.net/manual/en/book.http.php) instead of Fuel's. I attempted to fix that by simply respecting the `Fuel\Core` namespace.

image

### DIFF
--- a/classes/httpexceptions.php
+++ b/classes/httpexceptions.php
@@ -13,7 +13,7 @@
 namespace Fuel\Core;
 
 
-class HttpNotFoundException extends \HttpException
+class HttpNotFoundException extends HttpException
 {
 	public function response()
 	{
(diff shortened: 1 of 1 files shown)


it-can:
Do you have the pecl_http extension active? Can that extension cause this error?

WanWizard:
That is probably the case. Nevertheless, we would not want collisions, so it's not easy to solve.

taha:
@it-can Yeah, it's pecl_http. There's no mention of it in http://pl1.php.net/http, you can find it being used in the example here though http://php.net/manual/en/httprequest.send.php.

Issue #1739, opened 2014-07-30
Regression with PHPSecLib Upgrade


xrobau:
@it-can reported an issue with the output of crypt changing due to the phpseclib upgrade.

I have been able to confirm this with the existing snippet:

```
<?php
include 'bootstrap.php';
$valid = "N1Y6IyliWBn8f_aTvKCXeDdyYXFtUUtXc09MX0wyQVRERnA2N1pTTm0za3h3MThvaGZwYmFDRHFGQkE";
$test = \Crypt::encode('some string');

if ($valid != $test) {
        print "Failed - received $test\n";
} else {
        print "Encode OK\n";
}

$decode = \Crypt::decode($valid);
if ($decode != "some string") {
        print "Decode known valid failed - handed $decode\n";
} else {
        print "Decode Known valid OK\n";
}

$decode = \Crypt::decode($test);
if ($decode != "some string") {
        print "Decode encoded failed - handed $decode\n";
} else {
        print "Decode encoded OK\n";
}
```

The '$valid' string was generated with phpseclib from 1.7.1

xrobau:
As the crypt output is somewhat system dependant (at the moment), this is on CentOS 7 machine, with php 5.4.16.

xrobau:
Ah, ok. I did some playing around with it, and when you SPECIFY A KEY it works.

Replace the two lines above with this, and it works fine on both 1.7.1 and 1.7.2 

```
$test = \Crypt::encode('some string', "thisisthekey");
$valid = "KkeD6KpsK5qbUbWbK8hrNS0tSHYwcFFfRHlGTmZ5bFFTc0pvY0N2MHFhSHNpY0dRQkdwTmNvaEJGVVk";
```

kenjis:
I wrote tests and worked fine on both 1.7.1 and 1.7.2.
I checked PHP 5.3 and 5.5 on Linux 32bit.

xrobau:
I've found the issue, patch coming up.

xrobau:
Ok, pull request https://github.com/fuel/core/pull/1741 fixes it. There's also a link to a really shonky test that demonstrates the issue on 1.7.1 and 1.7.2. Hopefully someone can fix that up and stick it into fuel properly.

xrobau:
Updated my RPM with this fix, too, btw

https://github.com/xrobau/fuelphp

kenjis:
@xrobau Great fix!

WanWizard:
[...]

PR #1740, opened 2014-07-31
Add tests for Crypt

kenjis:
Tests for  #1739


kenjis:
I've updated tests.
I could reproduce the error, and confirmed your fix works fine.

PR #1741, opened 2014-07-31
This appears to fix the encrypt/decrypt issues with the new phpseclib.

xrobau:
I wrote a quick test to confirm this is fixed.. you can see it at http://pastebin.com/ZmiHujmS

I didn't try to covert it to a proper phpunit test, as I haven't ever looked at getting that working with fuel. Hopefully there'll be enough there for someone with a proper build environment to convert it to a sensible test.

Also, it's interesting that the new phpseclib is -massively- faster at encrypting and decrypting.  I actually thought there was an error with my tests when I got it working against 1.7.2, it ran almost instantly, as opposed to a 1 or 2 second delay.

### DIFF
--- a/classes/crypt.php
+++ b/classes/crypt.php
@@ -96,12 +96,25 @@ class Crypt
 	 *
 	 * @param	string	value to encrypt
 	 * @param	string	optional custom key to be used for this encryption
+	 * @param	int	optional key length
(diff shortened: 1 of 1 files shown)


xrobau:
Tagging issue #1739

PR #1756, opened 2014-08-19
Adds replace_keys function to Arr

sagikazarmark:
Adds a replace_key implementation. It is important the order of the array remains the same without any array slicing.

### DIFF
--- a/classes/arr.php
+++ b/classes/arr.php
@@ -1266,4 +1266,30 @@ class Arr
 
 		return $result;
 	}
+
+	/**
+	 * Replace keys in an array and preserve the original order
+	 *
+	 * @param  array  $array the array
+	 * @param  array  $keys  the array containing the keys in old => new format
+	 *
+	 * @return array
+	 */
+	public static function replace_keys(array $array, array $keys)
+	{
+		$all_keys = array_keys($array);
+
+		// Loop all keys
+		foreach ($all_keys as &$key)
+		{
+			// and replace old ones
+			if (array_key_exists($key, $keys))
+			{
+				$key = $keys[$key];
+			}
+		}
+
+		// Recombine the array with the updated keys
+		return array_combine($all_keys, $array);
+	}
 }


it-can:
Please add some tests... :-)

sagikazarmark:
I will, if someone confirms that it will be merged. I will also add docs then. ;)

WanWizard:
A potential problem I see is a key conflict, which causes an overwrite, and a key array that is out of sync with the values.

```
$array = ['a' => 1, 'b' => 2, 'c' => 3];
$array = replace_keys($array, ['a' => 'c', 'c' => 'a']);
```

result?

classes/arr.php
```diff
@@ -1266,4 +1266,30 @@ public static function subset(array $array, array $keys, $default = null)
...
+	 */
+	public static function replace_keys(array $array, array $keys)
+	{
+		$all_keys = array_keys($array);
+
+		// Loop all keys
+		foreach ($all_keys as &$key)
+		{
+			// and replace old ones
+			if (array_key_exists($key, $keys))
```
sagikazarmark, line 1286: @WanWizard Correct. Would a check for new key existence solve it?
`array_key_exists($key, $keys) and ! array_key_exists($keys[$key], $keys)`

WanWizard, line 1286: If you do that, you get incomplete/incorrect results. Given the fact you have an indexed array to start with, perhaps it's better to create a new array?

```
$new_keys = array();
foreach ($all_keys as &$key)
{
    // and replace old ones
    if (array_key_exists($key, $keys))
    {
        $new_keys[] = $keys[$key];
    }
    else
    {
         $new_keys[] = $key;
    }
}
```

WanWizard, line 1286: Scratch that, that fails too, on

```
$array = ['a' => 1, 'b' => 2, 'c' => 3];
$array = replace_keys($array, ['a' => 'c']);
```

WanWizard:
If you can document this with a fat warning that in case of conflicts values are overwritten, and it's the dev's responsibility this doesn't happen, this can be merged.

WanWizard:
@sagikazarmark what to do this one? No tests and docs were ever made?

sagikazarmark:
I don't know. Is it necessary at all? V1 is in feature freeze, isn't it. If you think it can be useful, I can add it to the v2 repo.

WanWizard:
Correct. Just cleaning up old issues for a hopefully final 1.7.3. release. Moved to v2.

PR #1763, opened 2014-08-27
the rules of 'Inflector' are load in config file.

diegoszch:
I removed Inflector's rules of the array to config file, because in latin language the grammatical rules follow different patterns and today I can't change this values.

Thus, the core stayed more flexible.

### DIFF
--- a/classes/inflector.php
+++ b/classes/inflector.php
@@ -3,7 +3,7 @@
  * Part of the Fuel framework.
  *
  * @package    Fuel
- * @version    1.7
+ * @version    1.8
  * @author     Fuel Development Team
  * @license    MIT License
  * @copyright  2010 - 2014 Fuel Development Team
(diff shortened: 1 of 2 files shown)


emlynwest:
How would this work for apps where there are multiple languages? Eg, my app can be in English and Chinese and the user can select which language they want.

diegoszch:
Interesting @stevewest . 
Primarily I wanted remove the rules of the code e put it in config file.

But, do you think better to put the file in lang folder instead of config folder?

emlynwest:
This is something that is dependant on the enabled language so it would be work making it obey the currently active language.

From the docs: `Config::set('language', 'cy');`
http://fuelphp.com/docs/classes/lang.html

I think it is good to make inflector language aware but it should be done in a flexable way that allows for configs for multiple languages, not just one at a time.

diegoszch:
Ok...I'm going to close pull request and change my code to respect these conditions.

Thanks @stevewest.

PR #1764, opened 2014-08-27
Inflector's rules load via Lang

diegoszch:
Now I altered my code for response via Lang file instead of Config file.


WanWizard:
If you remove the defaults from the class, this code goes wrong if there is no lang file available with inflector rules. Which will happen if you don't define it yourself, and you don't use 'en' as active or fallback.

PR #1766, opened 2014-09-10
Adding the dump and die function.

jedkirby:
Signed-off-by: James Kirby jedkirby@me.com

### DIFF
--- a/base.php
+++ b/base.php
@@ -524,3 +524,18 @@ if ( ! function_exists('call_fuel_func_array'))
 		return call_user_func_array($callback, $args);
 	}
 }
+
+/**
+ * Dump passed args and end any further processing
+ *
+ * @param  mixed
+ *
+ * @return void
+ */
+if ( ! function_exists('dd'))
+{
+	function dd()
+	{
+		array_map(function($x) { var_dump($x); }, func_get_args()); die;
+	}
+}


emlynwest:
Not sure this should be in the framework. I don't see what's wrong with `var_dump($a, $b, $c); die();`.

steadweb:
Apart from you're calling one method instead of two. Laravel has it https://github.com/laravel/framework/blob/4b6ff608ee06b53956412ef5e40b291dfe9579f5/src/Illuminate/Support/helpers.php

emlynwest:
To be frank, why should we add it just because laravel has it and to save writing an extra 5 characters?

steadweb:
I was just pointing it out where it came from. It's useful, hence why it's been PR'd.

jedkirby:
Typing 5 characters instead of 15 each time is a no brainer for me..

emlynwest:
Regardless I am not happy for this to be included in the framework in any form. Any decent editor has code templates and imo this also encourages bad development practices, you should be using things like xdebug instead.

WanWizard:
I never use var_dump() anyway, I use `die(\Debug::dump($a, $b, $c));`.

And I have seen applications that define something like this in their app bootstrap:

``` php
function dump()
{
    call_user_func_array('Debug::dump', func_get_args());
    die();
}
```

which reduces the 5 characters to 4... :-)

PR #1771, opened 2014-09-17
Fixed a problem that alpha value becomes zero

TakeshiOnishi:
Fixed a problem that alpha value becomes zero next to transparent when you enter, such as "# FF0000" or "# F00" when you specify the color in the Image class

Signed-off-by: Onishi Takeshi march.hare.come@gmail.com


TakeshiOnishi:
Please do not merge because the wrong argument to be given to the Hexdec

PR #1773, opened 2014-09-23
Only use alpha value from hex if format is rgba

julesjanssen:


kenjis:
Duplicate: #1772 
Which is better?

And would you please read https://github.com/fuel/fuel/wiki/Contributing before sending PR?
You missed sign-off.

julesjanssen:
It is almost a duplicate indeed. I had not seen that PR before implementing my update.
I do think my own is better. #1772 sets alpha to a hex string in the first case, instead of a decimal number which the rest of the method expects.
I hope the missing sign-off isn't a problem. It wasn't with earlier PRs. I will sign off next time.
I do certify what's in _Developer's Certificate of Origin 1.1_

WanWizard:
I agree, that bit of code extracts decimal values from the hex string, setting it to the string 'FF' is going to fail (or silently converted to zero, which I consider a fail too).

PR #1777, opened 2014-10-06
Fix global namespaces style

it-can:
in addition to #1734


emlynwest:
Thanks! It's good to start standardising the way the global classes are used.

it-can:
glad to help

PR #1779, opened 2014-10-07
Strict match collection test

it-can:
Added test for 07cc8f59399b3741e61c31cf39474cfe444886fb

I hope this test is correct, I am pretty new at testing :-)


emlynwest:
Tests are always welcome, thanks!

PR #1781, opened 2014-10-09
Model_Crud::count() bug fix

Onuma0519:
http://fuelphp.jp/docs/1.7/classes/model_crud/methods.html#/method_count
where is error.

### DIFF
--- a/classes/model/crud.php
+++ b/classes/model/crud.php
@@ -279,7 +279,7 @@ class Model_Crud extends \Model implements \Iterator, \ArrayAccess, \Serializabl
 			{
 				throw new \FuelException(get_called_class().'::count where statement must be an array or a closure.');
 			}
-			$query = $query->where($where);
+			$query = $query->where(array($where));
 		}
 
 		if ( ! empty($group_by))


emlynwest:
What exactly is the problem that this "fixes"?

Onuma0519:
In the same code as the document, it is an error.

$nbr_of_records = Model_User::count('id', true, array('group', '=', 100));
↓
This is can run successfully
$nbr_of_records = Model_User::count('id', true, array(array('group', '=', 100)));

emlynwest:
@WanWizard Model_Crud is more your area, thoughts?

WanWizard:
Who? Me? Never used it. :-)

Looking at the purpose and the code, I'd say the docs are incorrect. You would need this code to be able to do:

```
$nbr_of_records = Model_User::count(
    'id', 
    true, 
    array(array('group', '=', 100),array('field', '=', 'value'))
);
```

emlynwest:
Indeed does seem to be the docs then, so this PR can be closed.

Issue #1784, opened 2014-10-11
Validation "valid_string" alpha is not working with special chars


dercoder:
I am trying to validate customers last and first name so i tried this code

```
$validation->add('name', 'Name')->add_rule('valid_string', 'alpha');
```

will generate this

```
preg_match('/^([a-zA-Z])+$/u', $string);
```

But this will not work for German names like "Maierhöfer"

This Example is not working

```
if (preg_match('/^([a-zA-Z])+$/u', 'Maierhöfer')) {
    echo 'Valid';
}
```

This Example is working

```
if (preg_match('/^[[:alpha:]]+$/u', 'Maierhöfer')) {
    echo 'Valid';
}
```

emlynwest:
Valid string's `alpha` rule only includes `a-zA-Z` you would have to add extra characters as a regex as part of the valid string's conditions.

dercoder:
Ok but why is "('valid_string', 'alpha')" = "a-zA-Z" and not "[[:alpha:]]"?

emlynwest:
My guess would be that the original author did not know about such a thing. Both do the same job so I don't see the issue.

dercoder:
Its not the same because in "a-zA-Z" such letters like "äöüß" are not allowed but in "[[:alpha:]]" they are

emlynwest:
According to the little documentation I can find after a quick google `[[:alpha:]]` only includes `a-zA-Z`. But that is not the issue, changing the behaviour of this would break backwards compatibility, it would be better if a new rule was added that includes `a-zA-Z` and accented characters.

dercoder:
All english sites describes that `[[:alpha:]]` only includes `a-zA-Z` but on german sites they describe that it also includes other special letters like `äöü`. 

Some if them describes that this depends on setlocales but thats not true. I have installed and uninstalled many locales on my test server to detect if there is some problem when you change the locales but its working fine. So locales has no effect on this.

Its also working with arabic and turkish words like  `ستخدم` and `şifre`

I can understand backwards compatibility must be given but it would be nice to have a new rule

emlynwest:
Ah curious, that's good to know. Thank you for the info.

A new rule would be good, I'll reopen this issue then and either yourself or someone else can implement it. It should just be a case of adding the extra handling into the validation class. https://github.com/fuel/core/blob/1.8/develop/classes/validation.php#L892

dercoder:
[...]

PR #1785, opened 2014-10-11
Added "specials" into Validation

dercoder:
This validation will allow specials letters in each language like described here in this Issue https://github.com/fuel/core/issues/1784


emlynwest:
Looks good, thanks. Would you mind adding the new rule to to the docs as well please?

emlynwest:
Closes #1784

PR #1786, opened 2014-10-18
Update XML boolean format.

aaharu:
## XML Schema

http://www.w3.org/TR/xmlschema-2/#boolean

> An instance of a datatype that is defined as ·boolean· can have the following legal literals {true, false, 1, 0}.
## FuelPHP behaviour

``` php
array('true' => true, 'false' => false)
```
### current format

``` xml
<xml><true>1</true><false/></xml>
```

`false` is formatted to empty string.
It's PHP type casting works.
http://php.net/manual/en/language.types.string.php#language.types.string.casting
### new format

I add `bool_representation` option to change XML outputs.

If `bool_representation` is 
`true`:

``` xml
<xml><true>true</true><false>false</false></xml>
```

`1`:

``` xml
<xml><true>1</true><false>0</false></xml>
```

false value, such as `null`:

``` xml
<xml><true>1</true><false/></xml>
```

Signed-off-by: Haruaki Otake aaharu@hotmail.com

### DIFF
--- a/classes/format.php
+++ b/classes/format.php
@@ -136,9 +136,10 @@ class Format
 	 * @param   null         $structure
 	 * @param   null|string  $basenode
 	 * @param   null|bool    whether to use CDATA in nodes
+	 * @param   mixed        if true, element values are true/false. if 1, 1/0.
 	 * @return  string
 	 */
-	public function to_xml($data = null, $structure = null, $basenode = null, $use_cdata = null)
+	public function to_xml($data = null, $structure = null, $basenode = null, $use_cdata = null, $bool_representation = null)
 	{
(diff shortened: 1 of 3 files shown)


emlynwest:
Would need docs too please.
@WanWizard Thoughts on BC for this?

WanWizard:
Not happy about it, as it breaks BC.

I understand the need for this change, perhaps use the 'null' value of the flag as default, and use that for the current behaviour, and true/false for the new behaviour?

aaharu:
Thanks for comments.
I fix the default.

WanWizard:
Thanks!

PR #1788, opened 2014-10-19
Fix enclose_numbers config

aaharu:
[doc](http://www.fuelphp.com/docs/classes/format.html) and [config](https://github.com/fuel/core/blob/22c047ff7a9418a7043f41dfba8611c98f174797/config/format.php#L37) are `enclose_numbers`.
But, `delimit_numbers` in `to_csv()`.

Signed-off-by: Haruaki Otake aaharu@hotmail.com


WanWizard:
Very well spotted!

Issue #1791, opened 2014-11-04
Update htmLawed


it-can:
I saw there is a new version of htmLawed...

http://www.bioinformatics.org/phplabware/internal_utilities/htmLawed/

changelog:
http://www.bioinformatics.org/phplabware/internal_utilities/htmLawed/htmLawed_README.htm#s4.3

PR #1793, opened 2014-11-04
update htmLawed to 1.1.18

it-can:
See #1791


WanWizard:
This merge overwrote the changes we made to the library (which is the reason we don't use it from an external source).

it-can:
you mean the code between line 93 - 129?

WanWizard:
yes. it is needed to deal with microsoft crap.

PR #1796, opened 2014-11-12
fix soap execute & parameters

julesjanssen:
It seems the `__soapCall` method expects the parameters to be wrapped in an array with 'parameters' as its key. It's mentioned in the comments at http://php.net/manual/en/soapclient.soapcall.php

With this update, the Soap service at http://ec.europa.eu/taxation_customs/vies/checkVatService.wsdl worked for me.

### DIFF
--- a/classes/request/soap.php
+++ b/classes/request/soap.php
@@ -128,7 +128,7 @@ class Request_Soap extends \Request_Driver
 		// Execute the request & and hide all output
 		try
 		{
-			$body = $this->connection()->__soapCall($this->function, $this->params, array(), $this->get_headers(), $headers);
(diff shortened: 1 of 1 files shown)


julesjanssen:
Oh, forgot to 'sign-off' again.
I do certify what's in Developer's Certificate of Origin 1.1

WanWizard:
No, it doesn't. The definition of the array depends completely on the SOAP server being called. 

It is therefore coded the way it is to allow you to set a 'parameters' value to make it work for you, while others use other values.

p..s I don't see anything on that PHP docs page that states $arguments should be an associative array with a single key called 'arguments'?

PR #1801, opened 2014-12-11
Removed the need for the validation request method to be post

aidenfoxx:
I was having a problem when using the validate function in a sub-request which used a custom request method.

When creating a new request with a different method, the Input::method() will always return the newly defined method inside that request, rather than GET, POST or ACTION causing this function to return false, making validating impossible.

Removing this solves that problem, and doesn't appear to cause any problems that I can see. Please, correct me if I'm wrong?

### DIFF
--- a/classes/validation.php
+++ b/classes/validation.php
@@ -329,7 +329,7 @@ class Validation
 	 */
 	public function run($input = null, $allow_partial = false, $temp_callables = array())
 	{
-		if (is_null($input) and \Input::method() != 'POST')
+		if (is_null($input))
 		{
 			return false;
 		}


WanWizard:
Input::method() should return the HTTP method used, afaik "ACTION" is not a valid method?

Analyzing run(), I see it calls the input() method to retrieve a field value, and that in turn uses Input::param(), which returns a value independent of the HTTP method used to send the value to the server (it also supports PUT and DELETE for example).

So my initial response would be that this is ok. @stevewest ?

aidenfoxx:
You're correct. ACTION is not a valid HTTP method as far as I'm aware. However, I'm in a funky situation in that under a subrequest Input::method() is returning the custom request method defined when the request is created. This might be the way it's supposed to work (?), but it's causing the validation function to fail under said subrequest.

A simple example of the problem can be seen here: http://pastebin.com/dt4n0ABx

It's a basic form example which validates an email field, however, you'll notice that the subrequest always fails to validate due to Input::method() returning 'custom' and causing the validation->run() function to fail.

emlynwest:
Personally I don't think we should support non-standard HTTP actions as I think this is largely application specific. If you wanted you can grab the data and pass that as the first parameter to `run()` which skips the automatic use of `Input`

aidenfoxx:
I understand the idea behind only validating on HTTP actions, but the way it works seems sort of like a bug to me. It tripped me up when writing my application, as I couldn't for the life of me figure out why my fieldset wouldn't validate. 

I now know I can pass in the first parameter into the `run()` method, but this discrepancy between HTTP actions and the method defined in the request is confusing. The `set_method($method)` on a request changes the `Input::method()`, which I'd assume should only support HTTP actions?

I'm probably just using the `set_method` function incorrectly, but it's something that works for my application (allows me create a module request which isn't accessible from a standard route).

WanWizard:
`set_method()` is meant to set the HTTP method, yes.

I miss the relation between this method and the desire to have a non-routable HMVC request...

PR #1803, opened 2014-12-14
enable create table in sqlite3

KenjiOhtsuka:
Modified classes/dbutil.php to enable create table in sqlite3.
Previous code can't create table with primary index in sqlite3,
because previous code generate

``` sql
PRIMARY KEY 'id' ('id')
```

but sqlite3 supports

``` sql
PRIMARY KEY ('id')
```

I used config/db.php setting string to check whether the database is sqlite3.

The following code using `Database_Connection` is another way to do it.

``` php
$sql .= ",\n\tPRIMARY KEY ";
if (strtolower(\Config::get('db.'.($db ? $db : \Config::get('db.active')).'.type')) === 'pdo')
{
  if (strtolower(\Database_Connection::instance($db)->driver_name()) !== 'sqlite')
  {
    $sql .= $key_name;
  }
}
$sql .= $key_name." (" . implode(', ', $primary_keys) . ")";
```

### DIFF
--- a/classes/dbutil.php
+++ b/classes/dbutil.php
@@ -123,7 +123,17 @@ class DBUtil
 		{
 			$key_name = \DB::quote_identifier(implode('_', $primary_keys), $db ? $db : static::$connection);
 			$primary_keys = \DB::quote_identifier($primary_keys, $db ? $db : static::$connection);
-			$sql .= ",\n\tPRIMARY KEY ".$key_name." (" . implode(', ', $primary_keys) . ")";
+			$sql .= ",\n\tPRIMARY KEY ";
+			if (strtolower(\Config::get('db.'.($db ? $db : \Config::get('db.active')).'.type')) === 'pdo')
+			{
+				$dsn = \Config::get('db.'.($db ? $db : \Config::get('db.active')).'.connection.dsn');
+				$_dsn_find_collon = strpos($dsn, ':');
+				if (($_dsn_find_collon ? substr($dsn, 0, $_dsn_find_collon) : null) !== 'sqlite')
+				{
+					$sql .= $key_name;
+				}
+			}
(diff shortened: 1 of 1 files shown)


kenjis:
If you use PDO and not SQLite, `$key_name` is duplicated. See #1943

WanWizard:
This code is not correct, the key name is not needed there at all, MySQL just ignores it when it's there, I haven't checked the other RDBMS syntax yet.

WanWizard:
Checked all PDO supported RDBMS engines:
- DB2, Cubrid, Informix, MSSQL, PostgreSQL and SQLite do not accept a key name at all
- MySQL does accept a keyname, but ignores it if it is there
- Firebird, Oracle: SQL syntax not supported by Fuel's DBUtil

So there should not be a key/constraint name at all, that is only relevant for secondary indexes.

KenjiOhtsuka:
I'm so sorry.
So should it be like below?

``` php
$sql .= ",\n\tPRIMARY KEY (" . implode(', ', $primary_keys) . ")";
```

WanWizard:
Instead of the code snippet above? Yes, I think it's as simple as that. 

No constraint or key name in all cases, only Oracle requires a constraint name, and Firebirds SQL syntax is completely diffferent, but we don't support those anyway.

If you can amend your PR, and test it, then I'll merge the update.

KenjiOhtsuka:
I changed the code and tested with MySQL, SQLite and PostgreSQL.

I'm plannig to test with DB2, Cubrid, Informix, MSSQL.

Which branch should I create PullRequest to? 1.8/develop?

KenjiOhtsuka:
@WanWizard 

I created new PR below.
https://github.com/fuel/core/pull/1947

PR #1814, opened 2014-12-30
Query Bug

Onuma0519:
Update Query Not Excute.
// $hogeArray = array('primary_key' = 1, 'property1' => 'hoge1', 'property2' => 'hoge2');
// $modelHoge = Model_Hoge::forge($hogeArray);
// $modelHoge->save();

### DIFF
--- a/classes/model/crud.php
+++ b/classes/model/crud.php
@@ -405,7 +405,7 @@ class Model_Crud extends \Model implements \Iterator, \ArrayAccess, \Serializabl
 	 */
 	public function __construct(array $data = array())
 	{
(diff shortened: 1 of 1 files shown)


emlynwest:
What exactly is the bug that you are fixing here? Ideally you should not be setting the PK of a model in such a way.

WanWizard:
The issue is valid, but the fix isn't.

The problem is that it checks for a primary key value before setting the values passed, instead of after. The fact that it checks `$this->_data` is actually valid, this is required for DB's `as_object()`.

PR #1822, opened 2015-01-31
add some tests of response class, prevent exit when redirect while testing

KenjiOhtsuka:
1. added some tests to `core/test/response.php`.
2. modified `redirect` method in `core/classes/response.php` to prevent `exit` while testing, especially in testing controllers method which contains redirection.

### DIFF
--- a/tests/response.php
+++ b/tests/response.php
@@ -20,5 +20,87 @@ namespace Fuel\Core;
  */
 class Test_Response extends TestCase
 {
- 	public function test_foo() {}
+	/**
+	 * Tests Response set_header
+	 * @test
+	 */
+	public function test_set_header() {
+		$response = Response::forge();
+
+		// insert name and value into header values
+		$name = 'header name';
+		$value = 'header value';
+		$response->set_header($name, $value);
+		$this->assertEquals($value, $response->headers[$name]);
+
+		// update previously added header value
+		$new_value = 'new header value';
+		$response->set_header($name, $new_value);
+		$this->assertEquals($new_value, $response->headers[$name]);
+
+		// insert name and value array into header values
+		$response->set_header($name, $value, false);
+		$this->assertContains(array($name, $value), $response->headers);
+	}
+
+	/**
+	 * Tests Reponse get_header
+	 * @test
+	 */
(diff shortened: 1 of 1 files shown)


emlynwest:
I don't agree with removing the exit in testing. It creates the issue that code in the controller will run when it should not, thereby making the tests inaccurate. Ideally unit tests should not cover controllers, but the core logic of your application. If you want to test that a redirect is working that I think that acceptance tests are more useful.

kenjis:
> Ideally unit tests should not cover controllers

Why?

emlynwest:
Because your controllers should contain as little logic as possible and since they are stitching parts of your app together its harder to test them as a single unit, or often pointless. IMO it's better to use acceptance or functional tests when it comes to controllers.

KenjiOhtsuka:
> It creates the issue that code in the controller will run when it should not, thereby making the tests inaccurate.

I understand, and erased the change of `classes/response.php`.

kenjis:
I agree:

> controllers should contain as little logic as possible

But don't agree:

> unit tests should not cover controllers

If we want to test controllers, we should test them.

By the way, how do you write functional tests? What tool do you use?

emlynwest:
I agree that controllers should be tested but I don't feel that unit tests are suitable as controllers potentially use lots of components so you'd have to end up mocking them all to make it a "true" unit test, because of this functional/acceptance tests can be used to ensure the controllers are producing the intended results.

I use codeception for testing, it's built on phpunit and other common testing frameworks, it allows you to define and run multiple suites of tests, so you can have unit, functional, acceptance and w/e else all built on the same platform.

kenjis:
In my understanding, functional tests of codeception do not use web server.

http://codeception.com/docs/05-FunctionalTests

> functional tests don't require a web server to run tests.

And codecetion functional tests do not support FuelPHP. Is it possible to write functional tests for controller which has `Response::redirect`?

emlynwest:
Currently there's no support for codeception->fuel integration but I plan to write an adaptor for v2. At the moment it's not possible to easily test `Response::redirect` because of the exit.

kenjis:
I got it. Thank you!

PR #1825, opened 2015-02-02
Fixing issue within dbutilities posting a INT as a string causing an error

lukepolo:


classes/dbutil.php
```diff
@@ -724,10 +724,14 @@ protected static function table_maintenance($operation, $table, $db = null)
 		{
 			if ( ! in_array($type, array('info', 'warning' , 'error')))
 			{
-				$type = \Fuel::L_INFO;
+                                $type = \Fuel::L_INFO;
```
emlynwest: This should be tabs

lukepolo: These are tabs , 
screen shot 2015-02-04 at 10 31 36 am

Am I missing something ?

emlynwest: Well your editor has saved them to the file as spaces, if you check the raw copy of the code on github you'll see spaces on this line instead of tabs.

lukepolo: My stupid editor! Ill see if I can change how it saves it

WanWizard:
oops, typo in the issue number...

WanWizard:
That was weird, and probably old code. logger() accepts either a integer level number (100-700), or an uppercase string, this has changed when we switched to Monolog in I think 1.5.

So instead of merging your workaround, I fixed the issue.

PR #1848, opened 2015-03-14
Discard Database_Connection instance on disconnect for mysqli

kanonji:
## What I did

``` php
$db = \Database_Connection::instance('my_db_setting');
$this->assertEquals(true, $db->start_transaction());
\DB::query('SELECT 1;')->execute($db);
$this->assertEquals(true, $db->disconnect());
$db = \Database_Connection::instance('my_db_setting');
\DB::query('SELECT 2;')->execute($db); // <- /path/to/fuel/app/tests/foo.php:13
$this->assertEquals(true, $db->start_transaction());
```

Note that I have used mysqli. (`'type' => 'mysqli'` on config/db.php)
## What I got

> Fuel\Core\PhpErrorException: mysqli::ping(): Couldn't fetch mysqli
> 
> /path/to/fuel/core/bootstrap.php:109
> /path/to/fuel/core/classes/database/mysqli/connection.php:205
> /path/to/fuel/core/classes/database/query.php:287
> /path/to/fuel/app/tests/foo.php:13

Because `$this->_connection` has mysqli object already [`mysqli::close()`](http://jp2.php.net/manual/en/mysqli.close.php) called after `disconnect()`.
mysqli object does not have connection status nor method for reconnect.
## Detail

At first I try to replace `$status = $this->_connection->close();` with `$this->_connection = null;` to be reconnected on `execute()`.

```
150313  6:08:56   166 Connect   foo@localhost on test_foo
          166 Query SET NAMES utf8
          166 Query START TRANSACTION
          166 Query SELECT 1
          166 Quit
          167 Connect   foo@localhost on test_foo
          167 Query SET NAMES utf8
          167 Query SELECT 2
          167 Query SAVEPOINT LEVEL1
          167 Quit
```

However Database_Connection object has state of transaction.
There is `SAVEPOINT LEVEL1` instead of `START TRANSACTION` in MySQL general log avobe.

I think it is better to recreate object than to control state of object carefully.

```
150313  6:05:43   160 Connect   foo@localhost on test_foo
          160 Query SET NAMES utf8
          160 Query START TRANSACTION
          160 Query SELECT 1
          160 Quit
          161 Connect   foo@localhost on test_foo
          161 Query SET NAMES utf8
          161 Query SELECT 2
          161 Query START TRANSACTION
          161 Quit
```

This is general log with code in this pull request.

### DIFF
--- a/classes/database/mysqli/connection.php
+++ b/classes/database/mysqli/connection.php
@@ -183,6 +183,11 @@ class Database_MySQLi_Connection extends \Database_Connection
 			// Database is probably not disconnected
 			$status = ! ($this->_connection instanceof \MySQLi);
 		}
+		if($status)
+		{
+			unset(self::$instances[$this->_instance]);
+			unset(static::$_readonly[$this->_instance]);
+		}
 
 		return $status;
 	}


WanWizard:
I don't entirely agree with simply destroying the object. The fact that the transaction seems to be hanging and isn't committed or rolled-back is an indication the connection isn't closed at all. The PHP docs seem to agree to this, and suggest do a kill() first, before calling close().

kanonji:
> The fact that the transaction seems to be hanging and isn't committed or rolled-back is an indication the connection isn't closed at all.

I could close connection without COMMIT or ROLLBACK that is described MySQL general log I put on main comment of this pull request.

```
150313  6:08:56   166 Connect   foo@localhost on test_foo
          166 Query SET NAMES utf8
          166 Query START TRANSACTION
          166 Query SELECT 1
          166 Quit  # <- connection closed by $this->assertEquals(true, $db->disconnect());
          167 Connect   foo@localhost on test_foo   # <- new connection
          167 Query SET NAMES utf8
          167 Query SELECT 2
          167 Query SAVEPOINT LEVEL1
          167 Quit
```

And the code I wrote destroys the object after `$status = $this->_connection->close();` that code is already in the HEAD of `classes/database/mysqli/connection.php`.
So I am sure that the connection wil be closed before the object destroyed.

Because of my English proficiency
, properly I may not understand what you mean. 
Are you worried about any other thing?

### About `mysqli::kill()`

I am not familiar with `mysqli::kill()`.
I googled and understand this kills MySQL thread.
I think killing MySQL thread is responsibility of MySQL daemon unless user wants stop very slow query.
Is it really needed before calling `mysqli::close()`?

WanWizard:
As far as I can see, the MySQLi driver doesn't really close the connection, it just sets a flag on the object that it can no longer be used. So the connection to the database remains open, creating a new object just re-uses that same connection, and that is why the previous transaction wasn't rolled back (or committed) on closing of the connection. Kill() will actually disconnect the connection with the database. So if you run your test with a kill() before the close(), and just null the connection variable after that, you should get the required result.

The old MySQL driver, and the PDO driver, don't have this problem.

kanonji:
To understand what you mean, I want make sure some points.

`Quit` is logged in MySQL general log when [`mysqli::close()`](http://jp2.php.net/manual/en/mysqli.close.php) is called.
- 1. Do you mean that MySQL and PHP(MySQLi driver) are still connected in situation of that avobe?
- 2. Only flag in MySQLi object changed not to be used and the connection is kept?

>  creating a new object just re-uses that same connection
- 3. Does new object mean MySQLi object that will be created when `new \MySQLi()` is called?

WanWizard:
If your server logs a `quit`, how is it possible that it keeps the current transaction hanging if you create a new connection?
1. Yes
2. That is wat some comments suggest. The actual tcp connection is not cleared, like with a persistent connection.
3. Yes

WanWizard:
Looks like your log results are caused elsewhere. When closing the connection, the savepoint pointer is not reset. So this must be an issue with the other drivers too.

kanonji:
Anyway thank you for fix, although I have not understood about how MySQLi driver treat connection yet.

WanWizard:
The close() issue seems to be limited to Windows.

The reason you got `SAVEPOINT LEVEL1` instead of `START TRANSACTION` on the second instance was because the transaction depth counter wasn't reset on close, causing the wrong SQL to be generated on starting a new transaction.

Appearently you're the first in 4 years that actually closes and then re-uses a DB instance...

kanonji:
> The close() issue seems to be limited to Windows.

I see. I did not write my environment, sorry.

> The reason you got SAVEPOINT LEVEL1 instead of START TRANSACTION on the second instance [snip]

I understand this problem. I just prefer to recreate object than to control transaction depth counter carefully.

> Appearently you're the first in 4 years that actually closes and then re-uses a DB instance...

I have sevral PHPUnit tests that use transaction.
Each test is valid when I run with `oil test --file=<path/to/test.php>`. But I got error with `oil test`.
So I want to make sure each test starts with not connected to MySQL.
Because all tests run in one runtime with `oil test`, I close and then re-use a DB instance as a result.

WanWizard:
That should now work fiine, even without re-creating the object.

PR #1851, opened 2015-03-18
fixed bug: indexed param gets lost on recursion

lsnider:
Recursive arrays where sub-arrays have indexed columns were getting ignored even if the $indexed parameter was set to true.  Also recursed on static::flatten as calling static::flatten_assoc was a straight pass through and unnecessary.


WanWizard:
@stevewest looks ok to you?

emlynwest:
I don't see a problem with it :+1:

PR #1852, opened 2015-03-24
Remove ORDER BY from sql count_last_query

bernhardh:
Since we don't need "ORDER BY" on count statements and its slows down some queries, we can remove it

### DIFF
--- a/classes/database/connection.php
+++ b/classes/database/connection.php
@@ -254,7 +254,7 @@ abstract class Database_Connection
 			if (stripos($sql, 'ORDER BY') !== false)
 			{
 				// Remove ORDER BY from the SQL to improve performacnce
-				$sql = preg_replace('/ORDER BY.*?(?=\\)|$)/mi', '', $sql);
+				$sql = preg_replace('/ORDER BY [^,\s)]*(?:\s*,\s*[^,\s)]+)*/mi', '', $sql);
 			}
 
 			// Get the total rows from the last query executed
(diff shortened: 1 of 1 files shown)


WanWizard:
This regex seems to strip everything starting with the `ORDER BY`

So `OFFSET 0 ORDER BY table.field LIMIT 10` will become  `OFFSET 0`, and not `OFFSET 0 LIMIT 10`

bernhardh:
I know, but thats not a problem, since we don't need OFFSET or LIMIT as well (and we remove both of theme some lines above:

```
        if (stripos($sql, 'LIMIT') !== false)
        {
            // Remove LIMIT from the SQL
            $sql = preg_replace('/\sLIMIT\s+[^a-z\)]+/i', ' ', $sql);
        }
        if (stripos($sql, 'OFFSET') !== false)
        {
            // Remove OFFSET from the SQL
            $sql = preg_replace('/\sOFFSET\s+\d+/i', '', $sql);
        }
```

WanWizard:
Maybe I used the wrong example. The regex strips anything, it could also be an `GROUP BY` for example. The other regexes only strip the keyword and it's value.

bernhardh:
You are right.
Please have a look at the following regex (and usecases): https://regex101.com/r/jS5fN1/5
(Thanks to User vks from StackOverflow). 

Have I overseen another possible situation?

WanWizard:
That looks a lot better!

WanWizard:
This regex doesn't deal with ASC or DESC which can be part of the ORDER BY clause.

PR #1854, opened 2015-03-25
Use variadic when php56 or higher

it-can:


emlynwest:
Seeing that the minimum version for v1 is 5.3, what is the value of this?

it-can:
Yeah, well I run 5.6 for my fuel apps... Just a nice to have ;-)

emlynwest:
But what is the actual value of this? Is it faster? Did you benchmark it?

it-can:
I did a bit of testing, and the current implementation falls back to `call_user_func_array` (when having more than 4 variables), this seems to happen a lot with the DB class because sometimes this can have up to 44 variables (in my case)... I did not benchmark it, but this implementation feels a lot of faster.

emlynwest:
I can't accept a merge because "it feels faster" :wink: Another issue with this is also that it does a version check every time the function is called.

it-can:
https://wiki.php.net/rfc/argument_unpacking

"The ... argument unpacking syntax is about 3.5 to 4 times faster than call_user_func_args. This solves the performance issue."

emlynwest:
But you have still not benchmarked it, how much faster does this make fuel's code, and how does it impact performance when running fuel on lower php versions?

WanWizard:
Looked at this, and it does make an measurable impact on 5.6, altough imo not very significant. The reason the `call_fuel_func_array()` method exists in the first place is because the original is so slow, especially on 5.3. I don't think it hurts to implement this, given the fact that it will eradicate the call completely on 5.6+.

PR #1865, opened 2015-05-02
Added support to Controller_Rest for additional headers in response

evilwalrus:
Signed-off-by: Andrew Heebner andrew.heebner@gmail.com

Simply put, I needed a quick option to add additional headers to the output of response().  Added a quick additional parameter to response(), as well as a quick loop to set the headers.

This was required by myself for rate-limiting, and access control headers.  Figured it could be handy for other users as well, rather than working around the response to set the headers.

Also updated `$status` to default to a 200 OK.  You technically can't return a null HTTP status code, so this ensures that something (valid) is at least returned.

### DIFF
--- a/classes/controller/rest.php
+++ b/classes/controller/rest.php
@@ -170,15 +170,23 @@ abstract class Controller_Rest extends \Controller
 	 *
 	 * @param   mixed
 	 * @param   int
+     * @param   array
 	 * @return  object  Response instance
 	 */
-	protected function response($data = array(), $http_status = null)
+	protected function response($data = array(), $http_status = 200, $headers = array())
 	{
 		// set the correct response header
 		if (method_exists('Format', 'to_'.$this->format))
 		{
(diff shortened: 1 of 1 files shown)


emlynwest:
This looks good, but why not merge the header array rather than looping over them? I would have thought that would be slightly more efficient.

WanWizard:
There are some coding standard issues with this PR. But the biggest problem I see is the lack of added value.

```
$headers = array(
    'some-header' => 'some-value',
    'other-header' => 'other-value',
);
$response = $this->response($data, 200, $headers);
```

vs

```
$headers = array(
    'some-header' => 'some-value',
    'other-header' => 'other-value',
);
$response = $this->response($data, 200);

foreach($headers as $key => $value)
{
    $response->set_header($key, $value);
}
```

For the occasion you actually need it? I can imagine if you always have additional headers to pass, you find a better solution that just passing it on every call to response() ?

PR #1870, opened 2015-06-18
Fix issue if path is symlink in File_Area:get_path

MacTwister:
My app/tmp folder is symlinked to a folder outside the root, as I use a deployment system. When ever I tried to do File::download it would tell me file not found. 

I am using an Area restricted to `APPPATH.'tmp'`and downloading a file using `File::download`, but the `File_Are::get_path` would return a long path which included path of the area concatenated with path of the file Im downloading

ie. `/var/www/shared/fuel/app/tmp/var/www/release/xxxxxxx/fuel/app/tmp/download.zip`

So I quickly thought this would resolve the issue.

### DIFF
--- a/classes/file/area.php
+++ b/classes/file/area.php
@@ -143,6 +143,13 @@ class File_Area
 		// make sure we have a dirname to work with
 		isset($pathinfo['dirname']) or $pathinfo['dirname'] = '';
 
+		//Make sure dirname path is realpath (symlink?)
+		$dirname_realpath = realpath($pathinfo['dirname']);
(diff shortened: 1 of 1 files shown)


WanWizard:
I can't reproduce this.

I symlinked my "tmp" folder to elsewhere, placed a file called "test.txt" there, 644 rights, and used this controller code:

``` php
$area = \File::forge([
    'basedir' => APPPATH.'tmp'.DS,
]);

echo '<pre>',$area->read('test.txt'),'</pre>';
```

And I can read the file without problems. If you still think this is an issue, please re-open this with a code example to reproduce the issue.

PR #1884, opened 2015-07-09
do not work php 5.3

yorozuone:
display

Fatal error: Using $this when not in object context in /var/www/html/f/fuel/core/classes/view.php on line 282

php 5.3 do not use $this closures inside ?

### DIFF
--- a/classes/view.php
+++ b/classes/view.php
@@ -270,7 +270,8 @@ class View
 	 */
 	protected function get_data($scope = 'all')
 	{
-		$clean_it = function ($data, $rules, $auto_filter)
+		$filter_closures = $this->filter_closures;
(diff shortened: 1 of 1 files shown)


kenjis:
Yes, it is new feature of PHP 5.4:

> Closures now support $this. 
> http://php.net/manual/en/migration54.new-features.php

emlynwest:
Thanks again. Ideally you should not be using 5.3 as it is EOL now though ;)

yorozuone:
You're right. I want to use PHP5.5.
But I have several reason. Because I must use php5.3 :,-(
Near future I challenge update PHP5.5.
Thank you.

PR #1893, opened 2015-07-15
Fix valid_date function

kz23s:
When validate date format values such as 'Y' or 'Y-m' using 'valid_date' the validated value changes from before validation.

**Example**
input value: 2015-07
validated value: 2015-06

``` php
$val = Validation::forge();
$val->add('date_format_value', '')->add_rule('valid_date', 'Y-m');   // 2015-07 
$val->run();
$validated_input = $val->validated();
echo $validated_input['date_format_value'];  // 2015-06
```

Signed-off-by: Kazufumi Suenaga kzfms07@gmail.com

### DIFF
--- a/classes/validation.php
+++ b/classes/validation.php
@@ -1045,7 +1045,7 @@ class Validation
 		{
 			if ($format)
 			{
-				return date($format, mktime($parsed['hour'], $parsed['minute'], $parsed['second'], $parsed['month'], $parsed['day'], $parsed['year']));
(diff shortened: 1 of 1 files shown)


it-can:
+1

WanWizard:
Good find, but not complete. All the other fields in the array can be `false` as well, so those require a fallback value too. 

I think for hour, minute and section the default can be zero, not sure what default value to take for the year. 1970?

PR #1895, opened 2015-07-16
Fix phpdoc for dbutil.

ghost:

### DIFF
--- a/classes/dbutil.php
+++ b/classes/dbutil.php
@@ -29,6 +29,7 @@ class DBUtil
 	/**
 	 * Sets the database connection to use for following DBUtil calls.
 	 *
+	 * @throws \FuelException
 	 * @param  string  $connection  connection name, null for default
 	 */
 	public static function set_connection($connection)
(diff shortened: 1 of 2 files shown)


classes/dbutil.php
```diff
@@ -167,8 +168,9 @@ public static function add_fields($table, $fields, $db = null)
 	 * Modifies fields in a table.  Will throw a Database_Exception if it cannot.
 	 *
 	 * @throws  Fuel\Database_Exception
-	 * @param   string  $table    the table name
-	 * @param   array   $fields   the modified fields
+	 * @param   string  $table   the table name
+	 * @param   array   $fields  the modified fields
+     * @param   string  $db      the database connection to use
```
emlynwest: Indentation looks wrong here, seems to be spaces instead of tabs.

ghost: @stevewest 
Sorry, fixed.

emlynwest: Thanks!

PR #1899, opened 2015-07-17
Fix phpdoc for finder.

ghost:


emlynwest:
Thanks again for all the doc updates!

PR #1900, opened 2015-07-17
Fix phpdoc, invalid return value and remove dead code of event.

ghost:

### DIFF
--- a/classes/event/instance.php
+++ b/classes/event/instance.php
@@ -203,7 +203,7 @@ class Event_Instance
 				return json_encode($calls);
 				break;
 			case 'none':
-				return false;
+				return null;
 			case 'serialized':
 				return serialize($calls);
 				break;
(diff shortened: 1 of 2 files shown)


classes/event/instance.php
```diff
@@ -209,7 +203,7 @@ protected function _format_return(array $calls, $return_type)
 				return json_encode($calls);
 				break;
 			case 'none':
-				return;
+				return false;
```
emlynwest: Not sure about this. @WanWizard any idea if this is going to cause BC issues?

WanWizard: I don't see or understand why this has to change? It returns `null` here by design, why should it return false now? If would break BC if people use the returned value, so if there's no compelling reason for changing this, I'd say not change it.

ghost: @stevewest @WanWizard 
Sorry, I was misunderstanding design and specification of PHP. I fixed it, please peer-review again.

PR #1908, opened 2015-07-22
Fix phpdoc, invalid argument for html.

ghost:

I found some formal parameter(s) uses false as default value(e.g. L266). It was not only in this class.
I think that those it should use empty array(array case) or null(other), if this project needs more strict code and sophisticated phpdoc. Because parameter type declared as non-boolean(e.g. string, array...) in phpdoc.
If everyone will agree this opinion, I will fix that matter.

Sorry for my poor english.

### DIFF
--- a/classes/html.php
+++ b/classes/html.php
@@ -125,7 +125,7 @@ class Html
 	 * @param	string	$text		the text value
 	 * @param	string	$subject	the subject
 	 * @param	array	$attr		attributes for the tag
-	 * @return	string	the javascript code contain email
+	 * @return	string	the javascript code containing email
 	 */
 	public static function mail_to_safe($email, $text = null, $subject = null, $attr = array())
 	{
--- a/classes/html.php
+++ b/classes/html.php
@@ -276,11 +276,11 @@ class Html
 		{
 			if ( ! is_array($val))
 			{
-				$out .= $indent."\t".html_tag('li', null, $val).PHP_EOL;
+				$out .= $indent."\t".html_tag('li', array(), $val).PHP_EOL;
 			}
 			else
 			{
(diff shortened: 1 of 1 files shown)


classes/html.php
```diff
@@ -120,11 +121,11 @@ public static function mail_to($email, $text = null, $subject = null, $attr = ar
...
-	 * @param	string	the email address
-	 * @param	string	the text value
-	 * @param	string	the subject
-	 * @param	array	attributes for the tag
-	 * @return	string	the javascript code containg email
+	 * @param	string	$email		the email address
+	 * @param	string	$text		the text value
+	 * @param	string	$subject	the subject
+	 * @param	array	$attr		attributes for the tag
+	 * @return	string	the javascript code contain email
```
emlynwest: `code containing the email` would be better English here.

emlynwest:
Im not all too sure of your reasoning behind changing the `false` to `null`, at the moment the internal code style does not affect how the code is used. The inconsistent/bad style is something that we are addressing with the v2 rewrite.

ghost:
OK. Thank you for the explanation.

WanWizard:
In this case `html_tag()` uses `empty()`, so you can get away with a lot.

But given the fact that it's the tag's attribute list that is passed there, and it's docblock says it accepts "array|string", I would suggest to neither use `null` nor `false`,but use `array()`.

ghost:
@WanWizard 
You mean [this](https://github.com/TOYOZUMIKouichi/core/commit/39f189e726527bce3c8a8b04b9c76473b2c9c55e) ?

WanWizard:
Yup, that looks fine.

PR #1916, opened 2015-08-05
Fix phpdoc, typo for model/crud.

ghost:

Why many static variables is commented out? That used in some methods.


WanWizard:
@TOYOZUMIKouichi 

They are commented out because the code in the base class checks for their existence using property_exists(). They are there because you can use them in your Crud classes that extend the base class, to control behaviour. So they should remain there, and commented out. They are there as part of the in-code documentation.

PR #1925, opened 2015-08-26
Fix phpdoc, typo for session.

ghost:
Signed-off-by: TOYOZUMI Kouichi kouichi.toyozumi@gree.net

### DIFF
--- a/classes/session.php
+++ b/classes/session.php
@@ -326,7 +326,7 @@ class Session
 	 */
 	public static function create()
 	{
-		static::instance()->create();
+		return static::instance()->create();
 	}
 
 	// --------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


classes/session.php
```diff
@@ -332,12 +322,11 @@ public static function delete_flash($name = null)
 	/**
 	 * create a new session
 	 *
-	 * @access	public
 	 * @return	void
 	 */
 	public static function create()
 	{
-		return static::instance()->create();
+		static::instance()->create();
```
emlynwest: This could possibly cause a BC issue if someone is relying on the return value from `create()` from a custom session driver.

ghost:
@stevewest OK, I reverted that change.

emlynwest:
Thanks

Issue #1942, opened 2015-10-21
Fatal error: Uncaught Error: Call to undefined method Error::shutdown_handler()


kenjis:
Is this a bug of PHP or Fuel?

> Fatal error: Uncaught Error: Call to undefined method Error::shutdown_handler() in .../fuel-1.8-dev/fuel/core/bootstrap.php:71 Stack trace: #0 [internal function]: {closure}() #1 {main} thrown in .../fuel-1.8-dev/fuel/core/bootstrap.php on line 71

Fix:

``` diff
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -68,7 +68,7 @@ register_shutdown_function(function ()
            logger(\Fuel::L_ERROR, 'shutdown - ' . $e->getMessage()." in ".$e->getFile()." on ".$e->getLine());
        }
    }
-   return \Error::shutdown_handler();
+ return Fuel\Core\Error::shutdown_handler();
 });

 set_exception_handler(function (\Exception $e)
```

```
$ php -v
PHP 7.0.0RC5 (cli) (built: Oct 14 2015 19:49:05) ( NTS )
Copyright (c) 1997-2015 The PHP Group
Zend Engine v3.0.0-dev, Copyright (c) 1998-2015 Zend Technologies
    with Zend OPcache v7.0.6-dev, Copyright (c) 1999-2015, by Zend Technologies
```

WanWizard:
Fuel v1 hasn't been tested on PHP7 yet, so I couldn't say. 
The Error class should have been aliased there.

kenjis:
Can't use `Error` class on PHP7. See http://php.net/manual/en/class.error.php

WanWizard:
I know (just didn't think of it). 

It's something that needs to be addressed, because your fix means you can't overload the Error class anymore.

emlynwest:
Looks like we need to rename the `Error` class for PHP7

kenjis:
Yes, we need to rename the `Error` class.

WanWizard:
There might be other PHP7 fixes needed. Changing this will break BC for v1 (in regards to extending the Error class), so I need a careful think about how to address this issue.

We might have to use the same sort of trick we used for Redis when that became a PHP extension.

emlynwest:
This could simply be something we need to break to continue with php7, how many people extend the `Error` class anyway?

WanWizard:
Already addressed, fix is on it's way ;) I just hate breaking BC if it's not needed.

[...]

PR #1944, opened 2015-10-23
Fix format.php (data format is invalid)

shucream0117:
When I exec `php oil test`, following error occured.
I thought that provided xml data format of `<field4>` is invalid.

```
1) Fuel\Core\Test_Format::test_to_xml_boolean with data set #0 (array(array('Value 1', 35, true, false)), '<?xml version="1.0" encoding=.../xml>\n', '<?xml version="1.0" encoding=.../xml>\n', '<?xml version="1.0" encoding=.../xml>\n')
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
 '<?xml version="1.0" encoding="utf-8"?>
-<xml><item><field1>Value 1</field1><field2>35</field2><field3>1</field3><field4/></item></xml>
+<xml><item><field1>Value 1</field1><field2>35</field2><field3>1</field3><field4></field4></item></xml>
 '

/home/vagrant/chachat/fuel/core/tests/format.php:414

FAILURES!
Tests: 389, Assertions: 481, Failures: 1.
```

### DIFF
--- a/tests/format.php
+++ b/tests/format.php
@@ -122,7 +122,7 @@ line 2","Value 3"',
 					array('field1' => 'Value 1', 'field2' => 35, 'field3' => true, 'field4' => false),
 				),
 				'<?xml version="1.0" encoding="utf-8"?>
-<xml><item><field1>Value 1</field1><field2>35</field2><field3>1</field3><field4/></item></xml>
+<xml><item><field1>Value 1</field1><field2>35</field2><field3>1</field3><field4></field4></item></xml>
 ',
 				'<?xml version="1.0" encoding="utf-8"?>
 <xml><item><field1>Value 1</field1><field2>35</field2><field3>true</field3><field4>false</field4></item></xml>


WanWizard:
Which PHP version are you using?

See also #1942

shucream0117:
I'm using PHP5.6.

WanWizard:
Now that is interesting.

I run PHP 5.6.14 here, and I don't have this problem. So it's not PHP version related as we initially thought, but it has an external cause, perhaps a newer version of libxml?

What OS are you running PHP on?

kenjis:
And what version of libxml do you use?

```
$ php -i | grep libxml
```

WanWizard:
On my machine:

```
libxml Version => 2.9.2
libxml2 Version => 2.9.2
libxslt compiled against libxml Version => 2.9.1
```

There is a libxml setting that controls this behavoir, LIBXML_NOEMPTYTAG, but the Format class uses simplexml, and that doesn't support this setting. So it looks like a default has changed somewhere?

shucream0117:
I'm using CentOS6.6(64bit).
And, version of libxm is 2.7.6

```
[vagrant@localhost ~]$ php -i | grep libxml
libxml Version => 2.7.6
libxml
libxml2 Version => 2.7.6
libxslt compiled against libxml Version => 2.7.6
```

WanWizard:
Found a reference: https://github.com/facebook/hhvm/issues/2983

According to this issue it seems to have changed in 2.7.8. So perhaps we need to check the libxml version in the test instead of the PHP version...

WanWizard:
There is a LIBXML_VERSION constant, that could be checked against 20708 as version number.

kenjis:
The version of libxml I ran php7 is 2.7.6. It is on CentOS 6.
I got the same failure with PHP 5.5.

shucream0117:
Ok, I created https://github.com/fuel/core/issues/1945 and close this PR!

PR #1953, opened 2015-11-17
Update connection.php

fufu70:
Changed the Regex to work properly with removing the Order By clause. It did not remove the ASC or DESC string because of a missing space search in the regex


WanWizard:
Brillant find, I thought I'd had all permutations...

PR #1959, opened 2015-12-07
Properly escape HTTP Accept header before inserting into regex

thewebface:
Properly escape the data from the HTTP Accept header before inserting into preg_match. This will prevent 'Compilation failed' errors from being thrown if the header contains, for example, an unmatched parenthesis.


thewebface:
Ah I had misread an italic pipe as a forward slash in the preg_quote docs.

It looks like there's still a bug though, I've added another PR #1960

PR #1966, opened 2015-12-23
Fuelphp \Image::crop_dim($x,$y,$w,$h)

JefferyHus:
\Image::crop_dim($x,$y,$w,$h): this function will let you crop and re-size the image after loading it.

### DIFF
--- a/classes/image/driver.php
+++ b/classes/image/driver.php
@@ -75,7 +75,7 @@ abstract class Image_Driver
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


WanWizard:
Can you add the corresponding docs to as well (https://github.com/fuel/docs/blob/1.8/develop/classes/image.html in fuel/docs)?

JefferyHus:
yes sure i will do it

PR #1968, opened 2016-01-06
Fix module 403,500 routing

nnssn:
module 403,500 routing isn't functioning.

### DIFF
--- a/classes/request.php
+++ b/classes/request.php
@@ -280,14 +280,22 @@ class Request
 				// load and add the module routes
 				$module_routes = \Fuel::load($module_path);
 
+				$reserve_routes = array(
+					'_root_' => $module,
+					'_403_'  => '_403_',
+					'_404_'  => '_404_',
+					'_500_'  => '_500_',
+					$module  => $module,
+				);
+
 				$prepped_routes = array();
 				foreach($module_routes as $name => $_route)
 				{
-					if ($name === '_root_')
+					if (isset($reserve_routes[$name]))
 					{
-						$name = $module;
+						$name = $reserve_routes[$name];
 					}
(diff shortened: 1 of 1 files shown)


WanWizard:
This is not a bug.

You can't "route" to these pages, and all of them are global. Like your application can only have one "_root_", it can also only have one 403, 404, 500, etc page.

_root_ is only used when you get a request for the root of your app ( i.e. http://example.org/ ), the others are exceptions, i.e. used when you throw HttpNotFoundException (404), HttpNoAccessException(403), etc.

kenjis:
@WanWizard `_root_` and `_404_` in module's config work.

If we set `routes.php` in `api` module:

``` php
return array(
    '_root_'  => 'api/index/index',
    '_404_'   => 'api/error/404',
);
```

If we access `http://example.org/api`, go to `api/index/index`.
If we access `http://example.org/api/404_not_found_page`, go to `api/error/404`.
So I think many users are using this functionality.

And if we can have only one global `_404_`, what do we do when I want to return JSON 404 page in a module? Same as 403 and 500.

WanWizard:
Learning all the time.

So if it works, why is this commit needed? Or if you say it works, and this commit is needed, what broke it?

I would like to understand if before I merge this commit...

If you add these routes to the module routes file, they will get merged with the global routes with the name of the module prefixed, so you get a route entry like `api/_404_`. Nothing in the framework will route to that?

WanWizard:
Ah, wait. The old code only supported the `_404_` route, but not the later added `_403_` and `_500_` routes. I completely missed that.

kenjis:
Yes. Sorry for my little explanation.

WanWizard:
No problem, it opened my eyes... :-)

PR #1969, opened 2016-01-11
Fix config to not cache closure results

andrewwheal:
A closure can only be used as a config value in a php config file, so this could be moved to classes/config/php.php, but that would make it basically a duplication of code.

This changes the check for whether to skip caching to be done on the raw value from config, rather than the value parsed by Fuel::value(), so as to also catch closures being returned.

This fix means that with the use of closures for config options we are running the closure every time the config option is retrieved which could slow down code, however it means that a closure can be used to generate different results each time it is called. For example a closure as a config option to generate tokens which you don't want to be the same every time.

### DIFF
--- a/classes/config.php
+++ b/classes/config.php
@@ -192,14 +192,15 @@ class Config
 		}
 		elseif ( ! isset(static::$itemcache[$item]))
 		{
-			$val = \Fuel::value(\Arr::get(static::$items, $item, static::$default_check_value));
+			$raw_val = \Arr::get(static::$items, $item, static::$default_check_value);
+			$val = \Fuel::value($raw_val);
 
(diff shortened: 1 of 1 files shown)


WanWizard:
Looking at it, can't we just reduce it to

``` php
    public static function get($item, $default = null)
    {
        if (isset(static::$items[$item]))
        {
            return static::$items[$item];
        }
        elseif ( ! isset(static::$itemcache[$item]))
        {
            if ( ! \Arr::key_exists(static::$items, $item))
            {
                return $default;
            }

            static::$itemcache[$item] = \Arr::get(static::$items, $item);;
        }

        return static::$itemcache[$item];
    }
```

which stores the closure in the cache instead of the value returned by the closure?

PR #1974, opened 2016-02-10
Add security.csrf_auto_token option to add CSRF token automatically

kenjis:
This PR adds `security.csrf_auto_token`.
If you set it true, CSRF token is added automatically when you call `Form::open()`.

### DIFF
--- a/classes/form/instance.php
+++ b/classes/form/instance.php
@@ -140,6 +140,12 @@ class Form_Instance
 			$form .= PHP_EOL.$this->hidden($field, $value);
 		}
 
+		// Add CSRF token automatically
+		if (Config::get('security.csrf_auto_token', false))
+		{
+			$form .= PHP_EOL.\Form::csrf();
+		}
+
 		return $form;
 	}
 
--- a/tests/form.php
+++ b/tests/form.php
@@ -275,4 +275,66 @@ class Test_Form extends TestCase
 		
 		\Config::set('form.auto_id', $config);
 	}
+
+	/**
+	* Tests Form::open()
+	*
+	* @test
+	*/
+	public function test_open()
+	{
+		$form = \Form::forge(__METHOD__);
+		
+		$output = $form->open('uri/to/form');
+		$expected = '<form action="uri/to/form" accept-charset="utf-8" method="post">';
+		$this->assertEquals($expected, $output);
+	}
+
+	/**
+	* Tests Form::open()
+	*
+	* @test
+	*/
+	public function test_open_auto_csrf_token()
+	{
+		$config = \Config::get('security.csrf_auto_token');
+		\Config::set('security.csrf_auto_token', true);
+
+		$form = \Form::forge(__METHOD__);
+		
+		$output = $form->open('uri/to/form');
+		$expected = '<form action="uri/to/form" accept-charset="utf-8" method="post">'.PHP_EOL.'<input name="fuel_csrf_token" value="%s" type="hidden" id="form_fuel_csrf_token" />';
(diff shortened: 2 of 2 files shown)


kenjis:
Thank you for merging.
We need to update docs and `config.php`. I sent PR for docs: https://github.com/fuel/docs/pull/712

By the way, why the default value for `security.csrf_autoload` is `true` in the below?
https://github.com/fuel/core/blob/1.8/develop/classes/security.php#L56

I think the default value is `false`: https://github.com/fuel/core/blob/1.8/develop/config/config.php#L153
But the docs says `true`: https://github.com/fuel/docs/blob/1.8/develop/classes/security.html#L159-L161

All of them should be `false`, right?

WanWizard:
From a security point of view, you would say the default should be `true`, because that forces you to have csrf protection on every form. Problem with that is that currently every POST is checked, so all REST calls will fail too, unless they contain a token, which is not likely.

I did a quick history check, it seems it has always been `false` in the config, so for BC reasons I'd say all of them should be `false`. And a solution for REST calls must be added to the security class.

kenjis:
> From a security point of view, you would say the default should be true

Yeah, I agree with you!
But for now, all of them should be `false` for BC.

I sent PRs to fix:
- https://github.com/fuel/core/pull/1976
- https://github.com/fuel/docs/pull/713

kenjis:
> Problem with that is that currently every POST is checked, so all REST calls will fail too, unless they contain a token, which is not likely.

Do we add ignore URI list like this?
https://gist.github.com/vzool/87b13a3d0c8e168a3587

WanWizard:
Would probably do it, but not very elegant.

Thinking about it, I think when you enable automatic checking, you intend to check on every POST, and you as a developer should make sure every POST contains a token. If you don't intend to do that, don't enable auto checking, but check manually, for example in the `before()` of a controller.

kenjis:
If we have a site which has web apis, we may not use tokens on the apis.
But we may want to enable automatic checking on every POST to normal web pages.

I don't think manual checking is safer. Because if a developer forgets to implements checking, the site would be vulnerable. If we have a lot of forms or important actions, it costs much to confirm every important action has checking logic.

In my opinion, we should enable automatic checking always.

WanWizard:
I think there two different conditions:
1. Do we automatically check the CSRF token on every POST?
2 If yes, do we also check posts to a REST API?

Currently, we don't do neither by default because of BC reasons. In the app, we can do 1. by enabling `csrf_autoload` in the config, but there is no option to omit POSTS to a REST API if it is enabled.

Perhaps we should introduce two new config keys, something like `csrf_validate_form_posts` and `csrf_validate_rest_posts` And if they don't exist, fallback to the `csrf_autoload` value for both, and if that not exist fallback to `false`.

This way you can define the behavior for both cases independently, and have BC covered.

kenjis:
How do we know whether REST posts or not?

WanWizard:
is_ajax() ?

kenjis:
1. Every Web API does not expect Ajax calls. For example, API like GitHub API.
2. I'm not sure attackers can't make a victim send Ajax requests to cross domains. XHR2 could send cross domain requests.

WanWizard:
I know that, but I don't think we can come up with something that fits all possible scenario's. If you have a specific environment, you need specific checks.

So I'm looking for something that follows the 80/20 rule, and covers the majority of the use-cases.

kenjis:
I opened a issue for this new functionality: https://github.com/fuel/core/issues/1981
Because I think we need to think how to implement it more.

PR #1975, opened 2016-02-11
Fix Test_Form

kenjis:
See https://github.com/fuel/core/commit/f5601153bf623bec4a9252deafbf818ebb8a394e#commitcomment-16009315


it-can:
👍

PR #1984, opened 2016-02-25
Fix call to inaccessible protected method within PhpErrorException class.

benplummer:
The `PhpErrorException` class currently calls a protected static method on the `ErrorHandler` class which is not directly accessible. If we made the protected function public, PHP errors generated here would be displayed even in Production environments. We do not want this, thus we are calling the public `exception_handler` method instead.

### DIFF
--- a/classes/errorhandler.php
+++ b/classes/errorhandler.php
@@ -34,7 +34,7 @@ class PhpErrorException extends \ErrorException
 			if (\Fuel::$env != \Fuel::PRODUCTION and ($this->code & error_reporting()) == $this->code)
 			{
 				static::$count++;
-				\Errorhandler::show_php_error(new \ErrorException($this->message, $this->code, 0, $this->file, $this->line));
+				\Errorhandler::exception_handler(new \ErrorException($this->message, $this->code, 0, $this->file, $this->line));
 			}
 		}
 		elseif (\Fuel::$env != \Fuel::PRODUCTION


codeChap:
Why do you want to access the protected method? Also you should not display errors in production environments.

benplummer:
I think the initial description may have been misunderstood. I've updated it to make it clearer.

WanWizard:
I don't see why the change to exception_handler is needed, you do need to display PHP errors there, the if() will prevent it being called when in Production.

xploSEoF:
By the looks of things, there are 2 classes in the one file, one of which is trying to call a protected function in the other.

This PR is a bugfix, resolving the issue by calling the appropriate function.

As a side note, if the function was not protected, it would cause an issue by exposing errors in the production environment.

:+1:

andrewwheal:
It seems it was broken by bcc64a728c87519e31a769a3fbfbcca307932e35 where [Errorhandler::show_php_error() was changed to protected](https://github.com/fuel/core/commit/bcc64a728c87519e31a769a3fbfbcca307932e35#diff-9ec4b42d91f02a98896f3cc73c6008ccR210), which means [PhpErrorException::recover() can't call it](https://github.com/fuel/core/blob/bcc64a728c87519e31a769a3fbfbcca307932e35/classes/errorhandler.php#L37)

WanWizard:
I'm not debating that (it should indeed be public), I wonder why show_php_error() was replaced by calling the exceptionhandler instead.

xploSEoF:
No, it shouldn't be public - using that function elsewhere exposes errors on the production environment.

Instead it should be changed to calling the exception handler function, as this PR does.

andrewwheal:
I think @benplummer was just not wanting to make a protected method public. If it should be public then that's a better change.

WanWizard:
Ok,  clear.

@xploSEoF Only if you call it yourself, so who's problem is that?

xploSEoF:
@WanWizard All errors should be routed through the same exception handler.

Allowing any by-passing of this function could result in uses of php_show_error by-passing the logic that prevents errors being shown on the production environment.

This PR handles it correctly - these functions should remain protected.

WanWizard:
Already merged, was only a remark.

PR #1993, opened 2016-04-23
Add option to validate a specific IP format (IPv4 or IPv6)

kenjis:
This PR adds option to validate only IPv4 or IPv6 address.

### DIFF
--- a/classes/validation.php
+++ b/classes/validation.php
@@ -884,11 +884,22 @@ class Validation
 	 * Validate IP using PHP's filter_var()
 	 *
 	 * @param   string  $val
+	 * @param   string  ipv4|ipv6
 	 * @return  bool
 	 */
-	public function _validation_valid_ip($val)
+	public function _validation_valid_ip($val, $flag = null)
 	{
-		return $this->_empty($val) || filter_var($val, FILTER_VALIDATE_IP);
+		switch (strtolower($flag))
+		{
+			case 'ipv4':
+				$flag = FILTER_FLAG_IPV4;
+				break;
+			case 'ipv6':
+				$flag = FILTER_FLAG_IPV6;
+				break;
+		}
+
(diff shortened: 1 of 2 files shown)


classes/validation.php
```diff
@@ -884,11 +884,22 @@ public function _validation_valid_url($val)
...
 	{
-		return $this->_empty($val) || filter_var($val, FILTER_VALIDATE_IP);
+		switch (strtolower($flag))
+		{
+			case 'ipv4':
+				$flag = FILTER_FLAG_IPV4;
+				break;
+			case 'ipv6':
+				$flag = FILTER_FLAG_IPV6;
+				break;
```
WanWizard, line 899: perhaps add a

```
default:
    $flag = null;
```

to avoid all sorts of funny errors if some other flag is passed? Or an exception with "invalid flag" or something"?

WanWizard:
Can you update the docs too?

classes/validation.php
kenjis, line 899: The current code allows you to put filters other than IPv4 and IPv6.
http://php.net/manual/en/filter.filters.flags.php
Eg:

``` php
$flag = FILTER_FLAG_NO_PRIV_RANGE | FILTER_FLAG_NO_RES_RANGE;
$val->add_field('ip', 'IP address (not private/reserved)', 'valid_ip['.$flag.']');
$val->run(array('ip' => '192.168.0.1')); // false
```

But if we add the default, we can use only two filters.

I think this is flexible, but it does not check invalid filters. I have no strong opinion which is better.
So if you think we should add the default, I will do.
It seems there is no errors if we provide invalid flag value.

kenjis:
Done: https://github.com/fuel/docs/pull/727

classes/validation.php
WanWizard, line 899: ok.

PR #1998, opened 2016-05-24
add prefix to savepoint name

KoteiIto:
mysql syntax error occurred, because savepoint identifier can not use number.


WanWizard:
We only accept PR's on the develop branch, not on the production branch.

Problem is already addressed in develop: https://github.com/fuel/core/commit/46faed4210ef58c1fa23434b2cf1961ce66e2929

We're planning on releasing hot-fix 1.8.0.1 in a few days.

PR #1999, opened 2016-05-26
The Read Preferences option of mongoDB driver to be set from the Config

sanzo83:
The Read Preferences option of mongoDB driver to be set from the Config.

http://php.net/manual/en/mongo.readpreferences.php

For example,
/fuel/app/config/db.php

> 'mongo' => array(
>       'default' => array(
>                   'hostname'   => 'db1,db2,db3',
>                   'database'   => 'test',
>                   'replicaset' => 'myrep',
>                   'readPreference' => \MongoClient::RP_PRIMARY_PREFERRED
>                ),
>     ),

### DIFF
--- a/classes/mongo/db.php
+++ b/classes/mongo/db.php
@@ -155,6 +155,11 @@ class Mongo_Db
 		{
 			$options['replicaSet'] = $config['replicaset'];
 		}
+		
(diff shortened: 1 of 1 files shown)


sanzo83:
Thank you very much!

PR #2001, opened 2016-05-27
SQL statements use brackets

ishikawam:
fuelphp1.7, can use brackets for SQL statements.
But it doesn't work for above fuelphp1.8.

ex. (SELECT foo, ...) union (SELECT bar, ...)

https://github.com/fuel/core/commit/646240c631daa51777d0734ec61a348894c8a0ff#diff-30fb281dc6c1d9fc15067b17b7e3ac7cL273

This is repair commit for 1.9/develop.
And cherry-pick to 1.8/master please.

### DIFF
--- a/classes/database/query.php
+++ b/classes/database/query.php
@@ -271,7 +271,7 @@ class Database_Query
 		if (is_null($this->_type))
 		{
 			// get the SQL statement type without having to duplicate the entire statement
-			$stmt = preg_split("/[\s]+/", substr($sql, 0, 10), 2);
+			$stmt = preg_split("/[\s]+/", substr(ltrim($sql, '('), 0, 10), 2);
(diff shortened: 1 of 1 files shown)


WanWizard:
I can't merge this, this code will lead to copying the entire SQL statement in memory, which is a problem for large SQL strings (like bulk inserts).

I think it is better to reverse it, cut the first bit off, then do an ltrim.

```
$stmt = preg_split("/[\s]+/", ltrim(substr($sql, 0, 11), '('), 2);
```

ishikawam:
I see, fixed for large SQL strings.

WanWizard:
Thanks!

I'll try to push out an 1.8.0.1 hotfix over the weekend.

ishikawam:
Thank you! 😃

PR #2004, opened 2016-06-16
Check if the `Error` class is an ErrorHandler before aliasing it.

### LINKED ISSUE #2003: FuelPHP and PHP7.
Hi all, 

I am currently trying to upgrade an application from 1.7 to 1.8. However, as the upgrade includes the PHP7 `Error` class, I cannot finish the procedure. 

Let me explain, we are using a lot of differents components in our app and one of them pulls in the `symfony/polyfill7` package. That package aims to introduce the feature of PHP7 to prior PHP7 projects. It also includes an `Error` class. The problem is that Fuel aliases the `ErrorHandler` class to the `Error` class if the class is defined. 

The following code does that in `fuel/core/classes/fuel.php` : 

``` php
        // BC FIX FOR PHP < 7.0 to make the error class available
        if (PHP_VERSION_ID < 70000)
        {
            // alias the error class to the new errorhandler
            class_alias('\Fuel\Core\Errorhandler', '\Fuel\Core\Error');

            // does the app have an overloaded Error class?
            if (class_exists('Error'))
            {
                // then alias that too
                class_alias('Error', 'Errorhandler');
            }
        }
```

As the autoloading is managed by composer, the `Error` class that is being registered as `Errorhandler` is the symfony polyfill class. Meaning that when trying to call the `Errorhandler::shutdown_handler` function this fails. 

In my opinion, this was a mistake. Assuming the user defined a `Error` class extending the existing Error class and acting on that assumption makes it very dangerous code. Reflection would have been a much better choice here. Or at least check if the class has the `shutdown_handler` function before aliasing it.

iammichiel:
Fixes #2003

### DIFF
--- a/classes/fuel.php
+++ b/classes/fuel.php
@@ -204,10 +204,8 @@ class Fuel
 			// alias the error class to the new errorhandler
 			class_alias('\Fuel\Core\Errorhandler', '\Fuel\Core\Error');
 
-			// does the app have an overloaded Error class?
-			if (class_exists('Error'))
-			{
-				// then alias that too
+			$errorReflectionClass = new \ReflectionClass('Error');
+			if ($errorReflectionClass->hasMethod('shutdown_handler')) {
 				class_alias('Error', 'Errorhandler');
 			}
 		}


WanWizard:
This doesn't work. I think you miss the point here.

Fuel v1 aliases it's classes to the root namespace, and allows you to override that by defining your own class in your app, like so:

```
class Error extends \Fuel\Core\Error 
{
}
```

Your change breaks this functionality, and breaks every app thay uses it. This is unacceptable, so it can't be merged.

Also, we don't accept PR's on master branches (which contain release code), only on develop branches.

WanWizard:
Fixed in https://github.com/fuel/core/commit/e9da749e13d2725a3b3b6b79e2c4b8dc53b5ab82

PR #2015, opened 2016-08-19
Ability to stick the profiler (PQP_SHOWONLOAD)

filisko:
Hi,
So I just added the ability to stick the profiler (PQP_SHOWONLOAD) dynamically using a cookie, have a look!

Regards,
Filis.

### DIFF
--- a/vendor/phpquickprofiler/display.php
+++ b/vendor/phpquickprofiler/display.php
@@ -126,12 +126,13 @@ CSS
 	$return_output .=<<<JAVASCRIPT
 <!-- JavaScript -->
 <script type="text/javascript">
-	var PQP_SHOWONLOAD = (typeof PQP_SHOWONLOAD != "undefined" && PQP_SHOWONLOAD) ? "open" : "closed";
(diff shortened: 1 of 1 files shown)


filisko:
Woops, I will send another one in few minutes.

filisko:
Now is ok :+1:

WanWizard:
I am not sure having something client side trigger this is a good idea, it may be abused and expose a lot of internal information.

filisko:
how do other debug bars do it?

WanWizard:
What exactly do you want to archieve? Personally, I find it a bit pointless to have the profiler fill the entire screen on every page load, obscuring the page, only so you don't have to click the "show" link.

filisko:
It's only a toggle for PQP_SHOWONLOAD, so you could enable it and disable it without touching code.

WanWizard:
I got that, but I never understood why one would always want to do that. You look at the profiler when you need to, by clicking the link. Why do you need it to obscure your app output on EVERY page load?

filisko:
If you want to compare a lot of pages in a couple of seconds, it's faster than going to the page and having to open it every time. But hey, you're also right. And you can achieve this by putting some javascript there, which is nice.

PR #2016, opened 2016-08-29
changed $config to $this->_config as the assumption is that we want t…

fufu70:
Errors occurred when trying to call the list_tables function. The solution that I found was changing $config to $this->_config so as to access the local _config variable and set the dsn field for that variable.


WanWizard:
Good find!

fufu70:
ty :)

PR #2019, opened 2016-09-21
Corrected the configuration's DSN index in mysql/connection.php

cmark89:
This solves an undefined index notification that was occurring as of v1.8, even with the $this->config change. Looks like it was a simple oversight in 23e88d.


WanWizard:
Very well spotted, thanks!

PR #2021, opened 2016-09-21
Fix inifite loop if the connection is lost.

you21979:
Fix inifite loop if the connection is lost.

see original commit.
https://github.com/jdp/redisent/commit/8de5c6a9fc0869e3bc09869bf513577fb08cce3e


WanWizard:
Well spotted.

Can you recreate this PR against the 1.9/develop branch? We don't accept PR's against master branches.

WanWizard:
p.s. while you are at it, you might also want to implement https://github.com/jdp/redisent/commit/0a9e40710153a5f36fbafe7d1f7fe0de3f723be4 which is another update we missed.

PR #2031, opened 2016-10-12
Fixed issue with the Fuel\Core\Database_Query_Builder_Join::quote_ide…

fufu70:
…ntifier() error. As the quote_identifier method should be called from the Database_Connection ($db) object in Database_Query_Builder_Join class


WanWizard:
Oops. Thanks for finding it, and your PR!

PR #2034, opened 2016-11-07
String concatenation does not work in the initialization of static variables

fufu70:
When running in php 5.5.9 I receive the error:

Parse error: syntax error, unexpected '.', expecting ',' or ';' in /var/www/platform.beachfrontreach.com/fuel/core/classes/fuel.php on line 378

### DIFF
--- a/classes/fuel.php
+++ b/classes/fuel.php
@@ -375,7 +375,7 @@ class Fuel
 	{
 		// framework default paths
 		static $search = array('\\', APPPATH, COREPATH, PKGPATH, DOCROOT, VENDORPATH);
-		static $replace = array(DS, 'APPPATH'.DS, 'COREPATH'.DS, 'PKGPATH'.DS, 'DOCROOT'.DS, 'VENDORPATH'.DS);
+		static $replace = array(DS, 'APPPATH/', 'COREPATH/', 'PKGPATH/', 'DOCROOT/', 'VENDORPATH/');
 
 		// additional paths configured than need cleaning
 		$extra = \Config::get('security.clean_paths', array());


Art4:
This will not work on Windows systems. Alternative:

``` diff
        // framework default paths
        static $search = array('\\', APPPATH, COREPATH, PKGPATH, DOCROOT, VENDORPATH);
-       static $replace = array(DS, 'APPPATH'.DS, 'COREPATH'.DS, 'PKGPATH'.DS, 'DOCROOT'.DS, 'VENDORPATH'.DS);
+       $apppath = 'APPPATH'.DS;
+       $corepath = 'COREPATH'.DS;
+       $pkgpath = 'PKGPATH'.DS;
+       $docroot = 'DOCROOT'.DS;
+       $vendorpath = 'VENDORPATH'.DS;
+       static $replace = array(DS, $apppath, $corepath, $pkgpath, $docroot, $vendorpath);

        // additional paths configured than need cleaning
        $extra = \Config::get('security.clean_paths', array());
```

fufu70:
I'd be down with that :) Should I change the pull request?

Art4:
I'd suggest it, but that's up to @WanWizard.

WanWizard:
Where does this come from all of a sudden? I don't have that error on 5.6.27. Some PHP weirdness?

```
[wanwizard@catwoman] $ php -l fuel/core/classes/fuel.php 
No syntax errors detected in fuel/core/classes/fuel.php
[wanwizard@catwoman] $ php fuel/core/classes/fuel.php 
[wanwizard@catwoman] $
```

What paths are displayed in for example the profiler when on Windows?

Art4:
Good question. I can't reproduce it. This even worked in PHP 4.3
https://3v4l.org/ZO4Ec#version=5.5.9

WanWizard:
you missed the static, but even then no version gives this error: https://3v4l.org/QkABL

edit: oops, forgot EOL versions. I looks like all version up until 5.5.38 show this error, 5.6+ not. Since Fuel is 5.3.3+, it needs to be addressed. I'll revert it back to the hardcoded slash.

Issue #1833, opened 2015-02-12
[FEATURE] Logging context and Monolog format handling


rickmacgillis:
Monolog is used for FuelPHP's logging system backend, and Monolog supports arbitrary context data. (An array of arbitrary data with information about what caused the logging event) Both `\Log::write()`, and `logger()` interface directly with Monolog's [`\Monolog\Logger::log()`](https://github.com/Seldaek/monolog/blob/master/src/Monolog/Logger.php#L439) method to write entries.

**Proposed Change**
Alter `\Log::write()` and `logger()` to accept a fourth parameter for $context, and pass it along to `\Monolog\Logger::log()` for processing.

It would also be ideal if you could separate `\Log::_init()` into a secondary method for handling exactly how [logs are stored through Monolog](https://github.com/fuel/core/blob/1.8/develop/classes/log.php#L99). That way people can extend the \Log class, and override the new method without having to call `parent::_init()` to keep the directory structure, while needlessly setting up the original Monolog recording methods. (See: http://fuelphp.com/docs/general/extending_core.html#extend_and_replace)

emlynwest:
While a perfectly valid idea I feel that any new major changes should be left out of v1 now in favour of working on v2. I'd rather that our efforts went into the new version of the framework so we can feature freeze v1.

@WanWizard thoughts?

WanWizard:
I'll have to look at the impact and the amount of work. A PR would be handy.

Since it functionally doesn't change anything, if it's not a lot of work, it could be part of the upcoming1.7.3.

I agree major v1 should be frozen functionally, I think a last release will be a good moment for that.

PR #2046, opened 2017-01-29
Allow additional monolog context to be passed via Log::write

Josh-G:


Josh-G:
This also relates to #1833

PR #2047, opened 2017-02-02
Fix Composer dependency issue

jaymoulin:
Allow developers to use any compatible version of Fuel Core dependencies instead of specified ones (for example, monolog 1.22.X instead of monolog 1.18.X)

Merging this in 1.8.0 branch will allow us to upgrade to 1.8.0.4 instead of keeping stuck at 1.8.0.3

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,8 +12,8 @@
 	],
 	"require": {
 		"composer/installers": "~1.0",
-        "phpseclib/phpseclib": "2.0.0",
-        "monolog/monolog": "1.18.*",
-        "michelf/php-markdown": "1.*"
+		"phpseclib/phpseclib": "^2.0.0",
+		"monolog/monolog": "^1.18.0",
+		"michelf/php-markdown": "^1.0"
 	}
 }


emlynwest:
I think the versions where constrained like that to prevent breaking changes @WanWizard to confirm?

I'm also closing this as we don't accept pull requests in to master branches, please re-open against `1.9/develop`

jaymoulin:
IMHO it's what Composer is meant to do

WanWizard:
Monolog is pinned on this version for a simple reason: Fuel nas not been tested with other versions, and we're not going to change the version without making sure it doesn't break anything. Newest is not always best.  Same is true for the other dependencies.

I'm afraid you miss the point of composer. Composer deals with dependency management, it doesn't fix compatibility issues. If you manually change versions in your composer.json, you break dependencies we have set. 

You may debate that composer package maintainers must make sure there is compatibility between minor versions (and you may have a point), but it is our job to keep this framework stable, we can not delegate that responsibility (sorry, Fuel doesn't work anymore, go complain with the Monolog guys! Doesn't work, don't you think?).

I fail to see the relation between this and updating the Fuel core. That is only an issue if you have made manual changes. In which case you have created the problem, not we. ;-)

WanWizard:
p.s. If this is an akward way of asking us to upgrade some dependencies, please create a new issue for that, so we can have a look at it, and properly test it.

jaymoulin:
I think I am mistaken.
The problem we have is not with Monolog but the use of Monolog inside FuelPHP.

This PR follows indications defined in semver manifest (http://semver.org/) that packages like Monolog (also Composer (same creator BTW)) follows. 

There is no reason ^1.18 rule break anything with FuelPHP : https://getcomposer.org/doc/articles/versions.md#caret

Art4:
I think @WanWizard gets you right, but he sets great value on stability and BC (what I am very thankful for 😊). Semver is only a manifest but not a guaranty that nothing will break with a higher minor version. Not every library follows semver or follows it strictly (but I assume, Monolog is doing it very well).

So to be sure there is no instability or BC-break you can only test one (or multiple) version and pin composer to it. Like @WanWizard said:

> Fuel nas not been tested with other versions, and we're not going to change the version without making sure it doesn't break anything.

A fair notice to this sentence: To be consequent and sure that a bugfix on Monolog 1.18 will not break something, composer should pin to `1.18.0` instead of `1.18.*` like it is now.

The downside is that this strict pinning makes it impossible to use other libraries with a dependency to Monolog 1.19 or higher. Same problem with phpseclib and php-markdown.

So I'm with you, @jaymoulin and would recommend to use the caret.

Art4:
Oh, and as a sidenote: 

> sorry, Fuel doesn't work anymore, go complain with the Monolog guys! Doesn't work, don't you think?

From the view of a developer this works for me. I can either report a bug (and help the library getting better) and/or try to use an older version of Monolog. :-)

WanWizard:
From the point of a developer, I agree. But as Fuel Project Lead, I agree with your previous reply.

If someone using Fuel updates, and their apps break because Monolog was updated, they will all come and complain here. It is true that in that case Monolog has broken the semver rules, but that doesn't solve the problem for people that now have a broken framework.

So we have decided to pin external dependencies, and once in a while check to see if an upgrade is needed, and if so, bump the version in the composer.json.

If on the other hand, you have a project that requires a newer version, you are free to bump the version yourself. If you then report back to us by optioning an issue and saying "he, there's a new version, I've tested it, and it works fine", we can quite quickly bump the pinned version.

jaymoulin:
> If on the other hand, you have a project that requires a newer version, you are free to bump the version yourself. If you then report back to us by optioning an issue and saying "he, there's a new version, I've tested it, and it works fine", we can quite quickly bump the pinned version.

We're not able to bump our version due to this 1.18.* constraint. I actually use ^1.22.0 because we needed a fix that appeared in this version and it works great but I had to downgrade fuel-core to 1.8.0.3 because this constraint doesn't exist in this version

Art4:
What about this:

1. In [fuel/core](https://github.com/fuel/core/blob/1.8/master/composer.json) you define the monolog (and other libraries) with a caret `^1.18.0`.
2. In [fuel/fuel](https://github.com/fuel/fuel/blob/1.8/master/composer.json) you add Monolog and pin it to `1.18.0`. 

This way composer will decide to use Monolog 1.18 by default, but the developer has the chance to change it to the version he likes.

WanWizard:
@jaymoulin,

 If you use 1.22 without problems, that is all I need to know, then I'll update the dependency.

@Art4, that doesn't work:
````
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - fuel/core dev-1.9/develop requires monolog/monolog 1.18.* -> satisfiable by monolog/monolog[1.18.0, 1.18.1, 1.18.2] but these conflict with your requirements or minimum-stability.
    - fuel/core dev-1.9/develop requires monolog/monolog 1.18.* -> satisfiable by monolog/monolog[1.18.0, 1.18.1, 1.18.2] but these conflict with your requirements or minimum-stability.
    - Installation request for fuel/core dev-1.9/develop -> satisfiable by fuel/core[dev-1.9/develop].
````
(when you add for example `^1.20.0` to the main composer.json file)

edit: @Art4 ok, now get what you meant.

**Comment by Art4** [comment]
@WanWizard I think you have first to make the change in fuel/core and bump the version to 1.8.0.5. Then in fuel/fuel you can pin fuel/core to 1.8.0.5 and add monolog.

**Comment by Art4** [comment]
I can't try it at the moment, sorry. I have to fork fuel/fuel and fuel/core, change the composer.json to my fork-repositories and create the new release.

**Update:** It works. Thanks to your commit I only have to change the dependency to
```
"fuel/core": "dev-1.9/develop",
"monolog/monolog": "^1.22.0"
```

and removed the minimum-stability.
This works like expected:

```
$ php composer.phar update
Loading composer repositories with package information
Updating dependencies (including require-dev)
  - Removing monolog/monolog (1.18.2)
  - Installing monolog/monolog (1.22.0)
    Loading from cache

  - Removing phpseclib/phpseclib (2.0.0)
  - Installing phpseclib/phpseclib (2.0.4)
    Downloading: 100%

  - Removing fuel/core (1.8.0.4)
  - Installing fuel/core (dev-1.9/develop 727dabc)
    Cloning 727dabc2ccc803ff23854f0ce8f25eb82b6c724c from cache

Writing lock file
Generating autoload files

```

**Comment by WanWizard** [comment]
No worries, I'll figure it out. Have some issues with stability at the moment, when you do this and want dev versions of Fuel, you also get dev versions of these dependencies, instead of the stables...

**Comment by Art4** [comment]
With the latest commit  dcf5790db8638 everything works as expected. I see no problems with the minimum-stability and can update to Monolog 1.22.0 👍 

Next step would be creating release fuel/core 1.8.0.5 and update composer.json in fuel/fuel.

**Comment by WanWizard** [comment]
The problem is when you have "dev" as minimum stability (which we have when we work on the framework), you get "1 x-dev" from monolog, instead of the latest stable. For that you need an entry in fuel/fuel, which I have now added.

https://github.com/fuel/fuel/commit/fc48c0c9920bfef12296f1a54ca2fdeafe14e862

(note, this will only be needed in the dev branch, and can be removed when merged with master)

**Comment by Art4** [comment]
@WanWizard Maybe a better solution for the develp branch would be:
- set minimum-stability to stable
- set the fuel-dependencies to `@dev`

This will save you some work on a release, because you only have to touch the fuel-dev branches.

I've tested it and it works.

```
{
    ...
    "require": {
        "php": ">=5.3.3",
        "composer/installers": "~1.0",
        "fuel/core": "dev-1.9/develop@dev",
        "fuel/auth": "dev-1.9/develop@dev",
        "fuel/email": "dev-1.9/develop@dev",
        "fuel/oil": "dev-1.9/develop@dev",
        "fuel/orm": "dev-1.9/develop@dev",
        "fuel/parser": "dev-1.9/develop@dev",
        "phpseclib/phpseclib": "2.0.*",
        "monolog/monolog": "1.18.*",
        "michelf/php-markdown": "1.7.0",
        "fuelphp/upload": "2.0.6"
    },
    ...
    "minimum-stability": "stable"
}
```

jaymoulin:
https://github.com/fuel/core/pull/2048

WanWizard:
It doesn't make any difference, since the develop will always be merged to a master before a release. So we either have to remove the current "@stable" additions during that mergem or remove all these "@dev"'s, the work remains the same: we need to update the composer.json.

Art4:
Sure, but it helps you from making errors in the future. A merge will affect less lines to change and you can't forget to change the stability. Also you can specify precise which project should be dev. So you cannot forget to set the stable flag it if you add a new dependency in the future. 

Am 3. Februar 2017 18:26:44 MEZ schrieb Harro Verton <notifications@github.com>:
>It doesn't make any difference, since the develop will always be merged
>to a master before a release. So either remove the current "@stable"
>additions during that mergem or remove all these "@dev"'s, the work
>remains the same: we need to update the composer.json.

WanWizard:
Hmm... More changes...

WanWizard:
Done. This change works as expected, the additional `@dev` prefix was not required.

Art4:
Even better. :+1:

PR #2048, opened 2017-02-02
Fix Composer dependency issue

jaymoulin:
Allow developers to use any compatible version of Fuel Core dependencies instead of specified ones (for example, monolog 1.22.X instead of monolog 1.18.X)


xploSEoF:
:+1: for fixing the indentation!

The only worry would be any breaks to BC with minor releases, but that shouldn't be the case with any composer package.

PR #2054, opened 2017-03-17
Make create_dir only attempt to create directories under the basepath.

ghost:
We've run into an issue with `create_dir` and PHP's open_basedir when given params like:

```
$basepath = "/var/www/<sitedir>/fuel/core/logs/";
$name = "2017/03";
```

With open_basedir set to `/var/www/<sitedir>` this fails because create_dir tries to run `is_dir('/var')` - which is outside of the open_basedir paths.

This fix changes create_dir behaviour so that it only runs is_dir and mkdir on paths under the provided basepath - which it has already checked for existence and writability, anyway.


WanWizard:
Good find! Not 100% sure about the proposed solution though, I need to have a look at it.

PR #2064, opened 2017-06-19
fix overwriting the alpha value in imagick.php

j5a:
fix overwriting `$alpha` in classes/image/imagick.php

this fix is
* refert to https://github.com/fuel/core/commit/31209799ed0333ca8128163e585c18f41b22ab98
* for 1.9/develop branch.


WanWizard:
Thanks!

Issue #1973, opened 2016-02-06
Sanitizaion should be on only during View output


kenjis:
If we pass an instance which implements Sanitizaion to a view, the View object (output encoding) turns on sanitization automatically.

But it does not turns it off. So after rendering, sanitization of the instance is still enabled.

The below code shows how it works:

``` php
        // Orm\Model implements Sanitization interface
        $user = new Model_User();
        $user->id = 1;
        $user->username = '<s>abc</s>';
        $user->group = 1;

        // Not escaped
        var_dump($user->username);

        $data['user'] = $user;
        $view = View::forge('sanitization/index', $data);

        // Output filter htmlentities automatically turns on sanitization
        echo $view;

        // Escaped because sanitization is still on
        var_dump($user->username);

        // If we want true value, we have to turn off sanitization manually
        $user->unsanitize();
        // Not escaped
        var_dump($user->username);
```

If we save the instance in Session, we get the instance with sanitizaion enabled when next request.
Then if you get `$user->username`, you get HTML-escaped value.

The behaviour is difficult to understand. So I think we should turn off sanitization after view rendering.

WanWizard:
The View calls `Security::clean()` on all it's variables, which in turn cleans the variables and enables sanitation.  Which means that it will require quite a bit of processing to disable sanitation. 

Given the fact that I think your use case is quite limited, as for most people the View is the end of the request, I don't want to do this by default (which would also break BC).

I'll have to give this some thought. @stevewest, what is your opinion?

emlynwest:
If we can I think that we should turn off the sanitization after the view has been rendered. In quite a few of my projects I've constructed views in some "controller/helper" code and passed that through to a parent view, going on to reuse objects for other things later on.

WanWizard:
I'll see if I can come up with something. I would like to control it through config, so you can enable this when needed, both to not break BC, and to avoid a performance impact for those who don't need this.

WanWizard:
BTW, I now see that `Security::clean()` is not recursive, so if you pass a multi-dimensional array structure, it will only get a surface clean, not a deep clean. That doesn't seem to be correct?

WanWizard:
Changing the View behavior might have side-effects. Data is being cleaned in the `get_data()` method, which by default is protected and only called before rendering the view. But in case of overloading, `get_data()` could be called without rendering the view. 

So calling unsanitize() after rendering wouldn't have the desired effect in these circumstances.

WanWizard:
[...]

PR #2065, opened 2017-07-12
fix not to override view variable with empty value in `View::unsanitize()`

xKerman:
related to: #1973 

If view variable implements `Traversable` and `ArrayAccess` interface, the variable is processed in `View::unsanitize()` with the code below:

https://github.com/fuel/core/blob/ae82e94023f2ca9fb279b0b2e19a13e5c459d3d1/classes/view.php#L343-L351
```
		// deal with array's or array emulating objects
		elseif (is_array($var) or ($var instanceOf \Traversable and $var instanceOf \ArrayAccess))
		{
			// recurse on array values
			foreach($var as $key => $value)
			{
				$var[$key] = $this->unsanitize($value);
			}
		}
```

Since `View::unsanitize()` does not return value currently, this process overrides `$var` entry with empty value.

So, If the variable is shared among two views, first view is rendered with given variable, but second view is rendered with empty variable.

This pull request changes `View::unsanitize()` behavior to return processed `$var`, in order to address the problem above.

### DIFF
--- a/classes/view.php
+++ b/classes/view.php
@@ -349,6 +349,8 @@ class View
 				$var[$key] = $this->unsanitize($value);
 			}
 		}
+
+		return $var;
 	}
 
 	/**
--- a/tests/view.php
+++ b/tests/view.php
@@ -12,6 +12,41 @@
 
 namespace Fuel\Core;
 
+class Arraylike implements \ArrayAccess, \IteratorAggregate
+{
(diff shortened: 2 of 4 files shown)


WanWizard:
Very well spotted, thanks!

xKerman:
Thanks for your review!

PR #2071, opened 2017-10-02
Input_Instance->param doesn't work without HTTP Parameter

saitara:
A code `Input_Instance->param('size', 1)` without http PUT parameters doesn't work w/ PHP7.1.

```
#2 /usr/share/nginx/mandarin.io/fuel/core/classes/input/instance.php(302): array_merge(Array, Array, '', Array, Array) 
```

To fix that, cast as an array for every parameters in function all().

### DIFF
--- a/classes/input/instance.php
+++ b/classes/input/instance.php
@@ -299,7 +299,7 @@ class Input_Instance
 	 */
 	public function all()
 	{
-		return array_merge($this->input_get, $this->input_post, $this->input_put, $this->input_patch, $this->input_delete);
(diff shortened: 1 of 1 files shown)


WanWizard:
This fix is not correct, as it doesn't fix the root cause.

All these properties are defined as array's, and `hydrate()` should make sure they stay array's. So I suspect there is a parsing error in `hydrate()`, possibly related to using nginx, that needs debugging.

saitara:
Thanks, got it. Close this PR.

WanWizard:
Let us know if you find the root cause, I don't gave access to an nginx setup at the moment.

PR #2080, opened 2017-12-31
Update router.php

ghost:


WanWizard:
Can you explain why you feel this is needed? What use case does this address? Can you give a code example?

JefferyHus:
a commit without explanation is nothing honestly, please explain your case, as it may work for you but not the rest.

ghost:
@WanWizard @JefferyHus Why in the foreach loop many times declare the $ replace variable? I think that this will be better.

JefferyHus:
if the `$replace` is affected in the first condition inside `foreach` it will remain with the same value in the next loop, so if the first condition isn't fired in the second loop there will be an issue as the second route even without params it will execute it with the previous `$replace` value, therefor you should refresh the value each time you loop.

Issue #2081, opened 2018-01-05
Error


it-can:
On php7.2 I get this error:

```
Fuel\Core\PhpErrorException [ Warning ]:
get_class() expects parameter 1 to be object, null given

COREPATH/classes/security.php @ line 241
```

WanWizard:
Yup:

> Note: Explicitly passing NULL as the object is no longer allowed as of PHP 7.2.0. The parameter is still optional and calling get_class() without a parameter from inside a class will work, but passing NULL now emits an E_WARNING notice.

WanWizard:
This might pop up in other locations, the codebase hasn't been screened or tested for 7.2 yet.

PR #2082, opened 2018-01-17
Null don't have to be escaped in Security::htmlentities()

Art4:
I've stumbled upon the "get_class() expects parameter 1 to be object, null given" warning in PHP 7.2 and I have see #2081, but my PR is imho a faster solution than https://github.com/fuel/core/commit/8a44a112c7a9a7eee54f6990a9913daf96a230a1, because `null` will be immediately returned instead of go through all `if` cases.

### DIFF
--- a/classes/security.php
+++ b/classes/security.php
@@ -213,7 +213,7 @@ class Security
 		is_null($double_encode) and $double_encode = \Config::get('security.htmlentities_double_encode', false);
 
 		// Nothing to escape for non-string scalars, or for already processed values
-		if (is_bool($value) or is_int($value) or is_float($value) or in_array($value, $already_cleaned, true))
+		if (is_null($value) or is_bool($value) or is_int($value) or is_float($value) or in_array($value, $already_cleaned, true))
 		{
 			return $value;
 		}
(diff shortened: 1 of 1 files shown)


WanWizard:
Very valid point.

hoangnkvti:
image

I got this error, so I don't now why you remove null check in line 241 of Security class.

Art4:
I moved it to line 216 because `null` don't has to be sanitized and could be instantly returned.

Could you please tell us your PHP version and Fuel version?

hoangnkvti:
I am using `fuel 1.9`
image

And `php7.2`
image

Art4:
**Update:** Ok, I found the problem. You are using an old fuel/core version. Commit 27eef67e341dfd6720cba57001b3d89285dec59e was made Jul 5, 2017, the fixes for this issue were made in Jan 2018.

If you are using dev-1.9 you should remove the specific commit from your `composer.json` and run `composer update` frequently.

hoangnkvti:
image
Maybe this one

Art4:
@hoangnkvti please see my updated comment above.

hoangnkvti:
Thank you very much. I fixed my issue 👍

PR #2085, opened 2018-02-05
Fix - Issue of URI::update_query_string()

yama:
https://github.com/fuel/core/commit/35e65d649a6fcdad8a4e22ef026f8f3a2d0de6c4#commitcomment-24850312
I also suffered from this problem.
I think, explicitly with the trailing slash, it is necessary to reproduce this.
I thank FuelPHP. It's very good.


WanWizard:
Why do you need the trailing slash here? What goes wrong if it isn't there?

I want to understand why you want to revert to incorrect URI generation.

PR #2094, opened 2018-05-14
Allow higher versions of monolog/monolog

Art4:
I can't update fuel/core to 1.8.0.4 or 1.8.1.* because of a dependency that requires monolog/monolog ~1.22.
This PR will allow higher versions of monolog without BC breaks.


Art4:
Thanks :tada: I'm looking forward for fuel/core 1.8.1.4 :blush:

PR #2108, opened 2019-01-10
using symbol for boundary is not accepted

urbsny:
Sorry for bothering you.
If i use '+++++' for boundary, preg_split method is abended.
I suggest to use preg_quote.


WanWizard:
Sorry, can't accept this one, please issue PRs on the current develop branch only (now 1.9/develop).

PR #2110, opened 2019-01-10
using symbol for boundary is not accepted

urbsny:
Thank you for your quick reply and sorry for my misunderstanding.
Same suggestion with [#2108](https://github.com/fuel/core/pull/2108).


WanWizard:
Thanks!

PR #2112, opened 2019-01-22
allow to use ORDER BY `group`

urbsny:
I'm sorry to bother you. and thank you for your awesome framework.
If I use `group` in the key of ORDER BY, it's become syntax error.

before preg_replace
```
ORDER BY `group` ASC
```
after preg_replace
```
group` ASC
```

I propose to change the pattern.
I refered these select syntax.
https://dev.mysql.com/doc/refman/5.5/en/select.html
https://www.postgresql.org/docs/9.5/sql-select.html


WanWizard:
Very well spotted, thanks for your contribution !

PR #2114, opened 2019-03-06
Add a "float" type for routes

paullb514:
It would be nice to have a "float" type for routing. The "num" doesn't allow decimal points and it needn't be an entire segment that can grab everything.

### DIFF
--- a/classes/route.php
+++ b/classes/route.php
@@ -115,6 +115,7 @@ class Route
 			':everything',
 			':alnum',
 			':num',
+			':float',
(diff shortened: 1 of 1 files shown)


WanWizard:
It would, yes. But that isn't really a regex for a float, that would also match ",.,.,.,.".

A better one would be something like: ````^[-+]?[0-9]*[\.,]?[0-9]+([eE][-+]?[0-9]+)?$```` ?

PR #2115, opened 2019-05-06
Fix warning for PHP 7.3

pfuhrmann:
We encountered following error in the Fuel Core 1.8
selection_096

This is due to BC changes from `PHP 7.2` to `7.3`:
https://www.php.net/manual/en/migration73.incompatible.php
```
continue statements targeting switch control flow structures will now generate a warning.
In PHP such continue statements are equivalent to break, while they behave as continue 2 in other languages.
```

Not sure if to target `1.8` or `1.9` but it would be nice to maybe release this in `1.8.?`minor branch also.

### DIFF
--- a/classes/session/memcached.php
+++ b/classes/session/memcached.php
@@ -320,7 +320,7 @@ class Session_Memcached extends \Session_Driver
 
 						default:
 							// unknown property
-							continue;
+							continue 2;
 					}
 
 					$validated[$name] = $value;


WanWizard:
Thanks, there might be more of these. We're were planning a full app test on 7.3, now that Fedora 30 is out (which we use as our dev platform). I'll see if we can speed things up, and release a new 1.8 minor, I think there are other fixes in 1.9/dev that need releasing.

PR #2120, opened 2019-06-04
Added $spec parameter in Security::xss_clean

Thunder-Dan:
Since xss_clean is a wrapper for HTMLawed, it would be nice to be able to pass the $spec parameter also.

### DIFF
--- a/classes/security.php
+++ b/classes/security.php
@@ -167,7 +167,7 @@ class Security
 		return $var;
 	}
 
-	public static function xss_clean($value, array $options = array(), $spec = null)
+	public static function xss_clean($value, array $options = array(), $spec = '')
 	{
 		if ( ! is_array($value))
 		{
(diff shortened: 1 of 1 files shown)


classes/security.php
```diff
@@ -176,7 +176,7 @@ public static function xss_clean($value, array $options = array())
 				import('htmlawed/htmlawed', 'vendor');
 			}
 
-			return htmLawed($value, array_merge(array('safe' => 1, 'balanced' => 0), $options));
+			return htmLawed($value, array_merge(array('safe' => 1, 'balanced' => 0), $options), spec);
```
WanWizard: Shouldn't this be `$spec` instead of `spec`?

Thunder-Dan: OFC! Sorry! Keep up the good work!

PR #2130, opened 2019-09-29
FIX bug Core\Package::load() $result always true when $package is array

Illirgway:
$result = true;
$result = $result && static::load($pkg, $path);

=> $result is always true regardless of load subcall return val, due to ops priority: prio('and') < prio('=') < prio('&&')

for ex. http://sandbox.onlinephpfunctions.com/code/e3faf45b7ba1bf8b553d527cb2d1a781efbad057

```php
$result = true and false;
var_dump($result);  // bool(true)

$result = true && false;
var_dump($result);  // bool(false)
```

### DIFF
--- a/classes/package.php
+++ b/classes/package.php
@@ -55,7 +55,7 @@ class Package
 					$pkg = $path;
 					$path = null;
 				}
-				$result = $result and static::load($pkg, $path);
+				$result = $result && static::load($pkg, $path);
 			}
 			return $result;
 		}


Illirgway:
https://www.php.net/manual/en/language.operators.precedence.php

WanWizard:
This is not correct.

It should be "and", i.e. it should only execute the load() when `$result` is false.

You will find a lot of these in the Fuel code, it is a shortcut for
````
if ( ! $result)
{
    $result = do_something_here();
}
````

**Comment by Illirgway** [comment]
But result is always be true due to priority: ($result = $result /* true */) and fncall()

code slice with commented exec flow 

```
		if (is_array($package))
		{
			$result = true;                    // (1) <== first time init to true
			foreach ($package as $pkg => $path)
			{
				if (is_numeric($pkg))
				{
					$pkg = $path;
					$path = null;
				}
				$result = $result and static::load($pkg, $path); // (2) never changed due to < ($result = $result) and fncall() > {useless "$result = $result" === true} == simple equal ==> static::load($pkg, $path);
			}
			// (3) here $result alway be true due to (2) in any case
			return $result;
		}
```

**Comment by WanWizard** [comment]
Hmm, missed (1).

I now understand what the code is about: `$result` should only be true if all packages were loaded correctly, which means it should be `or`, not `and`.

So the first load that fails flips the result from true to false, and because of the `or`, once false it stays false.

**Comment by Illirgway** [comment]
Additionally, maybe we should do a "break & return false" immediately on the first package load failure?

**Comment by WanWizard** [comment]
No,  we still want to load the rest. 

Whether or not it is a permanent failure depends on the application and the application design, changing this will break existing code that tries to load optional packages.

**Comment by Illirgway** [comment]
then should rewrite that as
```
// load() always fire, accumulate with prev result status, honor ops prios
$result = static::load($pkg, $path) && $result; 
```

should I pull a new commit ?

WanWizard:
Yes please, but with `and` and not `&&`.

PR #2133, opened 2019-10-02
FIX bug Core\Package::load() $result always true when $package is array

Illirgway:
$result = true;
$result = $result and static::load($pkg, $path);

=> $result is always true regardless of load subcall return val, due to ops priority: prio('and') < prio('=') < prio('&&')

for ex. http://sandbox.onlinephpfunctions.com/code/e3faf45b7ba1bf8b553d527cb2d1a781efbad057

```php
$result = true and false;
var_dump($result);  // bool(true)

$result = true && false;
var_dump($result);  // bool(false)
```

Also **MUST** use brackets due to ops priorities

### DIFF
--- a/classes/package.php
+++ b/classes/package.php
@@ -55,7 +55,9 @@ class Package
 					$pkg = $path;
 					$path = null;
 				}
-				$result = $result and static::load($pkg, $path);
+				// MUST use external brackets due to prio('and') < prio('=') < prio('&&') - 
+				// don't remove them xor replace 'and' with '&&'
+				$result = (static::load($pkg, $path) and $result);
 			}
 			return $result;
 		}


WanWizard:
hmmm, that might go wrong in more places then, due to the prio of = vs and.

And I think I need a holiday, I don't know what all my rambling about this topic was about... :cry: 

Obviously, this is a boolean operation, so && would have been better. But brackets work to, and make maybe more clear what was intended.

Illirgway:
ideally, should gradually replace "and" by "&&" everywhere where it may be collisions, due to brackets is the implicit way to fix prio('and") if not provided with comment

WanWizard:
Agreed. Which is basically everywhere where there is an assignment. Things like
````
$result or $result = something();
$result and do_something();
````
should be fine.

**Comment by Illirgway** [comment]
should i start that work from this point and replace '(' 'and' ')' with simply "&&" ?

N.B. it is a strange thig, that prio("or") < prio("="), but in
``` $result or $result = something(); ```
"or" executed first

http://sandbox.onlinephpfunctions.com/code/3f7098a026f1281562dfb54d416d9298ad53ac62
```
function check($s)
{
    print_r($s . PHP_EOL);
    
    return new \stdClass;
}

$r = true;
$r or $r = check("invalid call");

var_dump($r);
```
```
bool(true)
```

WanWizard:
That seems to be correct, if "$r" is true, the `or` should cause the rest of the statement to be skipped. Likewise with:
````
$r = false;
$r and $r = check('valid call');
````

These "tricks" are even documented, https://fuelphp.com/docs/general/coding_standards.html, under "Alternative if statements"

If there were changes in precendence order between 5.3 (when this code was created) and 7.x, this might work unreliable now, which means we're in big trouble... Altough I have to say I haven't noticed this on our apps, and we run them on 7.3 in develop.

Maybe, if all code has to be checked anyway, it might be better to get rid of it all together, and just write compile `if` structures:
````
$r = true;
if ( ! $r)
{
    $r = check('no');
}
$r = false;
if  ( ! $r)
{
    $r = check('yes');
}
````

Illirgway:
I think brackets there will be enough

PR #2134, opened 2019-10-06
FIX Asset_Instance::render - unnecessary replace '\' with '/' in css/js when inlining asset file

Illirgway:
If I understand correctly, this code is intended to convert the Windows path to uri path:
```php
$file = str_replace('\\', '/', $file)
```

But if we already have inline contents in $file (L420, L437), then we replace backslash inside of it loaded content

### DIFF
--- a/classes/asset/instance.php
+++ b/classes/asset/instance.php
@@ -449,7 +449,7 @@ class Asset_Instance
 			}
 
 			// deal with stray backslashes on Windows
-			$file = str_replace('\\', '/', $file);
+			$inline or ( $file = str_replace('\\', '/', $file) );
 
 			// call the renderer for this type
 			if (isset($this->_renderers[$type]))


WanWizard:
No, if $inline is true, the replacement is skipped?

Illirgway:
currently, replace always performs for all cases

```php
			// === IF BEGIN
			// only do a file search if the asset is not a URL
			if ( ! preg_match('|^(\w+:)?//|', $filename))		
			{
				// and only if the asset is local to the applications base_url
				if ($this->_always_resolve or ! preg_match('|^(\w+:)?//|', $this->_asset_url) or strpos($this->_asset_url, \Config::get('base_url')) === 0)
				{
					if ( ! ($file = $this->find_file($filename, $type)))
					{
						if ($raw or $inline)
						{
							$file = $filename;	// (1) :: $file now is filename
						}
						else
						{
							...
						}
					}
					else
					{
						if ($raw or $inline)
						{
							$file = file_get_contents($file);	// (2) :: $file now is content
							$inline = true;				// $inline = true
						}
						else
						{
							$file = $this->_asset_url.$file.($this->_add_mtime ? '?'.filemtime($file) : '');
							$file = str_replace(str_replace(DS, '/', DOCROOT), '', $file); // (3) :: $file now is path
						}
					}
				}
				else
				{
					// a remote file and multiple paths? use the first one!
					$path = reset($this->_asset_paths[$type]);
					$file = $this->_asset_url.$path.$filename;
					if ($raw or $inline)
					{
						$file = file_get_contents($file);		// (4) :: $file now is content, 
						$inline = true;					// $inline = true
					}
					else
					{
						$file = str_replace(str_replace(DS, '/', DOCROOT), '', $file); // (5) :: $file now is path
					}
				}
			}
			else
			{
				$file = $filename;	// (6) :: $file now is url
			}				
			// === IF END

			// deal with stray backslashes on Windows
			$file = str_replace('\\', '/', $file);		// for (1-6) ===> should check $inline and pass this for (2) and (4)
```

WanWizard:
You are absolutely right, don't know where my brain has been hiding recently... :disappointed:

Illirgway:
I only don't exactly sure for (1) case, should we replace \ with / or not, due to we have ($row or $inline) === true, but $file = $filename in this case (this case fires only if $filename is not abs URL, and we don't find file in local assets)...

WanWizard:
Always, in case $inline is false, as it is either inline code, or a URL of some sort. 

Oldly enough, in (3) and (5) the replacement is done on DOCROOT (using DS instead of the hardcoded double backslash), but not on $file.

Illirgway:
In (1) case $inline may be true, but $file is $filename, not file content

```php
					if ( ! ($file = $this->find_file($filename, $type)))
					{
						if ($raw or $inline)
						{
							$file = $filename;	// (1)
						}
```

then we pass replace \ with / in this case

WanWizard:
Yes, but as I understand the code, in case a local find fails, it will be assumed to be some link to a CDN (for example), so it will be returned to be used as (part of) a URL, in which case only forward slashes are allowed.

PR #2136, opened 2019-10-08
generalize fix #2135 for appropriate DB drivers

Illirgway:
- next() should return next row (or null if out of result), as do Database_MySQLi_Result::next() (https://www.php.net/manual/en/mysqli-result.fetch-array.php "Returns an array of strings that corresponds to the fetched row or NULL if there are no more rows in resultset.") not only for mysqli, but for all appropriate drivers, e.g. PDO driver and all similar future drivers
- current() - rewind() now sanitize data in next() (rewind() -> next()) in appropriate drivers
- deduplicate sanitization code as helper fn of base abstract Result class \Database_Result

### DIFF
--- a/classes/database/mysqli/cached.php
+++ b/classes/database/mysqli/cached.php
@@ -118,13 +118,11 @@ class Database_MySQLi_Cached extends \Database_Result implements \SeekableIterat
 			$this->_row = $this->_results[$this->_current_row];
 
 			// sanitize the data if needed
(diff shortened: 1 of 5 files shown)


WanWizard:
On a roll... :+1:

Illirgway:
I think, in next refactor turn it needs one more abstraction class between \Database_Result and cache classes, say \Database_Result_Cache, where it should be deduplicated current(), next() and all similar fns from drivers concrete cache classes

PR #2144, opened 2020-07-09
Fix memcached cache delete_all()

### LINKED ISSUE #2143: \Cache::delete_all() using memcached driver not clearing all items
I just ran into an issue where delete_all wasn't working as I expected.

The first argument to Cache::delete_all() is `$section`, with the following description:
> Name of a section or directory of the cache, null to delete everything

On **fuel/core 1.8.2**, I'm getting this behavior using the memcached driver:

```php
\Cache::set('section_name.item', 'value');
\Cache::delete_all(); // expected to remove everything
\Cache::get('section_name.item'); // 'value'

\Cache::delete_all('section_name');
\Cache::get('section_name.item'); // null
```

Looking at https://github.com/fuel/core/blob/1.8.2/classes/cache/storage/memcached.php#L149

`$section` is prepended with the configured cache_id.

Then looking at https://github.com/fuel/core/blob/1.8.2/classes/cache/storage/memcached.php#L157

This is the branch for choosing between deleting everything or just the matching section - however,  ` ! empty($section)` looks like it can't ever be `false` because it will always contain the prepended cache_id.

Also note - it won't be an issue if `cache.memcached.cache_id` is an empty string.

Just checking to see if I'm seeing this correctly?

iturgeon:
The problem in the memcached cache driver prevents
\Cache::delete_all() from removingall the keys in the cache index.

The change allows that branch of the code to execute and includes
tests for the different states of cache_id and deleting with and
without a section in the arguments.

Signed-off-by: iturgeon <iturgeon@gmail.com>

fixes #2143

### DIFF
--- a/classes/cache/storage/memcached.php
+++ b/classes/cache/storage/memcached.php
@@ -145,22 +145,20 @@ class Cache_Storage_Memcached extends \Cache_Storage_Driver
 	 */
 	public function delete_all($section)
 	{
-		// determine the section index name
-		$section = $this->config['cache_id'].(empty($section) ? '' : '.'.$section);
-
-		// get the directory index
 		$index = static::$memcached->get($this->config['cache_id'].'__DIR__');
 
 		if (is_array($index))
 		{
-			// limit the delete if we have a valid section
-			if ( ! empty($section))
+			if (empty($section))
 			{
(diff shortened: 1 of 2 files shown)


WanWizard:
Doesn't this delete everything in the cache, regardless of the cache ID?

PR #2147, opened 2020-07-29
Do clearstatcache() before deleting files

yama:
How are you, WanWizard. I wish you good health.

FuelPHP is lightly processed, so we're using it in large projects. However, large projects are prone to errors when purging large numbers of files. I want to suppress this problem.
In Laravel and CakePHP, they are use clearstatcache() function. I think the same approach is ok.

Thanks.

### DIFF
--- a/classes/session/file.php
+++ b/classes/session/file.php
@@ -62,7 +62,7 @@ class Session_File extends \Session_Driver
 						clearstatcache(true, $this->config['path'] . $file);
 						if (is_file($this->config['path'] . $file))
 						{
-							@unlink($this->config['path'] . $file);
+							unlink($this->config['path'] . $file);
 						}
 					}
 				}
--- a/classes/session/file.php
+++ b/classes/session/file.php
@@ -59,7 +59,11 @@ class Session_File extends \Session_Driver
 						strpos($file, $this->config['cookie_name'].'_') === 0 and
 						filemtime($this->config['path'] . $file) < $expire)
 					{
-						@unlink($this->config['path'] . $file);
+						clearstatcache(true, $this->config['path'] . $file);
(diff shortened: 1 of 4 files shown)


WanWizard:
What is the exact purpose of this? What errors exactly? What does it fix?

yama:
https://www.google.com/search?q=unlink+error+clearstatcache
https://www.php.net/manual/en/function.clearstatcache.php
When I try to delete a file that should exist, I get a "This file does not exist" error.

WanWizard:
I know what `clearstatcache()` does, I am just wondering how and in what circumstances exactly you get that error.

Given the code, this can only be on a file_exist() that returns true on a file that doesn't exist, followed by an unlink of that file.

That can only happen if PHP did something with the file before so it has its existence in cache, and then some external process deletes the file, as an `unlink()` call would do a clear of the cache for that file.

classes/session/file.php
```diff
@@ -59,7 +59,11 @@ public function gc()
 						strpos($file, $this->config['cookie_name'].'_') === 0 and
 						filemtime($this->config['path'] . $file) < $expire)
 					{
-						@unlink($this->config['path'] . $file);
```
WanWizard, line 62: This line can go I think

classes/session/file.php
```diff
@@ -59,7 +59,11 @@ public function gc()
...
+						clearstatcache(true, $this->config['path'] . $file);
+						if (is_file($this->config['path'] . $file))
+						{
+							@unlink($this->config['path'] . $file);
```
WanWizard: And this "@" can be removed, as the file should exist after a `clearstatcache()` and an `is_file()` check

PR #2149, opened 2020-09-21
samesite configuration for cookies and session

nickhagen:
add session configuration parameter to allow developer to choose what samesite policy their cookies will use!

recent policy from Google has broken authenticated pages embedded via iframe:
https://www.chromium.org/updates/same-site

### DIFF
--- a/classes/cookie.php
+++ b/classes/cookie.php
@@ -100,17 +100,17 @@ class Cookie
 		$expiration = $expiration > 0 ? $expiration + time() : 0;
 		
 		if (PHP_VERSION_ID < 70300) {
-            return setcookie($name, $value, $expiration, "{$path}; samesite={$same_site}", $domain, $secure, $http_only);
-        }
+		    return setcookie($name, $value, $expiration, "{$path}; samesite={$same_site}", $domain, $secure, $http_only);
+		}
 
-        return setcookie($name, $value, [
-            'expires' => $expiration,
-            'path' => $path,
-            'domain' => $domain,
-            'samesite' => $same_site,
-            'secure' => $secure,
-            'httponly' => $http_only,
-        ]);
+		return setcookie($name, $value, [
+		    'expires' => $expiration,
+		    'path' => $path,
+		    'domain' => $domain,
+		    'samesite' => $same_site,
+		    'secure' => $secure,
+		    'httponly' => $http_only,
+		]);
 	}
 
 	/**
@@ -134,16 +134,16 @@ class Cookie
 		
 		// Nullify the cookie and make it expire
 		if (PHP_VERSION_ID < 70300) {
-            return setcookie($name, null, -86400, "{$path}; samesite={$same_site}", $domain, $secure, $http_only);
-        }
+		    return setcookie($name, null, -86400, "{$path}; samesite={$same_site}", $domain, $secure, $http_only);
+		}
 
-        return setcookie($name, null, [
-            'expires' => -86400,
-            'path' => $path,
-            'domain' => $domain,
-            'samesite' => $same_site,
-            'secure' => $secure,
-            'httponly' => $http_only,
-        ]);
+		return setcookie($name, null, [
+		    'expires' => -86400,
+		    'path' => $path,
+		    'domain' => $domain,
+		    'samesite' => $same_site,
+		    'secure' => $secure,
+		    'httponly' => $http_only,
+		]);
 	}
 }
--- a/classes/cookie.php
+++ b/classes/cookie.php
@@ -34,6 +34,7 @@ class Cookie
 		'domain'                => null,
 		'secure'                => false,
 		'http_only'             => false,
+		'same_site'             => 'none',
 	);
 
 	/*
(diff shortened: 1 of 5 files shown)


WanWizard:
Thanks. 

Please also update the cookie class documentation (https://github.com/fuel/docs/blob/1.9/develop/classes/cookie.html) ?

nickhagen:
I want to take another swing at this in the morning if you can hold off on
merging. I think SameSite None is only allowed if Secure is true ... so I
want to add a conditional to make sure that it is the case.

nickhagen:
Would you prefer that I update the config validator to throw some error or just update the documentation to say that in order to set samesite = none most browsers require s cure true?

WanWizard:
Not sure what you mean?

nickhagen:
Secure must be true in order for most browsers to allow Samesite set to None, otherwise, the browser will override the setting with Lax or Strict.

So I can add a check in the cookie config validate function that will catch and throw an error or I can update the documentation to just say this and if a developer sets secure false and samesite none, then browsers will just ignore it ...

Technically the code works as is of you have the config options set correctly

nickhagen:
This is all a new thing that chrome and firefox decided to roll out in august

https://blog.chromium.org/2019/10/developers-get-ready-for-new.html?m=1

WanWizard:
Maybe have a default for 'samesite' set to `false`, instead of `'none'`, and when set to `false` do not include `samesite` in the code. And when it is set to something else, validate it, and throw an exception if validation fails?

Than nothing changes for existing apps (we run all apps we build and maintain on develop, don't want them to fall over... :grin: )

nickhagen:
alright i think i got it now, if you have previous instances of Cookie::set() without the new parameter, it will default back to the old way of setting the cookie, developer must now explicitly set the configuration option if they want to change it.

i have updated the documentation here:
https://github.com/fuel/docs/pull/750

it is worth noting that if you **don't** set the samesite property ... chrome and firefox will start making your cookies samesite lax whenever they roll out their updates whether you use the old method or not, this will cause thing like iframe'd authenticated pages to start to fail if you leave your code unchanged

nickhagen:
also updated the Session class documentation here:
https://github.com/fuel/docs/pull/751

classes/cookie.php
```diff
@@ -93,11 +95,27 @@ public static function set($name, $value, $expiration = null, $path = null, $dom
 		is_null($domain) and $domain = static::$config['domain'];
 		is_null($secure) and $secure = static::$config['secure'];
 		is_null($http_only) and $http_only = static::$config['http_only'];
+		is_null($same_site) and $same_site = static::$config['same_site'];
 
 		// add the current time so we have an offset
 		$expiration = $expiration > 0 ? $expiration + time() : 0;
+		
+		if (is_null($same_site)) {
```
WanWizard: Fuel coding standards: brackets on a new line please.

classes/cookie.php
```diff
@@ -93,11 +95,27 @@ public static function set($name, $value, $expiration = null, $path = null, $dom
...
+		    return setcookie($name, $value, $expiration, $path, $domain, $secure, $http_only);
+		}
+		
+		if (PHP_VERSION_ID < 70300) {
```
WanWizard: Fuel coding standards: brackets on a new line please.

classes/cookie.php
```diff
@@ -93,11 +95,27 @@ public static function set($name, $value, $expiration = null, $path = null, $dom
...
+		    return setcookie($name, $value, $expiration, "{$path}; samesite={$same_site}", $domain, $secure, $http_only);
+		}
 
-		return setcookie($name, $value, $expiration, $path, $domain, $secure, $http_only);
+		return setcookie($name, $value, [
```
WanWizard: Fuel coding standards: don't use short array notation for PHP 5.x compatibility.

classes/cookie.php
```diff
@@ -110,15 +128,31 @@ public static function set($name, $value, $expiration = null, $path = null, $dom
...
 	 * @return  boolean
 	 * @uses    static::set
 	 */
-	public static function delete($name, $path = null, $domain = null, $secure = null, $http_only = null)
+	public static function delete($name, $path = null, $domain = null, $secure = null, $http_only = null, $same_site = null)
 	{
 		// Remove the cookie
 		unset($_COOKIE[$name]);
+		
+		if (is_null($same_site)) {
```
WanWizard: Fuel coding standards: brackets on a new line please.

classes/cookie.php
```diff
@@ -110,15 +128,31 @@ public static function set($name, $value, $expiration = null, $path = null, $dom
...
+		    return static::set($name, null, -86400, $path, $domain, $secure, $http_only);
+		}
 
 		// Nullify the cookie and make it expire
-		return static::set($name, null, -86400, $path, $domain, $secure, $http_only);
+		if (PHP_VERSION_ID < 70300) {
```
WanWizard: Fuel coding standards: brackets on a new line please.

classes/cookie.php
```diff
@@ -110,15 +128,31 @@ public static function set($name, $value, $expiration = null, $path = null, $dom
...
+		    return setcookie($name, null, -86400, "{$path}; samesite={$same_site}", $domain, $secure, $http_only);
+		}
+
+		return setcookie($name, null, [
```
WanWizard: Fuel coding standards: don't use short array notation for PHP 5.x compatibility.

nickhagen: fixed!

classes/cookie.php
nickhagen: fixed!

classes/cookie.php
nickhagen: fixed!

classes/cookie.php
nickhagen: fixed!

nickhagen:
Fixed all the coding styles! Sorry about that, I specifically tried to edit this without my editor config!

WanWizard:
No problem, thanks again.

PR #2154, opened 2020-11-06
See Issue #2153

### LINKED ISSUE #2153: Soap Trace Options
So it seems the way that the SOAP driver functions is to call set_defaults() during the execute() function, this seems to be problematic for the debugging functions found here: https://fuelphp.com/docs/classes/request/soap.html
* get_request_xml
* get_request_headers
* get_response_xml
* get_response_headers

These function all require the trace option to be enabled, however, the data from these functions only gets populated AFTER the call executes!  So if you try to print get_request_xml before you execute it is always null, if you call this function after execute, then the trace option gets reset on soap.php line 148 with $this->set_defaults();

Commenting out line 148 "fixes" the trace function and allows these functions to seemingly work properly, however, I am not sure why you would want to call set_defaults after a request is executed enough.  Perhaps you are calling execute() multiple times in a row ... but i would think you would still change the function and the parameter or headers that need to change before making the next couple calls?

Test case
```
$soap = Request::forge('https://api.cvent.com/soap/V200611.asmx?WSDL', 'soap');
$soap->set_function('Login');
$soap->set_option('trace', true);
$soap->set_option('connection_timeout', 45);
$soap->set_params([[
     'AccountNumber' => '123456',
     'UserName' => 'TestAccount',
     'Password' => 'thisCantBeCorrect',
]]);

$soap->execute();
```

nickhagen:
Remove resetting the parameters and options after executing the soap request.

fixes #2153

### DIFF
--- a/classes/request/soap.php
+++ b/classes/request/soap.php
@@ -145,12 +145,30 @@ class Request_Soap extends \Request_Driver
 			$mime = $this->response_info('content_type', 'application/soap+xml');
 			$this->set_response($body, $this->response_info('http_code', 200), $mime, $headers, isset($this->headers['Accept']) ? $this->headers['Accept'] : null);
 
+			$this->set_defaults();
 			return $this;
 		}
 		catch (\SoapFault $e)
 		{
+			$this->set_defaults();
 			throw new \RequestException($e->getMessage(), $e->getCode(), $e);
 		}
+	}
+	
+	/**
+	 * Extends parent to reset headers as well
+	 *
+	 * @return  Request_Soap
+	 */
+	protected function set_defaults()
+	{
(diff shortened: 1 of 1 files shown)


nickhagen:
Thank you for the clarification, for future reference:

>  It is indeed to use the same soap object for multiple requests, instead of forging a new one for each request.
> 
> If you pass "trace" as one of the options in the forge() request, it is considered a default option for the request, while set_option() is a runtime option, and only active for that request (execute call).
> 
> So you need to use
> 
> $soap = Request::forge('https://api.cvent.com/soap/V200611.asmx?WSDL', array('driver'=>'soap', 'trace' => true));

PR #2157, opened 2021-01-06
Update connection.php

hotlabs:


WanWizard:
You are not supposed to use the PDO driver, unless there is no platform driver for your RDBMS. Use a platform driver instead, which does have the platform specific code for this method.

PR #2158, opened 2021-04-10
Modify catch taret from Exception to Throwable, to clarify view error…

maruyama-arent3d:
When a view throws an Error (not an Exception), previous Fuel\Core\View::__toString() could not catch it.  
This behavior caused debugging difficulties at PHP 7.3.x or earlier, because the backtrace is destroyed and only COREPATH/\bootstrap.php's shutdown handler is shown on "Fatal Error" page.  
This behavior can be recognized by the following view code:

```
<?php
$foo = null;
$foo->bar(); // ErrorException [ Fatal Error ]: Method Parser\View::__toString() must not throw an exception, caught Error: Call to a member function bar() on null
```

This commit modifies try...catch target from Exception to Throwable.  
New Fuel\Core\View catches the Error objects and shows expected backtrace.

### DIFF
--- a/classes/view.php
+++ b/classes/view.php
@@ -226,7 +226,7 @@ class View
 		{
 			return $this->render();
 		}
-		catch (\Exception $e)
+		catch (\Throwable $e)
 		{
 			\Errorhandler::exception_handler($e);
 
@@ -257,7 +257,7 @@ class View
 				// Load the view within the current scope
 				include $__file_name;
 			}
-			catch (\Exception $e)
+			catch (\Throwable $e)
 			{
 				// Delete the output buffer
 				ob_end_clean();


WanWizard:
This can't be merged like this, Throwable doesn't exist in PHP 5.x,  and the current codebase is still 5.3+ compatible.

WanWizard:
Hmm, 5.x has been too long ago for me too :grinning:, in 5.x you can't catch a FATAL, only report after the facts.

Could you amend your PR with the same change in the Theme class (L#164) which also catches Exception, so it can be merged?

maruyama-arent3d:
Okay. I'll change Theme class.  
Don't I need fix Database_Query, Fieldset_Field, and so on which also catches Exception in __toString() method?

WanWizard:
I didn't look that far.

I think Theme, Presenter, Database_Query, Fieldset_Field. Others are ok, as they don't process anything that is user changeable.

maruyama-arent3d:
I think so, too.  
I'll close this PR and make a new PR. These changes will be greater than what this PR purposed at first.

WanWizard:
Fine by me.

PR #2159, opened 2021-04-10
Catching Throwable instead of Exception

maruyama-arent3d:
Some classes catch all exceptions to show error pages, but `Error` was miscatched on PHP 7.x because they catches only `Exception` class.  
This PR changes `catch (\Exception $e)` into `catch (\Throwable $e)` to prevent miscatching catchable fatal errors.

For example, the following view codes throw a `Call to a member function bar() on null` error, but it cannot be handled at `View::__toString()` on PHP 7.3 or earlier.  
Current codes will not catch this error, and `bootstrap.php`'s shutdown handler is called. At this time, no backtrace is shown on "Fatal Error" page, so it can be difficult to determine the error reasons.

```
<?php
$foo = null;
$foo->bar();
```

**This PR loses PHP 5.x compatibility** because `Throwable` is introduced on PHP 7.0.  
But, it is 2 years since PHP 5.x was no longer supported. Preventing debug difficulties on PHP 7.x or later may be more beneficial, I think....

### DIFF
--- a/classes/view.php
+++ b/classes/view.php
@@ -257,13 +257,10 @@ class View
 				// Load the view within the current scope
 				include $__file_name;
 			}
-			catch (\Throwable $e)
+			finally
 			{
 				// Delete the output buffer
 				ob_end_clean();
-
-				// Re-throw the exception
-				throw $e;
 			}
 
 			// Get the captured output and close the buffer
--- a/classes/view.php
+++ b/classes/view.php
@@ -226,7 +226,7 @@ class View
 		{
 			return $this->render();
 		}
-		catch (\Exception $e)
+		catch (\Throwable $e)
 		{
 			\Errorhandler::exception_handler($e);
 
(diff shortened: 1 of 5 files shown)


classes/view.php
```diff
@@ -257,13 +257,10 @@ protected function process_file($file_override = false)
 				// Load the view within the current scope
 				include $__file_name;
 			}
-			catch (\Exception $e)
+			finally
```
WanWizard, line 260: Why did you do this? This breaks everything, on all PHP versions !

maruyama-arent3d: I think finally is the better way than catching all exceptions ... ah, it will pass when no exception was thrown. It's my fault.
I requested revert PR #2160. Please merge it.

WanWizard:
Indeed. You'll end of with an empty page. I've already reverted it.

PR #2161, opened 2021-04-16
Catch high load file caching errors

nickhagen:
It seems to be a race condition that is occurring when multiple requests try to hit the expired cache at the same time!  Very difficult to reproduce in testing, these changes seemed to handle the issues in my production and I have not seen any more of these errors and the cache performance has not been impacted.  I was contemplating just wrapping the operation in a try/catch, but from what I have read these operations fopen and file_get_contents will not generate an exception.

Here are a few errors that pile up in the fuel logs during high load cache changing:
```
file_get_contents(/fuel/app/cache/db/6358477e0fb17783679ecb078ef6f053.cache): failed to open stream: No such file or directory in /fuel/core/classes/cache/storage/file.php on line 340
```

and this one as well:
```
fopen(/fuel/app/cache/db/74d391d280cc70ee51f8ecbe1cb59366.cache): failed to open stream: No such file or directory in /fuel/core/classes/cache/storage/file.php on line 333
```

### DIFF
--- a/classes/cache/storage/file.php
+++ b/classes/cache/storage/file.php
@@ -330,22 +330,25 @@ class Cache_Storage_File extends \Cache_Storage_Driver
 		// make sure it exists
 		if (is_file($file))
 		{
-			$handle = fopen($file, 'r');
-			if ($handle)
-			{
-				// wait for a lock
-				while( ! flock($handle, LOCK_SH));
-
-				// read the cache data
-				$payload = file_get_contents($file);
-
-				//release the lock
-				flock($handle, LOCK_UN);
-
-				// close the file
-				fclose($handle);
-
+			$handle = @fopen($file, 'r');
+			if ( ! $handle) {
+				return false;
 			}
+			
+			// wait for a lock
+			while( ! flock($handle, LOCK_SH));
+
+			// read the cache data
+			$payload = @file_get_contents($file);
+			if ( ! $payload) {
+				return false;
+			}
+
+			//release the lock
+			flock($handle, LOCK_UN);
+
+			// close the file
+			fclose($handle);			
 		}
 
 		try


WanWizard:
Please remove the "@".  If you expect errors, try/catch and handle them, instead of suppressing them. 
Its use is deprecated in PHP8 anyway, and won't catch most of the errors anymore.

nickhagen:
Unfortunately these are PHP warnings/errors and not exceptions that are filling the logs, and wrapping the entire operation in a try/catch is not actually catching them. 

It seems like this is the accepted method for suppressing file system warnings from what i have found:
* https://stackoverflow.com/questions/272361/how-can-i-handle-the-warning-of-file-get-contents-function-in-php
* https://www.geeksforgeeks.org/how-to-handle-the-warning-of-file_get_contents-function-in-php/
* https://mike42.me/blog/2019-03-07-handling-io-errors-in-php

If you aren't comfortable with this approach you can close this PR and I can extend this core class.  But I would think anyone using this cache driver in a relatively high traffic site would be seeing these errors in the logs pretty consistently.

WanWizard:
What warnings and errors are you talking about, and what PHP version? In general, Fuel catches all PHP errors, except E_FATAL in some PHP versions (cause the error handler isn't called on them).

And in PHP8, they don't work anymore: https://php.watch/versions/8.0/fatal-error-suppression

WanWizard:
Also, looking at the code, you now return if no payload could be retrieved, without closing the handle and releasing the lock?

In old PHP versions, this would happen automatically at the end of the script, but in newer versions, this no longer happens so the file stay locked in case of PHP-FPM (as the php process keeps running, so the OS can not do a force-close).

WanWizard:
I've tested it, a failed file_get_contents() can be caught just fine:
image

nickhagen:
ok, sorry about that, must be something going on with my error handling
rules then, let me take another pass and get back to you next week.

WanWizard:
Which PHP version are you using? Because they might behave differently. We run development on 8.0.3 at the moment.

nickhagen:
I am running on PHP 7.4

WanWizard:
Switched my development vhost to 7.4, same behaviour.

nickhagen:
I am thinking this must be a factor of the issues with my error handling levels and continue on error settings, sorry about that!

PR #2168, opened 2022-02-02
Supress deprecation warnings in PHP 8.1

Art4:
Hey 👋 

PHP 8.1 gives some deprecations warnings for some interfaces where the return types will be changed.

In this PR I've added the `#[\ReturnTypeWillChange]` attribute to keep BC with PHP 7 and supress the warnings and I added the future type hints as comments.


WanWizard:
Work on PHP 8.1 compatibility is on the way.

Art4:
> Work on PHP 8.1 compatibility is on the way.

And I hope that I can help with this PR. 😃

PR #2171, opened 2022-08-01
Fix redis connection issue

Rin3917:
Fix Redis connection issue in 8.1 mentionned on [this thread](https://fuelphp.com/forums/discussion/15418/php-8-1)

### DIFF
--- a/classes/redis/db.php
+++ b/classes/redis/db.php
@@ -254,8 +254,10 @@ class Redis_Db
 
 	protected function readResponse()
 	{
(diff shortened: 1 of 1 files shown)


WanWizard:
It looks like the root cause of this is a misplaced bracket
````
if (($reply = fgets($this->connection, 512) === false))
````
should be
````
if (($reply = fgets($this->connection, 512)) === false)
````
but I can live with this fix :grin:

PR #2172, opened 2022-08-03
Fix redis infinite loop issue

Rin3917:
This commit clean and fix Redis potential infinite loop by trimming value at the begining.

With previous code, the `$-1` case wasn't trimmed and would end with an infinite loop. Also, it's cleaner to trim it at the begining rather than in each case.

Also, fix `unserialize` issue in session driver class that would trigger a php 8.1 error if the input param is null.

### DIFF
--- a/classes/redis/db.php
+++ b/classes/redis/db.php
@@ -254,24 +254,19 @@ class Redis_Db
 
 	protected function readResponse()
 	{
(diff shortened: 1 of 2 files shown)


WanWizard:
Thanks. I've manually split the PR into two commits, as these two fixes are unrelated.

PR #2173, opened 2022-08-09
Php 8.1 update - multiples fixs

Rin3917:
This PR contain 5 differents commits related to php 8.1 support. The fixes are detailled in each individuals commits.

### DIFF
--- a/classes/cookie.php
+++ b/classes/cookie.php
@@ -86,7 +86,6 @@ class Cookie
 		{
 			return false;
 		}
-
 		$value = \Fuel::value($value);
 
 		// use the class defaults for the other parameters if not provided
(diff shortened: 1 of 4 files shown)


classes/security.php
```diff
@@ -191,7 +191,7 @@ public static function strip_tags($value)
 	{
 		if ( ! is_array($value))
 		{
-			$value = filter_var($value, FILTER_SANITIZE_STRING);
+			$value = strip_tags($value);
```
WanWizard: I don't think this is a proper replacement for what was intended. It might be better to replace it with
````
$value = filter_var(strip_tags($value), FILTER_SANITIZE_FULL_SPECIAL_CHARS);
````

File: classes/cookie.php [hunk 2]
```diff
@@ -86,7 +86,7 @@ public static function set($name, $value, $expiration = null, $path = null, $dom
 		{
 			return false;
 		}
-
+		
 		$value = \Fuel::value($value);
```
WanWizard, line 89: Also, get an editor that trims lines :grin: .

WanWizard:
Also, the Format change is unrelated to PHP 8.1, so should be seperate (a good idea btw !)

Rin3917:
Done.

PR #2175, opened 2022-08-16
Fix issue with package and module

Rin3917:
Fix bug induced by commit 2317bc2591f58bb9eec3849f72a1a6f6dd47ab7a from a couple of weeks in the `loaded` and `exists` methods of both classes `Package` and `Module`.

Since its capitalized on load, but the `array_key_exists` doesn't make sure it is.

### DIFF
--- a/classes/module.php
+++ b/classes/module.php
@@ -170,7 +170,7 @@ class Module
 			return static::$modules;
 		}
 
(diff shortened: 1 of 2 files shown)


WanWizard:
Hmm., yeah, well spotted! :+1:

PR #2185, opened 2022-11-16
PDO List tables support

phpukr:
DB PDO List tables support


WanWizard:
The PDO driver is a generic driver, which is why it doesn't have a LIST TABLES function, since that is RDBMS specific.

The PDO driver shouldn't be used on its own. Either use a RDBMS driver that extends PDO (dblib, mysql, sqlsrv, sqlite), or if you use a RDBMS that currently does not have a specific driver, please create one.

PR #2190, opened 2023-05-08
[php81] fix strstr string parameter

polmartinez:


WanWizard:
Given the fact strstr() wants a string, it is probably cleaner to do
````
$content_header = \Input::headers('Content-Type', "");
````
instead of casting `false` to string...

polmartinez (commented):
change default parameter

PR #2191, opened 2023-06-20
Fix PHP 7.4 deprecated code

Vegeeto:
On implode function, passing a glue string after array is deprecated. I have changed it to the same piece of code that is in the __call method


WanWizard:
Thanks.

However, wouldn't it be simpler to just swap the arguments, to move the separator string to the front?

PR #2197, opened 2023-10-24
enable connection options through the configuration

polmartinez:
be able to pass ssl connection information, I add the "options" group of the configuration array.


WanWizard:
Why this, instead of just adding 
````
'attrs' => array(),
````
directly to the config?

polmartinez:
yes, it works. I hadn't thought of that.

PR #2199, opened 2023-11-15
[php81] make MySQLi error reporting compatible with before PHP 8.1

shota-munemoto:
Migrations and tests fail when using MySQLi connection in PHP 8.1 or later.

So I have changed the error reporting mode to be compatible with pre-PHP 8.0 behavior.

PHP: Backward Incompatible Changes - Manual
https://www.php.net/manual/en/migration81.incompatible.php

> The default error handling mode has been changed from "silent" to "exceptions" See the MySQLi reporting mode page for more details on what this entails, and how to explicitly set this attribute. To restore the previous behaviour use: mysqli_report(MYSQLI_REPORT_OFF);

### DIFF
--- a/classes/database/mysqli/connection.php
+++ b/classes/database/mysqli/connection.php
@@ -66,6 +66,9 @@ class Database_MySQLi_Connection extends \Database_Connection
 			),
 			'enable_cache'   => true,
 		), $this->_config);
+
+		// Make error reporting compatible with the behavior prior to PHP 8.1
+		mysqli_report(MYSQLI_REPORT_OFF);
 	}
 
 	public function connect()


WanWizard:
What exactly fails, and why?

I've never seen a migration fail because of this, and we run PHP 8.2 everywhere.

shota-munemoto:
**First time** migration with MySQLi fails.  
So you cannot see the error when using another database connection or created already migration table in PHP 8.0 or older.

I created a simple project to explain of this pull request.

shota-munemoto/fuel-core-mysqli-error-php81
https://github.com/shota-munemoto/fuel-core-mysqli-error-php81

WanWizard:
The root cause of this problem is that the code expects a Database_Exception, not a mysqli_sql_exception.

shota-munemoto:
I didn't realize it!
Thank you!

PR #2200, opened 2023-11-17
chmod() expects 2 arguments

shota-munemoto:
just replaced "." with ",".

PHP: chmod - Manual
https://www.php.net/manual/en/function.chmod.php


WanWizard:
My eyes are clearly not what they used to be, old age is catching up.

Thanks for spotting this !

PR #2201, opened 2023-11-28
Update composer.json

phpukr:
phpseclib < 3.0.34 vulnerable to denial of service

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
 	],
 	"require": {
 		"composer/installers": "~1.0",
-		"phpseclib/phpseclib": "~2.0",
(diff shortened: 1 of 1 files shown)


WanWizard:
Yes, aware of that, we've got a dependabot alert.

Currently looking at the impact, as it is a major version upgrade..

WanWizard:
First indications are 3.x is not compatible. 

A simple login doesn't work anymore, but since Auth doesn't use Crypt, I think it may be related to session encryption.

WanWizard:
Updated Crypt to use PHPSecLib 3.

Thanks for reporting it.

PR #2202, opened 2023-12-04
undefined variable

polmartinez:
undefined variable


WanWizard:
Chips, I was absolutely convinced I'd pushed the fix for that one. Thanks for spotting it !

PR #2206, opened 2024-02-01
GET in php 8.2 sometimes has null values

Thunder-Dan:


WanWizard:
Do you have an example? And why specific to PHP 8.2?

Also, there should be a space between if and bracket `if (` as per the coding standards. 

And indentation still doesn't seem rights (tabs, not spaces), see https://github.com/fuel/core/blob/ea46e3291ad4bdb88f1bd59390cbae9c20a5273b/classes/security.php

PR #2208, opened 2024-04-21
feat(deps): permit sodium_compat v1 or v2 to be installed

paragonie-security:
https://paragonie.com/blog/2024/04/release-sodium-compat-v2-and-future-our-polyfill-libraries


WanWizard:
Thanks !

PR #2216, opened 2024-08-18
Implicitly marking parameter as nullable is deprecated

arshidkv12:


WanWizard:
We can't do that, Fuel is, until now, compatible with PHP versions that don't support that. 
That won't change until we go to a new major release, i.e. 2.0.

arshidkv12:
Thank you

WanWizard:
Where and how do you get this decpration message, as we use the latest PHP 8.3, and haven't seen any.

WanWizard:
Also, this is only an issue if the argument is typed, which is why typing has been removed, and replaced by a check in the code.

arshidkv12:
I tested with php 8.4 dev version.

WanWizard:
You shouldn't get a deprecation message, as that is only for typed arguments? See https://php.watch/versions/8.4/implicitly-marking-parameter-type-nullable-deprecated under "Dropping parameter type".

arshidkv12:
Correct. No issues.

```php
public static function forge($new = null, $input = null)
```

WanWizard:
Thanks for confirming. 

We'll try to make 1.9-dev compatible with 8.4, and release a formal 1.9 as soon as 8.4 is out, after which we can work on a 2.0 which will drop support for PHP 5.x and 7.x and will be strictly typed.

arshidkv12:
Okay. Thank you.

PR #2217, opened 2024-08-18
Implicitly marking parameter as nullable is deprecated

arshidkv12:
Implicitly marking parameter as nullable is deprecated


WanWizard:
We can't do that, Fuel is, until now, compatible with PHP versions that don't support that.
That won't change until we go to a new major release, i.e. 2.0.

PR #2218, opened 2024-08-19
phpunit InvalidArgumentException

arshidkv12:
Fix phpunit failer issues.


WanWizard:
I can't merge this one either, it is not compliant with PHP 5.3+, which Fuel officially still needs to support.

arshidkv12:
Ok

PR #2219, opened 2024-09-02
creates an extra level of cache directories

polmartinez:
To improve disk access, we add one more level of depth.

### DIFF
--- a/classes/cache/storage/file.php
+++ b/classes/cache/storage/file.php
@@ -183,7 +183,12 @@ class Cache_Storage_File extends \Cache_Storage_Driver
 	{
 		// replace dots with dashes
 		$identifier = str_replace('.', DS, $identifier);
-
+        	if (strpos($identifier, '.') !== false) {
+	            [$group, $hash] = explode('.', $identifier, 2);
+		    // creates an extra level
+	            $hash = substr_replace($hash, '.', 3, 0);
+	            $identifier = implode('.', [$group, $hash]);
+	        }
 		return $identifier;
 	}
 
--- a/classes/cache/storage/file.php
+++ b/classes/cache/storage/file.php
@@ -181,14 +181,14 @@ class Cache_Storage_File extends \Cache_Storage_Driver
 	 */
 	protected function identifier_to_path($identifier)
 	{
-		// replace dots with dashes
-		$identifier = str_replace('.', DS, $identifier);
+		// creates an extra level
         	if (strpos($identifier, '.') !== false) {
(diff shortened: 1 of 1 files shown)


WanWizard:
How much cached data do you have that this actually improves performance?

Also, code is not following the coding standards, and it uses short array notation, so it's not compatible with all supported PHP versions.

polmartinez:
In my case, cache 40Gb on disk. All in one directory is crazy.
We can make the 2 changes you need to follow coding standards.

WanWizard:
That is a lot of cache :grin: 

Now having time to look at this code, what exactly is this supported to do? Line 185 does
````
$identifier = str_replace('.', DS, $identifier);
````
after which all dots are gone from `$identifier`, and paths are created. So if you have an identifier like "a.b.c", this method converts that to "a/b/c" (on Linux), so you can already control the folder structure in your code, but using the correct identifiers.

Also, after this line is executed, all dots in the identifier are gone,  so
````
if (strpos($identifier, '.') !== false) {
````
will always return false and will never be executed.  The code you suggest is a no-op?

polmartinez:
the idea is to add one more 3-digit level. There was a mistake in the merge, but I have already fixed the PR.

WanWizard:
I'm still not on board.

If you want a more complex cache storage structure, you make sure, in your app, that the identifier used reflects that. For example by using `$identifier = 'order.'.substr(0, 3, $ordernumber).'.'.$ordernumber.` when you construct the cache identifier.

If you can't or won't do that (I can imagine that retroactive changes to an existing app is a bit of a bother), you can overload the Cache_Storage_File class in your app and add this change there. 

I don't really see why this, your use-case specific solution, should be in the generic framework code.

polmartinez:
Yes, I've had it overloaded for months, but I thought that as it is something generic, that affects directories in all systems, it could be interesting for it to be in the core.
It's ok, cancel the PR without any problem.

WanWizard:
Well, the issue I have with it is that it is not generic, it is specific to your use case.

We have an app in maintenance that has over a million files in cache (all small), but that app uses "modulename/a/b/c/abcvalue", so it uses the first 3 characters to create a 3-deep directory structure made up of single letters.

A problem similar to what you have, but a different solution (one extra level wasn't enough to keep the number of files in a directory low enough). 

In your app code you have logic to generate the identifier when you store or retrieve a cache item, this is the place where you construct the correct identifier for the kind of cache structure you want. In the above case, a generic method is used that gets a database model passed, and returns "modulename.a.b.c.abcvalue", based on data in the model object.

PR #2222, opened 2025-05-18
ignore configuration if log is not wanted

pablobegreen:


WanWizard:
Thanks for your contibution.

This PR violates coding standards (spaces instead of tabs, brace on the same line).

Also, it is potentially dangerous, as it leaves Log with an uninitialized Monolog instance, and a potential crash when something calls `\Log::instance()->log()`, which could also be in a package or external code.

WanWizard:
Still spaces instead of tabs?

PR #2223, opened 2025-06-19
Improve fallback logic for validation error messages

arshidkv12:
Adds fallback to fieldset validation and language messages when field-level error is not set.


WanWizard:
Don't know why you sneaked that last one in, as it is both unrelated and wrong.

PR #2224, opened 2025-09-26
Fix case conversion function calls in str.php

marcbush:
replace str_to_lower with strtolower


WanWizard:
That was stupid, thanks for catching it !

PR #2226, opened 2025-10-07
Update parameter type in docblock for set_global

marcbush:
Another little one to stop IDE squiggles!


WanWizard:
Thanks. I don't use an IDE, just an old-fashioned editor, so never notice it.

PR #2227, opened 2025-11-21
Cast string to ensure proper case conversion

pablobegreen:
Failure in PHP 8.4. I prefer to cast rather than force a parameter that is not backward compatible.

### DIFF
--- a/classes/str.php
+++ b/classes/str.php
@@ -441,8 +441,8 @@ class Str
 		$encoding or $encoding = \Fuel::$encoding;
 
 		return (MBSTRING and $encoding)
-			? mb_strtolower($str, $encoding)
-			: strtolower($str);
+			? mb_strtolower((string) $str, $encoding)
+			: strtolower((string) $str);
 	}
 
 	/**
@@ -458,8 +458,8 @@ class Str
 		$encoding or $encoding = \Fuel::$encoding;
 
 		return (MBSTRING and $encoding)
-			? mb_strtoupper($str, $encoding)
-			: strtoupper($str);
+			? mb_strtoupper((string) $str, $encoding)
+			: strtoupper((string) $str);
 	}
 
 	/**


WanWizard:
I can not reproduce that:
````
Fuel 1.9-dev - PHP 8.5.0 (cli) (Nov 18 2025 08:02:20) [Linux]

Commands
:q | quit - exit the console
:h | history - show transcript

>>> echo Str::strtoupper("asasas");
ASASAS
>>>
````

**Comment by pablobegreen** [comment]
```
$var = null;  
Str::strtoupper($var);
```

WanWizard:
That is a problem you need to fix in your app, not hack around in the framework. You would have done the same if you'd used the built-in function  `strtoupper()`.  

The framework does not, and has never done, any input validation. That will change for v2.0, which will drop support for PHP 5 and PHP 7.

Once quick fix is
````
Str::strtoupper($var ?? '');
````
Although if you expect a string at that point, you probably have a logic error somewhere.

pablobegreen:
Well, from my point of view, if the parameter is not typed, we could pass anything, such as a database field that can be null.
That's why everything related to strings should also be checked to ensure it is not null, either by casting or with an IF statement.
I think what I have proposed is a good solution.

WanWizard:
Then we have to agree to disagree.

Not only it creates a precedent for all other places in the current codebase were validation doesn't happen, it also means that when we introduce type hinting in 2.0-dev, we have to remove it all again.

pablobegreen:
Okay, by typing a method, you eliminate all possibilities of error.
I understand your point of view, and I even agree, but for those of us who have been using FuelPHP for 15 years, even with applications that have been running for 10 years, when PHP forces us to type so many things, such as strings, I believe, and this is my opinion, that the framework should make our lives easier.

WanWizard:
We have applications still in production that were written for 1.0, in 2011., And new applications released in the last few months.

It is always a challenge, with every new PHP release, to keep on top of things in your codebase, that is not to be avoided. 2.0 will introduce strict typing, I expect quite a bit of work will be needed when upgrading apps from 1.9 (due soon) to 2.0, just for cases just like this.

All our new apps are PHP 8+ and already use strict typing, but even for those there will be cases were ints are passed as strings, or visa versa. So no matter what, it is going to be work.

We aim to keep Fuel as compatible as possible, especially since some applications do have a longer life cycle than just a few years, but it is getting harder and hander to have one codebase to support everything from PHP 5.4 to 8.5...

pablobegreen:
ok, I'm closing pr

PR #2230, opened 2025-12-25
Update MySQL PDO compression attribute for PHP 8.5 compatibility

phpukr:
Currently, FuelPHP sets the PDO compression attribute using PDO::MYSQL_ATTR_COMPRESS. Starting from PHP 8.5, this constant is deprecated and replaced by Pdo\Mysql::ATTR_COMPRESS.


WanWizard:
Fixed as part of ded85bb89a93d961821cc68d32b62ee1d9abd9e6

